PR #3, opened 2017-04-24
Added support for sending MRPC_SES commands to the firmware.

donaldematthews:


cli/builtin.h
```diff
@@ -43,6 +43,7 @@ COMMAND_LIST(
 	ENTRY("evcntr-show", "Show an event counters setup info", evcntr_show)
 	ENTRY("evcntr-del", "Deconfigure an event counter", evcntr_del)
 	ENTRY("evcntr-wait", "Wait for an event counter to exceed its threshold", evcntr_wait)
+	ENTRY("mrpc-ses", "Send an MRPC SES command to the firmware", mrpc_ses)
 );
```
lsgunth: Every command is an mrpc command and none of them have an mrpc- prefix. This should just be 'ses' and save the user some typing.

cli/main.c
```diff
@@ -1071,6 +1071,134 @@ static int fw_update(int argc, char **argv, struct command *cmd,
...
+
+    if( (input >= 0x61) && (input <= 0x66) )
+    {
+        return ( (uint8_t)input - 0x57 );
+    }
+
+    return ( 0xFF );
+}
+
+int switchtec_send_mrpc_ses_command(struct switchtec_dev *dev, FILE *fp)
```
lsgunth: This should be in the library not in the CLI. Especially seeing you added the prototype to the librarie's header.

cli/main.c
```diff
@@ -1071,6 +1071,134 @@ static int fw_update(int argc, char **argv, struct command *cmd,
...
+
+                input = (char)fgetc( fp );
+                byte  = ascii_to_bin( input );
+                mrpc_input[index] |= byte;
+                index++;
+            }
+        }
+
+    }   while( (exit == 0) && (index < MRPC_MAX_DATA_LEN) );
+
```
lsgunth: This is an unacceptable interface. Expecting the user to construct hand crafted hex files and then have to parse the result is way too complicated. It's too reminiscent of the psx_app era. Everything in switchtec-user must largely be usable without needing to comb through the switch's device specification document.

cli/main.c
```diff
@@ -1071,6 +1071,134 @@ static int fw_update(int argc, char **argv, struct command *cmd,
...
+            if( byte != 0xFF )
+            {
+                mrpc_input[index] = byte << 4;
+
+                input = (char)fgetc( fp );
+                byte  = ascii_to_bin( input );
+                mrpc_input[index] |= byte;
+                index++;
+            }
```
lsgunth: Use strtol instead of reinventing the wheel...

lsgunth:
Per side conversation, this it's taking a different route. 

Closing.

Issue #7, opened 2018-02-22
device info is wrong in status command


glimchb:
To reproduce run : switchtec status /dev/switchtec1
Only few device info appear.
The parsing of sysfs seems to be wrong.

Here is the fix:

[switchtec-issue7.log](https://github.com/Microsemi/switchtec-user/files/1747784/switchtec-issue7.log)

PR #8, opened 2018-02-22
fix a bug (issue #7) : device info is wrong in status command

glimchb:


lsgunth:
Many thanks. Looks great. I've merged this commit into the devel branch.

Logan

PR #10, opened 2018-04-19
Fix cli input issue

JoeyZhang-Microsemi:
On 32 bit OS, the return value of function strtol uses 32 bits. Here we should use 64 bits.


lsgunth:
Generally, it's a better idea to merge things into 'devel' not straight into master. Especially seeing TravisCI has not been run on it.

That being said the change looks trivially correct so it's probably ok. Just a note for the future.

JoeyZhang-Microsemi:
Thanks for your review. I will follow it.

PR #11, opened 2018-04-20
Support firmware download timout.

JoeyZhang-Microsemi:
Add a firmware download status (SWITCHTEC_DLSTAT_DOWNLOAD_TIMEOUT) for MRPC command "Firmware Download - Get Status Sub-Command".


lsgunth:
Looks good except the patches are in the wrong order. Each patch should compile when applied so you should add the statuses first, then the error messages, and then finally make use of them.

Please rebase to fix it before merging.

Thanks,

Logan

JoeyZhang-Microsemi:
Hi Logan,

Thanks for your review and comments. I have fixed it.

Thanks
Joey

lsgunth:
Looks good to me.

Logan

wesleywesley:
Previously,  only system call write and read can introduce return value less than zero, while in other mrpc fw download execution failure cases, the return value is positive for switchtec_fw_wait function.
And switchtec_fw_write_fd/switchtec_fw_write_file functions exit only when system call failure happen, but stay in the firmware download loop in other mrpc fw download exection  failure case, because the return value greater than 0.

Change the return value check condition, and in any failure case, no matter it is system call or mrpc fw  packet unit download, will exit the fw download loop instantly.

PR #14, opened 2018-04-25
Ensure we print the platform perror in switchtec_fw_perror

wesleywesley:
Found this bug fix is not merge into devel and master


lsgunth:
Yup, I think I made that commit but we were digging into another bug but it somehow got dropped.

lsgunth:
Oh, oops I thought I posted a comment on that. That particular patch wouldn't compile anymore, and I'm not sure what I was thinking when I wrote it, so I backed it out.

PR #16, opened 2018-05-15
output data length of echo mrpc cmd bug fix

wesleywesley:
actually it is a typo


lsgunth:
Could have sworn we did that already...

Logan

PR #17, opened 2018-05-25
Fix firmware update error checking issue.

JoeyZhang-Microsemi:
Function switchtec_fw_wait() returns 0 if it succeed. Function switchtec_fw_write_fd() and switchtec_fw_write_file() use incorrect error checking for function switchtec_fw_wait().

### DIFF
--- a/lib/fw.c
+++ b/lib/fw.c
@@ -217,7 +217,7 @@ int switchtec_fw_write_fd(struct switchtec_dev *dev, int img_fd,
 			return ret;
 
 		ret = switchtec_fw_wait(dev, &status);
-		if (ret < 0)
+		if (ret != 0)
 		    return ret;
 
 		offset += cmd.hdr.blk_length;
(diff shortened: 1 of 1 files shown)


wesleywesley:
Previously,  only system call write and read can introduce return value less than zero, while in other mrpc fw download execution failure cases, the return value is positive for switchtec_fw_wait function.
And switchtec_fw_write_fd/switchtec_fw_write_file functions exit only when system call failure happen, but stay in the firmware download loop in other mrpc fw download exection case, because the return value greater than 0.

Change the return value check condition, and in any failure case, no matter it is system call or mrpc fw  packet unit download, will exit the fw download loop instantly.

wesleywesley:
Previously,  only system call write and read can introduce return value less than zero, while in other mrpc fw download execution failure cases, the return value is positive for switchtec_fw_wait function.
And switchtec_fw_write_fd/switchtec_fw_write_file functions exit only when system call failure happen, but stay in the firmware download loop in other mrpc fw download exection  failure case, because the return value greater than 0.

Change the return value check condition, and in any failure case, no matter it is system call or mrpc fw  packet unit download, will exit the fw download loop instantly.

lsgunth:
The change looks good but the information in the PR should have been in the commit message instead of a blank commit message...

Logan

wesleywesley:
Thx Logan for remind.

PR #18, opened 2018-05-25
Support firmware download timeout.

JoeyZhang-Microsemi:
Add a new firmware download status "SWITCHTEC_DLSTAT_DOWNLOAD_TIMEOUT" for firmware download timeout.


wesleywesley:
it is better to divide this commit into two commits, first for add new enumeration and the second for use it.

PR #20, opened 2018-06-15
Set errno to ENODEV when ret equal to NULL

wesleywesley:


lsgunth:
Please get in the habit of writing better commit messages. You should describe the motoviation for the change, etc; not just what was changed in words. Otherwise this looks fine.

wesleywesley:
Add commit messages for this bug fix.

PR #21, opened 2018-06-15
Uart dev stream

wesleywesley:
Background for UART based utility

This code is cooperate with Switchtec device's uart debug port, whose cli with bytes aligned stream (-s) and crc check (-c) capabilities, different from previous version's dword aligned and dword write/read unit based access mode.

A: Following are examples:
1. GAS read:
**************************************
0x00000000:1203>gasrd -c -s 0x3 5
gas_reg_read <0x3> [5 Byte]
00 58 00 00 00
CRC: 0x37
0x00000000:1204>
**************************************
Where 0x3 is the byte aligned gas offset, 5 is the length of bytes to read.
In echo on mode, the results(fw return string) of this read request is start at "gasrd -c -s ..." and end up with "0x00000000:1204>"
In echo off mode, the results is start at "gas_reg_read ..." and end up is same as echo on mode.
We turn off the echo mode(echo_off) at the beginning in "switchtec_open_uart" function, to save bandwidth; we also turn off the debug port's debug print(dbginfo_off), to make sure, only output from gasrd/gaswr request exclusively.

2. GAS write:
**************************************
Fail case:
0x00000000:0000>gaswr -c -s 0x5 0xaabbccddeeff 0xb
gas_reg_write() CRC Error
CRC: [0x84/0x0b]
0x00000000:0000>

Success case:
0x00000000:0000>gaswr -c -s 0x5 0xaabbccddeeff 0x84
gas_reg_write() success
CRC: [0x84/0x84]
0x00000000:1212>
**************************************
Where the crc byte after byte stream is calculated by utility.
The return results include the expected crc and device side received crc value.

B: Check the finishing of the return string send by FW:
By check the specific symbols in the return string to make sure all of FW's response to the previous gasrd/gaswr command have been collected:
For read, check the offset after "<0x", number after "[", crc after" : 0x" and cli index after ":"
sscanf(gas_rd_rtn, "%*[^<]<0x%x> [%d Byte]%*[^:]: 0x%x%*[^:]:%x>", &raddr, &rnum, &rcrc, &idx)
For write, check the fw calculated crc after ": [0x", fw expect crc after "/0x" and also cli index after ":"
sscanf(gas_wr_rtn, "%*[^:]: [0x%x/0x%x]%*[^:]:%x>", &crc_cal, &crc_exp, &idx)

Because the time duration for different MRPC command's execution is different (for example, fw-toggle mrpc command take much longer time than other mrpc's execution)
and also the mrpc return data with different length of data payload (fw-read/gui/bw/status with more data payload of feedback than fw-update)
So we implemented a big enough retry counter(experience value got from real test) companion with 5000us delay interval; until we collected all the response characters.

For echo on/off and debug print off cli commands, just check if a index after ":" symbol is received, if it is, utility code consider the response from the cli commands is all received.
Here check the index after ":", as:
sscanf(rtn, "%*[^:]:%x>", &idx)

### DIFF
--- a/lib/platform/linux-uart.c
+++ b/lib/platform/linux-uart.c
@@ -1,0 +1,444 @@
+/*
+ * Microsemi Switchtec(tm) PCIe Management Library
+ * Copyright (c) 2018, Microsemi Corporation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+ * OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ *
+ */
+
+#ifdef __linux__
+
+#include "../switchtec_priv.h"
+#include "../crc8.h"
+#include "switchtec/switchtec.h"
+#include "gasops.h"
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <sys/ioctl.h>
+#include <sys/mman.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <errno.h>
+#include <signal.h>
+#include <stddef.h>
+#include <assert.h>
+#include <string.h>
+#include <stdarg.h>
+
+#include <sys/file.h>
+#include <termios.h>
+
+struct switchtec_uart{
+	struct switchtec_dev dev;
+	int fd;
+};
+
+#define to_switchtec_uart(d) \
+	((struct switchtec_uart *) \
+	 ((char *)(d) - offsetof(struct switchtec_uart, dev)))
+
+#define UART_MAX_WRITE_BYTES			100
+#define UART_MAX_READ_BYTES			1024
+#define RETRY_NUM				3
+#define SWITCHTEC_UART_BAUDRATE			(B230400)
+
+static int send_cmd(int fd, const char *fmt, int write_bytes, ...)
+{
+	int ret;
+	int i;
+	int cnt;
+	char cmd[1024];
+	uint8_t *write_data;
+	va_list argp;
+
+	va_start(argp, write_bytes);
+
+	cnt = vsnprintf(cmd, sizeof(cmd), fmt, argp);
+
+	if (write_bytes) {
+		write_data = va_arg(argp, uint8_t *);
+		for (i = 0; i< write_bytes; i++) {
+			cnt += snprintf(cmd + cnt, sizeof(cmd) - cnt,
+				       "%02x", write_data[write_bytes - 1 - i]);
+		}
+
+		cnt += snprintf(cmd + cnt, sizeof(cmd) - cnt,
+				" 0x%x\r", va_arg(argp, int));
+	}
+
+	va_end(argp);
+
+	ret = write(fd, cmd, cnt);
+	if (ret < 0)
+		return ret;
+
+	if (ret != cnt) {
+		errno = EIO;
+		return -errno;
+	}
+
+	return 0;
+}
+
+static int read_resp_line(int fd, char *str)
+{
+	int ret;
+	int cnt = 0;
+
+	while(1) {
+		ret = read(fd, str + cnt, sizeof(str));
+		if (ret <= 0)
+			return ret;
+
+		cnt += ret;
+		str[cnt] = '\0';
+
+		/* Prompt "0x12345678:1234>" */
+		if (strrchr(str, ':') + 5 == strrchr(str, '>'))
+			return 0;
+	}
+
+	return -1;
+}
+
+static int cli_control(struct switchtec_dev *dev, const char *str)
+{
+	int ret;
+	char rtn[1024];
+	struct switchtec_uart *udev = to_switchtec_uart(dev);
+
+	ret = send_cmd(udev->fd, str, 0);
+	if (ret)
+		return ret;
+
+	ret =  read_resp_line(udev->fd, rtn);
+	if (ret)
+		return ret;
+
+	return 0;
+}
+
+#ifdef __CHECKER__
+#define __force __attribute__((force))
+#else
+#define __force
+#endif
+
+static void uart_close(struct switchtec_dev *dev)
+{
+	struct switchtec_uart *udev =  to_switchtec_uart(dev);
+	cli_control(dev, "echo 1\r");
+
+	if (dev->gas_map)
+		munmap((void __force *)dev->gas_map,
+		      dev->gas_map_size);
+
+	flock(udev->fd, LOCK_UN);
+	close(udev->fd);
+	free(udev);
+}
+
+static int map_gas(struct switchtec_dev *dev)
+{
+	void *addr;
+	dev->gas_map_size = 4 << 20;
+
+	addr = mmap(NULL, dev->gas_map_size,
+		   PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+	if (addr == MAP_FAILED)
+		return -1;
+
+	dev->gas_map = (gasptr_t __force)addr;
+
+	return 0;
+}
+
+#undef __force
+
+static gasptr_t uart_gas_map(struct switchtec_dev *dev, int writeable,
+			    size_t *map_size)
+{
+	if (map_size)
+		*map_size = dev->gas_map_size;
+
+	return dev->gas_map;
+}
+
+static void uart_gas_read(struct switchtec_dev *dev, void *dest,
+				const void __gas *src, size_t n)
+{
+	int ret;
+	int raddr, rnum, rcrc;
+	int i, j;
+	char *pos;
+	struct switchtec_uart *udev = to_switchtec_uart(dev);
+	uint32_t addr = (uint32_t)(src - (void __gas *)dev->gas_map);
+	uint8_t *ptr = dest;
+	uint8_t cal;
+	char gas_rd_rtn[4096];
+
+	for (i = 0; i< RETRY_NUM; i++) {
+		ret =  send_cmd(udev->fd, "gasrd -c -s 0x%x %zu\r", 0, addr, n);
+		if (ret)
+			continue;
+
+		ret = read_resp_line(udev->fd, gas_rd_rtn);
+		if (ret)
+			continue;
+
+		if(sscanf(gas_rd_rtn,
+			 "%*[^<]<0x%x> [%d Byte]%*[^:]: 0x%x%*[^:]:",
+				  &raddr, &rnum, &rcrc) != 3)
+			continue;
+
+		if ((raddr != addr) || (rnum != n))
+			continue;
+
+		pos = strchr(gas_rd_rtn, ']');
+		pos += 2;
+		for (j = 0; j < n; j++)
+			*ptr++ = strtol(pos, &pos, 16);
+
+		addr = htobe32(addr);
+		cal = crc8((uint8_t *)&addr, sizeof(addr), 0, true);
+		cal = crc8(dest, n, cal, false);
+		if (cal == rcrc)
+			break;
+	}
+
+	if (i == RETRY_NUM)
+		raise(SIGBUS);
+}
+
+static void uart_memcpy_from_gas(struct switchtec_dev *dev, void *dest,
+				const void __gas *src, size_t n)
+{
+	ssize_t cnt;
+
+	while(n) {
+		cnt = n > UART_MAX_READ_BYTES? UART_MAX_READ_BYTES : n;
+		uart_gas_read(dev, dest, src, cnt);
+		dest += cnt;
+		src += cnt;
+		n -= cnt;
+	}
+}
+
+#define create_gas_read(type, suffix) \
+	static type uart_gas_read ## suffix(struct switchtec_dev *dev, \
+		type __gas *addr) \
+	{ \
+		type ret; \
+		uart_gas_read(dev, &ret, addr, sizeof(ret)); \
+		return ret; \
+	}
+create_gas_read(uint8_t, 8);
+create_gas_read(uint16_t, 16);
+create_gas_read(uint32_t, 32);
+create_gas_read(uint64_t, 64);
+
+static void uart_gas_write(struct switchtec_dev *dev, void __gas *dest,
+			   const void *src, size_t n)
+{
+	int ret;
+	int i;
+	char gas_wr_rtn[4096];
+	uint32_t crc;
+	uint32_t cal, exp;
+	struct switchtec_uart *udev =  to_switchtec_uart(dev);
+	uint32_t addr = (uint32_t)(dest - (void __gas *)dev->gas_map);
+
+	addr = htobe32(addr);
+	crc = crc8((uint8_t *)&addr, sizeof(addr), 0, true);
+	for (i = n; i > 0; i--)
+		crc = crc8((uint8_t *)src + i - 1, sizeof(uint8_t), crc, false);
+
+	addr = htobe32(addr);
+
+	for (i = 0; i< RETRY_NUM; i++) {
+		ret =  send_cmd(udev->fd, "gaswr -c -s 0x%x 0x",
+			       n, addr, src, crc);
+		if (ret)
+			continue;
+
+		ret = read_resp_line(udev->fd, gas_wr_rtn);
+		if (ret)
+			continue;
+
+		if (sscanf(gas_wr_rtn, "%*[^:]: [0x%x/0x%x]%*[^:]:",
+			   &cal, &exp) != 2)
+			continue;
+
+		if ((exp == cal) && (cal == crc))
+			break;
+	}
+
+	if (i == RETRY_NUM)
+		raise(SIGBUS);
+}
+
+static void uart_memcpy_to_gas(struct switchtec_dev *dev, void __gas *dest,
+			      const void *src, size_t n)
+{
+	size_t cnt;
+
+	while(n){
+		cnt = n > UART_MAX_WRITE_BYTES ? UART_MAX_WRITE_BYTES : n;
+		uart_gas_write(dev, dest, src, cnt);
+		dest += cnt;
+		src += cnt;
+		n -= cnt;
+	}
+}
+
+#define create_gas_write(type, suffix) \
+	static void uart_gas_write ## suffix(struct switchtec_dev *dev, \
+					type val, type __gas *addr) \
+	{ \
+		uart_gas_write(dev, addr, &val, sizeof(type)); \
+	}
+
+create_gas_write(uint8_t, 8);
+create_gas_write(uint16_t, 16);
+create_gas_write(uint32_t, 32);
+create_gas_write(uint64_t, 64);
+
+static ssize_t uart_write_from_gas(struct switchtec_dev *dev, int fd,
+				  const void __gas *src, size_t n)
+{
+	ssize_t ret;
+	void *buf;
+
+	buf = malloc(n);
+
+	uart_memcpy_from_gas(dev, buf, src, n);
+	ret = write(fd, buf, n);
+
+	free(buf);
+
+	return ret;
+}
+
+static const struct switchtec_ops uart_ops = {
+	.close = uart_close,
+	.gas_map = uart_gas_map,
+
+	.cmd = gasop_cmd,
+	.get_fw_version = gasop_get_fw_version,
+	.pff_to_port = gasop_pff_to_port,
+	.port_to_pff = gasop_port_to_pff,
+	.flash_part = gasop_flash_part,
+	.event_summary = gasop_event_summary,
+	.event_ctl = gasop_event_ctl,
+	.event_wait_for = gasop_event_wait_for,
+
+	.gas_read8 = uart_gas_read8,
+	.gas_read16 = uart_gas_read16,
+	.gas_read32 = uart_gas_read32,
+	.gas_read64 = uart_gas_read64,
+	.gas_write8 = uart_gas_write8,
+	.gas_write16 = uart_gas_write16,
+	.gas_write32 = uart_gas_write32,
+	.gas_write64 = uart_gas_write64,
+
+	.memcpy_to_gas = uart_memcpy_to_gas,
+	.memcpy_from_gas = uart_memcpy_from_gas,
+	.write_from_gas = uart_write_from_gas,
+};
+
+static int set_uart_attribs(int fd, int speed, int parity)
+{
+	int ret;
+	struct termios uart_attribs;
+	memset(&uart_attribs, 0, sizeof(uart_attribs));
+
+	ret = tcgetattr(fd, &uart_attribs);
+	if (ret)
+		return -1;
+
+	cfsetospeed(&uart_attribs, speed);
+	cfsetispeed(&uart_attribs, speed);
+
+	uart_attribs.c_iflag &= ~IGNBRK;
+	uart_attribs.c_iflag &= ~(IXON | IXOFF | IXANY);
+	uart_attribs.c_lflag = 0;
+	uart_attribs.c_oflag = 0;
+	uart_attribs.c_cflag = (uart_attribs.c_cflag & ~CSIZE) | CS8;
+	uart_attribs.c_cflag |= (CLOCAL | CREAD);
+	uart_attribs.c_cflag &= ~(PARENB | PARODD);
+	uart_attribs.c_cflag |= parity;
+	uart_attribs.c_cflag &= ~CSTOPB;
+	uart_attribs.c_cflag &= ~CRTSCTS;
+	uart_attribs.c_cc[VMIN] = 0;
+	uart_attribs.c_cc[VTIME] = 50;
+
+	ret = tcsetattr(fd, TCSANOW, &uart_attribs);
+	if (ret)
+		return -1;
+
+	return 0;
+}
+
+struct switchtec_dev *switchtec_open_uart(int fd)
+{
+	int ret;
+	struct switchtec_uart *udev;
+
+	udev = malloc(sizeof(*udev));
+	if (!udev)
+		return NULL;
+
+	udev->fd = fd;
+	if (udev->fd < 0)
+		goto err_free;
+
+	ret = flock(udev->fd, LOCK_EX | LOCK_NB);
+	if (ret)
+		goto err_close_free;
+
+	ret = set_uart_attribs(udev->fd, SWITCHTEC_UART_BAUDRATE, 0);
+	if (ret)
+		goto err_close_free;
+
+	ret = cli_control(&udev->dev, "pscdbg 0 all\r");
+	if (ret)
+		goto err_close_free;
+
+	ret = cli_control(&udev->dev, "echo 0\r");
+	if (ret)
+		goto err_close_free;
+
+	if (map_gas(&udev->dev))
+		goto err_close_free;
+
+	udev->dev.ops = &uart_ops;
+	gasop_set_partition_info(&udev->dev);
+	return &udev->dev;
+
+err_close_free:
+	close(udev->fd);
+
+err_free:
+	free(udev);
+	return NULL;
+}
+
+#endif
+
--- a/lib/platform/linux.c
+++ b/lib/platform/linux.c
@@ -776,14 +776,20 @@ static const struct switchtec_ops linux_ops = {
 struct switchtec_dev *switchtec_open_by_path(const char *path)
 {
 	struct switchtec_linux *ldev;
+	int fd;
+
+	fd = open(path, O_RDWR | O_CLOEXEC);
+	if (fd < 0)
+		return NULL;
+
+	if (isatty(fd))
+		return switchtec_open_uart(fd);
 
 	ldev = malloc(sizeof(*ldev));
 	if (!ldev)
 		return NULL;
 
-	ldev->fd = open(path, O_RDWR | O_CLOEXEC);
-	if (ldev->fd < 0)
-		goto err_free;
+	ldev->fd = fd;
 
 	if (check_switchtec_device(ldev))
 		goto err_close_free;
@@ -797,7 +803,6 @@ struct switchtec_dev *switchtec_open_by_path(const char *path)
 
 err_close_free:
 	close(ldev->fd);
-err_free:
 	free(ldev);
 	return NULL;
 }
--- a/lib/platform/windows.c
+++ b/lib/platform/windows.c
@@ -561,4 +561,10 @@ struct switchtec_dev *switchtec_open_i2c_by_adapter(int adapter, int i2c_addr)
 	return NULL;
 }
 
+struct switchtec_dev *switchtec_open_uart(int fd)
+{
+	errno = ENOTSUP;
+	return NULL;
+}
+
 #endif
--- a/lib/switchtec.c
+++ b/lib/switchtec.c
@@ -70,6 +70,7 @@
  *   * An I2C adapter number and slave number (0@0x20)
  *   * An I2C device delimited with a colon (/dev/i2c-1:0x20)
  *     (must start with a / so that it is distinguishable from a BDF)
+ *   * A UART device (/dev/ttyUSB0)
  */
 struct switchtec_dev *switchtec_open(const char *device)
 {
--- a/cli/common.h
+++ b/cli/common.h
@@ -39,6 +39,7 @@ int switchtec_handler(const char *optarg, void *value_addr,
 			" * an index (0, 1, 2)\n" \
 			" * a pci address (3:00.1)\n" \
 			" * An I2C path with slave address (/dev/i2c-1@0x20)\n" \
+			" * A UART path (/dev/ttyUSB0)\n" \
 	}
 
 #endif
--- a/inc/switchtec/switchtec.h
+++ b/inc/switchtec/switchtec.h
@@ -222,6 +222,7 @@ struct switchtec_dev *switchtec_open_by_pci_addr(int domain, int bus,
 						 int device, int func);
 struct switchtec_dev *switchtec_open_i2c(const char *path, int i2c_addr);
 struct switchtec_dev *switchtec_open_i2c_by_adapter(int adapter, int i2c_addr);
+struct switchtec_dev *switchtec_open_uart(int fd);
 
 void switchtec_close(struct switchtec_dev *dev);
 int switchtec_list(struct switchtec_device_info **devlist);
--- a/lib/platform/platform.c
+++ b/lib/platform/platform.c
@@ -74,6 +74,14 @@ struct switchtec_dev *switchtec_open_by_pci_addr(int domain, int bus,
 struct switchtec_dev *switchtec_open_i2c(const char *path, int i2c_addr);
 
 /**
+ * @brief Open a switchtec device behind a uart device
+ * @ingroup Device
+ * @param[in] fd	file descriptor to uart device
+ * @return Switchtec device handle, NULL on failure
+ */
+struct switchtec_dev *switchtec_open_uart(int fd);
+
+/**
  * @brief Close a Switchtec device handle
  * @ingroup Device
  * @param[in] dev	Switchtec device handle to close


lib/platform/linux-uart.c
```diff
@@ -0,0 +1,441 @@
...
+struct switchtec_dev *switchtec_open_uart(const char *path)
+{
+	int ret;
+	struct switchtec_uart *udev;
+	
+	udev = malloc(sizeof(*udev));
+	if (!udev)
+		return NULL;
+
+	udev->fd = open(path, O_RDWR | O_NOCTTY);
```
lsgunth: Might be a good idea to take an exclusive lock on the file as well, using flock. Then multiple switchtec calls (which are relatively safe otherwise) don't really screw each other up.

lib/switchtec.c
```diff
@@ -80,6 +81,11 @@ struct switchtec_dev *switchtec_open(const char *device)
 	char *endptr;
 	struct switchtec_dev *ret;
 
+	if (sscanf(device, "/dev/tty%*[^0-9]%d", &idx) == 1) {
```
lsgunth: This is a very poor way to tell if you have been handed a tty.

I'd hook into switchtec_open_by_path() and call isatty() on the fd and split off from there if it is.

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,441 @@
...
+
+	ret = tcflush(udev->fd,TCIOFLUSH);
+	if (ret < 0)
+		return ret;
+
+	ret = write(udev->fd, str, strlen(str));
+	if (ret != strlen(str))
+		return -1;
+
+	ret =  tcdrain(udev->fd);
```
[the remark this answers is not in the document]
wesleywesley: Yes, the uart device opened with "BLOCK" (without flag: O_NONBLOCK).

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,441 @@
...
+	if (ret < 0)
+		return ret;
+
+	memset(rtn, 0, sizeof(rtn));
+
+cli_rtn_retry:
+	if (rty_cnt > RETRY_NUM)
+		return -1;
+
+	ret = read(udev->fd, rtn, sizeof(rtn));
```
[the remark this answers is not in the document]
wesleywesley: Rely on specific symbol appeared in the response string.
For example: 
by sscanf(rtn, "%*[^:]:%x>", &idx), check the ":" in the string.
Following are cli commands like echo and debug print off:
 
0x00000000:121c>echo 1
Echo - ON
0x00000000:121d>echo 0
Echo - OFF
0x00000000:121e>
Echo - ON
0x00000000:121f>pscdbg 0 all
Mask before 00000000
Mask after 00000000
0x00000000:1220>

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,441 @@
...
+
+	ret = write(udev->fd, str, strlen(str));
+	if (ret != strlen(str))
+		return -1;
+
+	ret =  tcdrain(udev->fd);
+	if (ret < 0)
+		return ret;
+
+	memset(rtn, 0, sizeof(rtn));
```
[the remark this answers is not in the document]
wesleywesley: It seems that it is the first time to set all zero to the eniter buffer.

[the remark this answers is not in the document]
wesleywesley: So the first is by compiler and the second is by this explicit memset?

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,441 @@
...
+
+static int cli_control(struct switchtec_dev *dev, const char *str)
+{
+	int ret;
+	int idx;
+	int rty_cnt = 0;
+	char rtn[1024];
+	struct switchtec_uart *udev = to_switchtec_uart(dev);
+
+	ret = tcflush(udev->fd,TCIOFLUSH);
```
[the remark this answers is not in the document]
wesleywesley: tcflush used to flushing both the out and in buffer, with TCIOFLUSH flag set.
yes, there maybe chars between flush and write, but the difficult is the chars from FW is a superset, utility code could not process all the cases.
Here use the tcflush, is mainly for flush the ">" symbol if exist, so the following check of the ">" symbol, if it appear, we consider, the request's respose is finished.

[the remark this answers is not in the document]
lsgunth: Yes, as I said, you are much better reading that data back than flushing. It's much better if you parse it all. It kind of looks like there's a counter in the prompt and if you can use it to ensure you get right prompt for the command you sent that would be much better.

lib/platform/linux-uart.c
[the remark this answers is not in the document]
lsgunth: I understand that but it doesn't really change things. On a closer look though it looks like you are using an inter-byte time out which shouldn't be as slow as I thought. However, you should definitely put the read in a loop until you see the end of the data you are looking for.

lib/platform/linux-uart.c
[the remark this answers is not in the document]
lsgunth: It's set first by memset and second by read. No need to set the bytes twice.

lib/platform/linux-uart.c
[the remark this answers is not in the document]
wesleywesley: Hi, Logan,

The code is put the read in a loop(for gasrd/gaswr/cli_control), until see the end of the data looking for.
Where counter is used to avoid unlimited loop in case of the low probability case: the specific symbol losted for the reason of a bad link.

************************************
preseduo code:

retry the read:
               if (retry counter saturated)
                             exit(1);
               
               return str = read();

               check the if the specific symbol exists in the "return str"
               Yes: go on
               No: loop to "retry the read"

************************************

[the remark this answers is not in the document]
lsgunth: Yes, but what happens if you get the first part of the prompt in one read and the second part of the prompt in another read? The scanf is going to fail both times even though you got the data back you expected.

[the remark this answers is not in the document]
wesleywesley: For different cli commands, the output(lines, patterns, symbols) is different.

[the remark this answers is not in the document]
lsgunth: Yes, so the point is you probably want to read line by line until you find a prompt. Not try to fit all possible output into 1024 characters and hope there's a prompt by the end of it....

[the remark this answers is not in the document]
wesleywesley: For cli_control, utility does not care the output, but just the end of prompt.
But for gasrd/gaswr, utility firstly care if the end of prompt comes, if so, then abstract the need data from the buffer which collect entire data, such as the crc and data payload for gasrd.
This is why I chose read all data instead of line by line.

lib/platform/linux-uart.c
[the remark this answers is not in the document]
wesleywesley: We just care about the end of the prompt, instead of the ensure the counter, because
1. if the counter return from FW is correct, but for the reaon of the link is bad the counter is corrpted, this will casue unnecessary retry
2. CRC only protect the offset and payload. 
to take counter sequence into account, it make the thing heavy, but not necessary
3. To track the counter, a global variable is introduce, because the counter is  for all the commands: cli_control, gasrd, gaswr.

lib/switchtec.c
wesleywesley: add isatty check in switchtec_open_uart

lib/platform/linux-uart.c
wesleywesley: add flock (LOCK_EX) in switchtec_open_uart and flock(LOCK_UN) in uart_close

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,441 @@
...
+		exit(1);
+
+	ret = write(udev->fd, gas_rd, strlen(gas_rd));
+	if (ret != strlen(gas_rd))
+		exit(1);
+
+	ret = tcdrain(udev->fd);
+	if (ret < 0)
+		exit(1);
+
```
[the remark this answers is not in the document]
wesleywesley: delete all the tcdrain in this file

lib/platform/linux-uart.c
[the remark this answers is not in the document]
wesleywesley: delete the tcdrain

lib/platform/linux-uart.c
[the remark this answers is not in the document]
wesleywesley: the read is accumulative, every time, append the read data to the previously read results.
after the memset with 0, it is easy to use strlen function, for the append.
if append a zero to the end of the read data, a counter need to implement to remember the previous read data length. make thing complex.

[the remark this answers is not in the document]
lsgunth: I'd say a counter makes more sense it makes things pretty clear what's going on. It's also faster than scanning the string with strlen every time.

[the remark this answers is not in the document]
wesleywesley: ok, I will change to use the counter instead of memset+strelen.

wesleywesley:
Hi, Logan,

Thank you for the suggestions.
TWI branch is almost ready for merge into devel.

For this pull request, if any suggestions?

regards,
Wesley

lsgunth:
Can you merge the I2C stuff then rebase this branch? Then ping me and I'll take another look.

Logan

wesleywesley:
Merged i2c_dev branch to uart_dev_stream.

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,437 @@
...
+	struct switchtec_uart *udev = to_switchtec_uart(dev);
+	uint32_t addr = (uint32_t)(src - (void __gas *)dev->gas_map);
+
+	b += snprintf(b, sizeof(gas_rd), "gasrd -c -s 0x%x %zu\r", addr, n);
+
+rd_crc_err:
+	if (crc_rty_cnt > RETRY_NUM)
+		exit(1);
+
+	ret = write(udev->fd, gas_rd, strlen(gas_rd));
```
lsgunth: I still want to separate out the common parts of cli_control, uart_memcpy_from_gas and uart_gas_write. Probably into a send_cmd() and a read_resp_line() or something similar. Right now the three functions are all very long, complicated, doing very similar things and are hard to review for correctness. More functions and loops, less goto mess.

lib/switchtec.c
lsgunth: Not good enough. If I want to specify a device file that doesn't start with "/dev/tty" your code will prevent me from doing so. There's lots of good reasons to do so too, like using socat to forward a TTY over the network or udev rules to symlink the tty device to a more useful name, etc.

lib/platform/windows.c
```diff
@@ -561,4 +561,11 @@ struct switchtec_dev *switchtec_open_i2c_by_adapter(int adapter, int i2c_addr)
 	return NULL;
 }
 
+struct switchtec_dev *switchtec_open_uart(const char *path)
+{
+	errno = ENOTSUP;
+	return NULL;
+}
+
+
```
lsgunth: Nit: Avoid extra lines in a few places.... (kernel style)

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,437 @@
...
+	int rty_cnt = 0;
+	char rtn[1024];
+	int cnt = 0;
+	struct switchtec_uart *udev = to_switchtec_uart(dev);
+
+	ret = write(udev->fd, str, strlen(str));
+	if (ret != strlen(str))
+		return -1;
+
+cli_rtn_retry:
```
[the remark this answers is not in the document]
wesleywesley: replace end of response from device check by loop instead of goto

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,437 @@
...
+
+	ret = read(udev->fd, rtn + cnt, sizeof(rtn));
+	if (ret < 0)
+		return ret;
+
+	cnt += ret;
+	rtn[cnt] = '\0';
+
+	if (sscanf(rtn, "%*[^:]:%x>", &idx) != 1){
+		usleep(5000);
```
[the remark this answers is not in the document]
wesleywesley: Actually, current usleep(5000) and RETRY_NUM (now 50) is a combination results after adjusting/tuning with real case.
For example for fw-toggle, which take very long time waiting for the response for the toggle MRPC command.
If remove the usleep, then the RETRY_NUM will increased proportionly, and more syscall read will invoked.
In real case, once the usleep is called, it means that the response is not fully collect from device, it is a punishment for this utility, to decrease its calling of syscall read.

lib/platform/linux-uart.c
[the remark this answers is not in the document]
wesleywesley: Keep the goto for crc error retry.

lib/platform/linux-uart.c
wesleywesley: I have this thoughts before.
But actually, for send_cmd(), it is mainly to invoke syscall write,  just with different input str of cli cmd.

And for read_resp_line(), it is for syscall read.
Besides these two same things, they have different scan template and expected match numbers.
Which will introduce variable length parameters funs, make thing complicated with limited gain.

lib/platform/windows.c
wesleywesley: fixed

lib/switchtec.c
wesleywesley: Yes, it will only process with /dev/tty started.
A question, in socat case or symlink case, will they start with /dev/ for any arbitrary string?

lib/platform/linux-uart.c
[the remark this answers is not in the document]
lsgunth: Then it would be better to remove the usleep and increase the timeout on the read command to match.

lib/platform/linux-uart.c
lsgunth: I was actually thinking a variable length parameter function would be a great idea for send_cmd().

read_resp_line() would read until the next prompt, encompassing the waiting for a full line. The calling function can then parse the prompt as it does now.

Not repeating yourself is a huge gain and your existing code is already too complicated..

lib/platform/linux-uart.c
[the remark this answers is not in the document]
lsgunth: I think you'll find once you refactor your code as I ask you can make a loop out of that retry too.

You should never use gotos for loops. Refactor your code as I've asked so you don't need the spaghetti gotos. It's already starting to look more readable but could use more.

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,438 @@
...
+	ret = write(udev->fd, str, strlen(str));
+	if (ret != strlen(str))
+		return -1;
+
+	for (i = 0; i< RETRY_NUM; i++) {
+		ret = read(udev->fd, rtn + cnt, sizeof(rtn));
+		if (ret < 0)
+			return ret;
+
+		cnt =+ ret;
```
lsgunth: Certainly you mean +=

lib/switchtec.c
lsgunth: No, they can be any path and matching on a path is a bad idea. As I said the first time, hook into switchtec_open_by_path(), once you have an fd call isatty() and if it is a tty call switchtec_open_uart(). Parsing the path in any way is a bad idea.

lib/platform/linux-uart.c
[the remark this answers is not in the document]
wesleywesley: I guess you would like FW to implement the GAS READ COMMD "gasrd -c -s <offset> [byte count]" with an parameter "-t timeout", right?
Current FW have no this timeout parameter for gasrd cmd.

lib/platform/linux-uart.c
wesleywesley: sorry, what i think is:
For send_cmd it is fix lenght parameter function, with fd, str for send.
For response read and check function, it is variable length parameter function, which need different variable for the sscanf output.

For send_cmd, it is just a wrapper for system call write
For read, it is a wrapper for system call read and sscanf.
I still feel it will make simple thing more complex(especially variable length param function).
 
After the command, FW keep send the response, SW determine if the end is comming by check several pattern combination in the entire response string.
Current code is read the entire respone, instead of line by line.

lib/platform/linux-uart.c
wesleywesley: thx

lib/platform/linux-uart.c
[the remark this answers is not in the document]
wesleywesley: The first goto is already replaced with loop, cause it is a normal case, wait for the response from FW till finished, once finished, break the loop.
But the second goto is for  the crc error case, instead a normal case loop.

Actually, I find replace the second goto with loop, will make the code unreadable.

lib/switchtec.c
wesleywesley: changed, pls comments.

lib/platform/linux-uart.c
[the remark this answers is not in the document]
lsgunth: No, that doesn't seem necessary. Just make sure the timeout has a sensible value.

Also, the FW has nothing to do with the timeout.  The timeout is a feature of the OS (ie. how long read() will wait with no data before returning).

lib/platform/linux-uart.c
lsgunth: read_resp_line() should just read an entire prompt (ie everything up to the '>') the caller can then do subsequent parsing. Alternatively, you could pass in a different function to parse each line. This is more than a wrapper around read and sscanf as it should handle getting partial reads back.

send_cmd() can take vargs so you can do something like:

```ret = send_cmd(dev, "gasrd -c -s 0x%x %zu\r", addr, n);```

It will wrap the sprintf and the write nicely.

Your code is already too complex using a huge **mess** of gotos. I guarantee if you do it right it will be simpler and easier to read and **much** easier to maintain and expand upon.

File: lib/platform/linux-uart.c (hunk 10 above)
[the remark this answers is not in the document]
Comment (lsgunth) [reply lib/platform/linux-uart.c]: You're wrong. Only because it's currently a mess does it look better with the goto. Goto's should simply *never* be used where a loop primitive can be used.

File: lib/switchtec.c [hunk 13]
```diff
@@ -93,6 +94,10 @@ struct switchtec_dev *switchtec_open(const char *device)
 
 	if (strchr(device, '/') || strchr(device, '\\')) {
 		ret = switchtec_open_by_path(device);
+		if (ret)
+			goto found;
+
+		ret = switchtec_open_uart(device);
```
lsgunth: I'd very much prefer the way I told you to do it twice. Inside switchtec_open_by_path(), when you have already opened the file,  check isatty() and, if it is, then branch off into the uart code. This way we don't close the file and reopen or return an error when it's not an error.

lib/platform/linux-uart.c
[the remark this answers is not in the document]
wesleywesley: Hi, Logan,

Actually, we have following set for the termios:
 
+	uart_attribs.c_cc[VMIN] = 0;  
+	uart_attribs.c_cc[VTIME] = 5; 

MIN == 0; TIME > 0: TIME specifies the limit for a timer in tenths of a second. The timer is started when read(2) is called. read(2) returns either when at least one byte of data is available, or when the timer expires. If the timer expires without any input becoming available, read(2) returns 0. 

But it does not help for our case:
because FW response to the gasrd/wr command in the form of strings, one by one.
from timeout aspect, it doesn't timeout, because FW keep sending strings.
but from the entire response from fw, maybe the response with long payload, need several read+usleep to collect all the response.

lib/switchtec.c
wesleywesley: so for uart platform, it need call "switchtec_open_by_path", which is a function defined in Linux.c file?

lib/platform/linux-uart.c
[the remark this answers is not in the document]
lsgunth: Ok, that's what I thought. But then what you are saying about needing the usleep() still makes no sense. Waiting for 5ms + 0.5s vs only 0.5s shouldn't change anything significantly as you're affecting the delay between reads by 1%.

The things you talk about the firmware also makes no sense. You keep reading lines until one times out or you've read all expected data (you may also want a longer timeout as it should only timeout when something goes wrong). Sleeping between the reads doesn't help and you probably don't even need a retry counter.  If it times out waiting for data you expect, that should signal something bad happened and there's a problem, no need to do additional read() calls.

lib/switchtec.c
lsgunth: It already calls switchtec_open_by_path()... You need to change that function. (Which, yes, is in linux.c).

lib/platform/linux-uart.c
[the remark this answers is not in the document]
wesleywesley: Sleeping between the reads doesn't help and you probably don't even need a retry counter. 

sleeping between the reads is reduce the read syscall.
the point is, when fw response in flight, the read will not timeout, remove the usleep, result in more syscall read needed.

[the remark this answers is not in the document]
wesleywesley: Sleeping between the reads doesn't help and you probably don't even need a retry counter.
sleeping between the reads is reduce the read syscall.
the point is, when fw response in flight, the read will not timeout, remove the usleep, result in more syscall read needed.

lib/switchtec.c
wesleywesley: change the function, pls comments.

lib/platform/linux-uart.c
wesleywesley: add the read_resp_line function. pls comments.
will add send_cmd function later.

lib/platform/linux-uart.c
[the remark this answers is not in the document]
lsgunth: That makes absolutely no sense.

Your are either waiting wastefully in a sleep syscall or productively in a read syscall. The read call will block until there's something useful to return. So the extra sleep is unnecessary.

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,423 @@
...
+		ret = read(fd, str + cnt, sizeof(str));
+		if (ret < 0)
+			return ret;
+	
+		cnt += ret;
+		str[cnt] = '\0';
+	
+		if (strrchr(str, '>') && strrchr(str, ':') &&
+			(strrchr(str, ':') + 5 == strrchr(str, '>'))){
+			return 0;
```
lsgunth: I think a scanf here would be clearer... At the very least calling identical strrchr()s twice is wasteful...

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,423 @@
...
+
+	ret = write(udev->fd, str, strlen(str));
+	if (ret != strlen(str))
+		return -1;
+
+	ret =  read_resp_line(udev->fd, rtn);
+	if (ret)
+		return ret;
+		
+
```
lsgunth: Nit: too many spaces

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,423 @@
...
+	b += snprintf(b, sizeof(gas_rd), "gasrd -c -s 0x%x %zu\r", addr, n);
+
+rd_crc_err:
+	if (crc_rty_cnt > RETRY_NUM)
+		exit(1);
+
+	ret = write(udev->fd, gas_rd, strlen(gas_rd));
+	if (ret != strlen(gas_rd))
+		exit(1);
```
lsgunth: Gotta get rid of these exits() again

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,423 @@
...
+	ret = read_resp_line(udev->fd, gas_rd_rtn);
+	if (ret)
+		exit(1);
+
+	if(sscanf(gas_rd_rtn, "%*[^<]<0x%x> [%d Byte]%*[^:]: 0x%x%*[^:]:%x>",
+			  &raddr, &rnum, &rcrc, &idx) != 4)
+		exit(1);
+
+	assert(raddr == addr);
+	assert(rnum == n);
```
lsgunth: Asserts are also bad as they will kill the program in the same way as the exits

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,423 @@
...
+		tok = strtok(NULL, " ");
+		*crc_ptr = strtol(tok, &tok, 16);
+		crc_ptr += sizeof(uint8_t);
+	}
+
+	crc_cnt = sizeof(crc_addr) + n;
+	crc_cal = crc8(crc_buf, crc_cnt, 0, true);
+	if (crc_cal != rcrc){
+		crc_rty_cnt++;
+		goto rd_crc_err;
```
lsgunth: Still need to get rid of this goto false loop.

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,423 @@
...
+		exit(1);
+
+	if(sscanf(gas_rd_rtn, "%*[^<]<0x%x> [%d Byte]%*[^:]: 0x%x%*[^:]:%x>",
+			  &raddr, &rnum, &rcrc, &idx) != 4)
+		exit(1);
+
+	assert(raddr == addr);
+	assert(rnum == n);
+
+	crc_addr = htobe32(addr);
```
lsgunth: Do we really need a second copy of addr?

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,423 @@
...
+	int ret, raddr, rnum, rcrc, idx;
+	int i;
+	int crc_rty_cnt = 0;
+	char *tok;
+	const char *delims = "]";
+	char gas_rd[1024];
+	char gas_rd_rtn[102400];
+	char *b = gas_rd;
+	uint8_t crc_cal;
+	uint8_t crc_buf[2*n];
```
lsgunth: Why 2*n? Shouldn't it be n + sizeof(addr)?

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,423 @@
...
+		exit(1);
+
+	assert(raddr == addr);
+	assert(rnum == n);
+
+	crc_addr = htobe32(addr);
+	memcpy(crc_ptr, &crc_addr, sizeof(crc_addr));
+	crc_ptr += sizeof(addr);
+
+	tok = strtok(gas_rd_rtn,delims);
```
lsgunth: Why do we have 'delims' when it's not used a second time? Inline it.

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,423 @@
...
+	for (i = 0; i<rnum; i++){
+		tok = strtok(NULL, " ");
+		*crc_ptr = strtol(tok, &tok, 16);
```
lsgunth: This is confusing. You overwrite tok with the endptr in strtol, but it's never used again (strtok does not use the value of tok to continue parsing on, it uses a stored value).

It also seems like strtok isn't really useful here....

Use strchr to find the first "[' then use strtol to parse each number (it ignores spaces and returns the end of the parsed value).

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,423 @@
...
+		crc_ptr += sizeof(uint8_t);
+	}
+
+	crc_cnt = sizeof(crc_addr) + n;
```
lsgunth: Why do we need a separate crc_cnt variable? It's used once.

Part of what scares me about this function is the sheer number of variables it defines. It's hard to keep it all in your head. It may make sense to separate the parsing out into it's own static function too to try and make this more readable.

wesleywesley:
Hi, Logan,

Updated the code, pls comments.

regards,
Wesley

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,432 @@
...
+
+enum switchtec_uart_cmd_type {
+	SWITCHTEC_UART_CONTROL = 0,
+	SWITCHTEC_UART_GAS_READ = 1,
+	SWITCHTEC_UART_GAS_WRITE = 2,
+};
+
+static int send_cmd(int fd, const char *str, enum switchtec_uart_cmd_type type,
+		    uint32_t addr, size_t n, const void *src, uint8_t crc)
+{
```
lsgunth: Try something like this. Having special cases in a function like this is not very good.

```c
#include <stdio.h>
#include <unistd.h>
#include <stdint.h>
#include <errno.h>

#include <stdarg.h>

/**
 * @brief Send a formatted command out to the UART device
 * @param[in] fd			File descriptor to send the command to
 * @param[in] fmt			Format string passed to sprintf
 * @param[in] write_bytes   Number of bytes in an array to append to the string
 * @param[in] *format_args  Arguments passed to sprintf
 * @param[in] write_data    If write_bytes is non-zero, a pointer to
 *                          data of size \p write_bytes to append to
 *                          the string
 * @param[in] crc           If write_bytes is non-zero, an extra CRC byte
 *                          to append to the data
 * @return 0 on success, negative on error
 */
static int send_cmd(int fd, const char *fmt, int write_bytes, ...)
{
    char cmd[1024];
    int cnt;
    uint8_t *write_data;
    int i;
    int ret;
    va_list argp;

    va_start(argp, write_bytes);

    cnt = vsnprintf(cmd, sizeof(cmd), fmt, argp);

    if (write_bytes) {
        write_data = va_arg(argp, uint8_t *);
        for (i = 0; i < write_bytes; i++) {
            cnt += snprintf(cmd + cnt, sizeof(cmd) - cnt,
                            "%02x", write_data[i]);
        }

        //Extra arg for the CRC
        cnt += snprintf(cmd + cnt, sizeof(cmd) - cnt,
                        "%02x\n", va_arg(argp, int));

    }

    va_end(argp);

    ret = write(fd, cmd, cnt);
    if (ret < 0)
        return ret;

    if (ret != cnt) {
        errno = EIO;
        return -errno;
    }

    return 0;
}

void main(void)
{
    unsigned long addr = 0xAABBCCDD;
    uint8_t data[] = {0xEE, 0xFF, 0x11, 0x22, 0x33};
    uint8_t crc = 0x44;

    send_cmd(STDOUT_FILENO, "control command\n", 0);
    send_cmd(STDOUT_FILENO, "gaswr -c -s 0x%x 0x", sizeof(data),
             addr, data, crc);
    send_cmd(STDOUT_FILENO, "gasrd -c -s 0x%x %zu\n", 0, addr, 5);
}

```

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,432 @@
...
+		return 0;
+}
+
+static int read_resp_line(int fd, char *str)
+{
+	int ret;
+	int i;
+	int cnt = 0;
+
+	for (i = 0; i < RETRY_NUM; i++) {
```
lsgunth: So RETRY_NUM is 10000 and the timeout on read() is 0.5s. So if the device doesn't respond it takes 5000s (more than an hour) before this function returns with a failure.

I think you should have a while(1) and then break if read() ever returns 0. Then increase the timeout to maybe 5s. That way if we don't receive anything for the timeout period we return failure. Otherwise if the response is large and divided into 1000 chunks (unlikely) the function will still be correct.

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,432 @@
...
+		ret = read(fd, str + cnt, sizeof(str));
+		if (ret < 0)
+			return ret;
+	
+		cnt += ret;
+		str[cnt] = '\0';
+	
+		if (strrchr(str, ':') + 5 == strrchr(str, '>'))
```
lsgunth: Yes, this is clear. Thanks.

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,432 @@
...
+#include <assert.h>
+#include <string.h>
+
+#include <sys/file.h>
+#include <termios.h>
+#include "../crc8.h"
+
+static const char *dbginfo_off = "pscdbg 0 all\r";
+static const char *echo_off = "echo 0\r";
+static const char *echo_on = "echo 1\r";
```
lsgunth: Please inline these. They are only used once and the extra level of indirection isn't informative.

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,432 @@
...
+		pos = strchr(gas_rd_rtn, ']');
+		pos += 2;
+		for (j = 0; j < n; j++)
+			*ptr++ = strtol(pos, &pos, 16);
+
+		cal = crc8(buf, sizeof(addr) + n, 0, true);
+		if (cal == rcrc){
+			memcpy(dest, buf + sizeof(addr), n);
+			break;
+		}
```
lsgunth: This is much easier to read, thanks.

However, one question: it looks like the only way to go through the rety loop is to fail the CRC. Every other type of failure causes a SIGBUS. So why is the CRC special. 

I think either a CRC should raise a SIGBUS (just like the other errors) and get rid of the loop; or every error should retry and we should raise a SIGBUS after at most 3 retries (not 10000).

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,432 @@
...
+		ptr += sizeof(addr);
+
+		pos = strchr(gas_rd_rtn, ']');
+		pos += 2;
+		for (j = 0; j < n; j++)
+			*ptr++ = strtol(pos, &pos, 16);
+
+		cal = crc8(buf, sizeof(addr) + n, 0, true);
+		if (cal == rcrc){
+			memcpy(dest, buf + sizeof(addr), n);
```
lsgunth: Why don't we just put the data directly into _dest_ instead of allocating another buffer and copying it?

You can always do two crc calls such as:

```c
crc = crc8(&addr, sizeof(addr), 0, true)
crc = crc8(dest, n, crc, false);
```

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,432 @@
...
+	uint32_t addr = (uint32_t)(dest - (void __gas *)dev->gas_map);
+	uint8_t buf[n+sizeof(addr)];
+
+	addr = htobe32(addr);
+	memcpy(buf, &addr, sizeof(addr));
+	for (i = 0; i< n; i++){
+		memcpy(buf + sizeof(addr) + i, (uint8_t *)src + n - 1 - i, sizeof(uint8_t));
+	}
+	crc = crc8(buf, sizeof(addr)+n, 0, true);
+
```
lsgunth: Why is the CRC calculated in reverse data order?

Probably no sense using memcpy() to copy a single byte.  `buf[i] = ` makes more sense.

I think the best solution (if reverse CRC is a thing with this hardware) would be to write a crc8r function which scans the buffer in reverse. Then you can do:

```c
crc = crc8(&addr, sizeof(addr), 0, true);
crc = crc8r(src, n, crc, false)
```

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,432 @@
...
+	udev->fd = fd;
+	if (udev->fd < 0)
+		goto err_free;
+
+	ret = flock(udev->fd, LOCK_EX);
+	if (ret)
+		goto err_close_free;
+
+	ret = set_uart_attribs(udev->fd, B230400, 0);
```
JoeyZhang-Microsemi: User may use different boudrate. It is better to define a micro in switchtec.h

lib/platform/linux-uart.c
wesleywesley: for gaswr -c -s <addr> <data string> <crc>, the crc calculation is protocol between utility and fw, namely start from MSB to LSB of both addr and data string.

reverse version crc8 is not needed, such as:
crc = crc8(&addr, sizeof(addr), 0, true);
for (i=n; i>0; i--)
       crc = crc8(src +i -1, sizeof(uint8_t), crc, false);

lib/platform/linux-uart.c
wesleywesley: thanks, crc8 have this feature for calculation based on the last result.

lib/platform/linux-uart.c
wesleywesley: inlined.

lib/platform/linux-uart.c
wesleywesley: change to variable parameters function. thx.

lib/platform/linux.c
```diff
@@ -785,6 +785,9 @@ struct switchtec_dev *switchtec_open_by_path(const char *path)
 	if (ldev->fd < 0)
 		goto err_free;
 
+	if (isatty(ldev->fd) != 0)
+		goto open_terminal;
```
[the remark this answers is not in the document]
wesleywesley: changed.

lib/platform/linux-uart.c
wesleywesley: update some data: (history for RETRY_NUM)
1. first version with 5000us delay, the RETRY_NUM is 100
it is generated by following time consuming mrpc command operations:
a1.firmware toggle take about 5 loops
b1.status/gui/bw, about 17 loops
c1.log-dump/fw read, about 29 loops

2. after remove the 5000us delay, the RETRY_NUM change to 10000, because the loop number collect when test also increased:
a2.firmware toggle take about 150 loops
b2.status/gui/bw, about 1800 loops
c2.log-dump/fw read, about 3200  loops

not dig into why the proportion between b1/c1 and b2/c2 keep almost no change
but a1/b1 and a2/b2 changes a lot. Maybe the first time log number into excel with some error. 

Anyway.  Yes, more than 1 hour is not acceptable.
change to code to 5ms timeout, cause fw-toggle is very time consuming, when read return 0, it may still fw in doing toggle action, event not in respone stage.

lib/platform/linux-uart.c
wesleywesley: for gasrd -c -s <addr> <n>
the crc is calculated by addr from MSB to LSB, and from data string from LSB to MSB.
it its fw implementation.

lib/platform/linux-uart.c
wesleywesley: previously, I think, all except crc error, is link/fw error, while crc error case is link unstable(some bit flip), so different treatment.
But find in all the error cases, even have no retry chance.
So change to all the case have 3 times retry.

lib/platform/linux-uart.c
wesleywesley: to 5s, typo.

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,422 @@
...
+
+static gasptr_t uart_gas_map(struct switchtec_dev *dev, int writeable, size_t *map_size)
+{
+	if (map_size)
+		*map_size = dev->gas_map_size;
+
+	return dev->gas_map;
+}
+
+static void uart_memcpy_from_gas(struct switchtec_dev *dev, void *dest, const void __gas *src, size_t n)
```
lsgunth: Line length should not be more than 80 chas.

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,422 @@
...
+
+		if ((exp == cal) && (cal == crc))
+			break;
+	}
+
+	if (i == RETRY_NUM)
+		raise(SIGBUS);
+}
+
+static void uart_memcpy_to_gas(struct switchtec_dev *dev, void __gas *dest, const void *src, size_t n)
```
lsgunth: Line length should not be more than 80 chars.

lib/platform/linux-uart.c
```diff
@@ -0,0 +1,422 @@
...
+		uart_gas_write(dev, addr, &val, sizeof(type)); \
+	}
+
+create_gas_write(uint8_t, 8);
+create_gas_write(uint16_t, 16);
+create_gas_write(uint32_t, 32);
+create_gas_write(uint64_t, 64);
+
+static ssize_t uart_write_from_gas(struct switchtec_dev *dev, int fd,
+				  const void __gas *src, size_t n)
```
lsgunth: Indent should align with the open parenthesis.

lib/platform/linux.c
```diff
@@ -776,14 +776,20 @@ static const struct switchtec_ops linux_ops = {
 struct switchtec_dev *switchtec_open_by_path(const char *path)
 {
 	struct switchtec_linux *ldev;
+	int fd;
+
+	fd = open(path, O_RDWR | O_CLOEXEC);
+	if (fd < 0)
+		return NULL;
+
+	if (isatty(fd) != 0)
```
lsgunth: `if (isatty(fd))`

wesleywesley:
Hi, Logan,

1. change the flock from block to non-block mode

2. change uart_memcpy_from gas, same as uart_memcpy_to_gas with limitation of bytes of read.
It is found the issue when test "gas dump" when -n with large input.

3. about the crc
for write, it is from the MSB, such as, 0x12345678, it is start from 0x12
for read, it is from the LSB, it is because when received the data, LSB output first by fw
It is just a rule for both fw and sw to generate the same crc value.

4. uart_memecpy_togas worked correct

5. Travis-ci check passed on the branch category
Following fail info by travis-ci is on the pull request category, found pull request #29 have fix it.
*****************************************************************************************
cli/gui.c:449:5: error: conflicting types for ‘gui_main’

 int gui_main(struct switchtec_dev *dev, unsigned all_ports, unsigned reset,

     ^

In file included from cli/gui.c:26:0:

cli/gui.h:30:5: note: previous declaration of ‘gui_main’ was here

 int gui_main(struct switchtec_dev *, unsigned, unsigned, unsigned, int,

     ^

cc1: warning: unrecognized command line option "-Wno-initializer-overrides" [enabled by default]

make: *** [build/cli/gui.o] Error 1

  CC    cli/main.c

  CC    cli/progress.c

  CC    cli/suffix.c

  RC    cli/cli.rc

  CC    examples/temp.c

  LD    examples/temp

make: Target `compile' not remade because of errors.

rm examples/temp.o



The command "./configure ${CONFIG_OPTS} && make -k W=1 ${MAKE_FLAGS}" exited with 2.


Done. Your build exited with 1.

*****************************************************************************************

wesleywesley:
@lsgunth 

Would you pls approved it if no other issues.

regards,
Wesley

lsgunth:
Hmm, the only thing is I don't think you want non-block on the flock. Seeing most operations are relatively short, I think it's more preferable to wait until another caller finishes instead of returning an error.

Logan

wesleywesley:
Hi, Logan,

If non-block mode for flock: two or more users log on the server remotely,  when one of them running the time consuming operation, such as fw-read/fw-update, the other will be prompted with following info(more like a info, instead of a error)
test@testserver:~/switchtec-user$ sudo ~/switchtec-user/switchtec fw-info /dev/ttyUSB0
/dev/ttyUSB0: Resource temporarily unavailable

But if it is block mode, the second user will uncertain if the tool is stuck or something wrong with the server setup, until the first user finished running the time consuming.

Anyway.
We have to create the switchtec-user release candidate today.
I have to merge the code into devel then to master now.

For it should be block or non-block, we can setup a bug-fix branch if necessary.

thank you very much for effort and help on the uart platform development.

regards,
Wesley

PR #22, opened 2018-06-25
Support I2C platform for switchtec-user 

JoeyZhang-Microsemi:


lib/platform/linux-i2c.c
```diff
@@ -162,41 +202,191 @@ static gasptr_t i2c_gas_map(struct switchtec_dev *dev, int writeable,
...
+	do {
+		ret = ioctl(idev->fd, I2C_RDWR, &rwdata);
+		if (ret < 0) {
+			/*
+			 * For now, this is all we can do. If it's a problem we'll
+			 * have to refactor all GAS access functions to return an
+			 * error code.
+			 */
+			perror("Could not communicate with I2C device");
+			exit(1);
```
lsgunth: Don't call perror from the library. Return an error number so the calling code can call perror.

lib/platform/linux-i2c.c
```diff
@@ -48,10 +49,44 @@ struct switchtec_i2c {
...
+#define CMD_GAS_WRITE_WITH_STATUS  0xE8
+#define CMD_GAS_READ  0xE9
+#define CMD_GAS_WRITE  0xEA
+
+#define MAX_RETRY_COUNT  100
+
+#define PEC_BYTE_COUNT  1
+#define DATA_TAIL_BYTE_COUNT  2
+
+#define CAP_MASK  0x80
```
lsgunth: This seems like a very poor name. What is the 8th bit actually mean? I don't think CAP_MASK describes it.

lib/platform/linux-i2c.c
```diff
@@ -48,10 +49,44 @@ struct switchtec_i2c {
...
+#define GAS_TWI_MRPC_ERR  0x20
+
 #define to_switchtec_i2c(d)  \
 	((struct switchtec_i2c *) \
 	 ((char *)d - offsetof(struct switchtec_i2c, dev)))
 
+static uint8_t gas_write_tag = 0;
+
```
lsgunth: This is not a good idea. What happens if code using this library opens two instances of the I2C device? They will both share the tag and weird things might happen.  You should put it in 'struct switchtec_i2c' instead.

lib/platform/linux-i2c.c
```diff
@@ -351,6 +624,15 @@ struct switchtec_dev *switchtec_open_i2c(const char *path, int i2c_addr)
 	if (i2c_set_addr(idev, i2c_addr))
 		goto err_close_free;
 
+	if (i2c_set_timeout(idev, 10))
+		goto err_close_free;
+
+	if (i2c_gas_cap_get(&idev->dev) != CAP_MASK)
+	{
+		printf("Gas capability check fail.\n");
```
lsgunth: Do not print from the library.

JoeyZhang-Microsemi:
Hi Logan,
Thanks for your comments. I split the last commit. Please help to review it.

Regards
Joey

lib/platform/linux-i2c.c
```diff
@@ -26,6 +26,7 @@
 
 #include "../switchtec_priv.h"
 #include "switchtec/switchtec.h"
+#include "../crc8.h"
```
lsgunth: Please put the private headers together first, before the public header.

lib/platform/linux-i2c.c
```diff
@@ -162,6 +177,57 @@ static gasptr_t i2c_gas_map(struct switchtec_dev *dev, int writeable,
...
+			/*
+			 * For now, this is all we can do. If it's a problem we'll
+			 * have to refactor all GAS access functions to return an
+			 * error code.
+			 */
+			return ret;
+		}
+
+		msg_0_pec = i2c_msg_pec(&msgs[0], msgs[0].len, 0, true);
+		pec = i2c_msg_pec(&msgs[1], msgs[1].len - PEC_BYTE_COUNT, msg_0_pec, false);
```
lsgunth: This line looks too long. 80 char line length limit.

lib/platform/linux-i2c.c
```diff
@@ -48,10 +49,24 @@ struct switchtec_i2c {
 	int i2c_addr;
 };
 
+#define CMD_GET_CAP  0xE0
+#define MAX_RETRY_COUNT  100
+#define PEC_BYTE_COUNT  1
+#define GAS_TWI_MODE_MASK  0x80
```
lsgunth: Should this not be called TWI_ENHANCED_MODE?

lib/platform/linux-i2c.c
```diff
@@ -351,6 +417,9 @@ struct switchtec_dev *switchtec_open_i2c(const char *path, int i2c_addr)
 	if (i2c_set_addr(idev, i2c_addr))
 		goto err_close_free;
 
+	if (i2c_gas_cap_get(&idev->dev) != GAS_TWI_MODE_MASK)
+		goto err_close_free;
+
```
lsgunth: So do we not support the legacy TWI mode? If so please not that in the commit message.

lib/platform/linux-i2c.c
```diff
@@ -47,17 +47,33 @@ struct switchtec_i2c {
...
 #define PEC_BYTE_COUNT  1
 #define GAS_TWI_MODE_MASK  0x80
+#define GAS_TWI_MRPC_ERR  0x20
 
 #define to_switchtec_i2c(d)  \
 	((struct switchtec_i2c *) \
 	 ((char *)d - offsetof(struct switchtec_i2c, dev)))
 
+static uint8_t get_tag(struct switchtec_dev *dev)
+{
```
lsgunth: Probably better for this function to take the switchtec_i2c struct and let the caller do the conversion.

lib/platform/linux-i2c.c
```diff
@@ -228,41 +244,134 @@ static uint8_t i2c_gas_cap_get(struct switchtec_dev *dev)
...
+	ret = ioctl(idev->fd, I2C_RDWR, &wdata);
+	if (ret < 0) {
 		/*
 		 * For now, this is all we can do. If it's a problem we'll
 		 * have to refactor all GAS access functions to return an
 		 * error code.
 		 */
-		perror("Could not communicate with I2C device");
-		exit(1);
+		return ret;
```
lsgunth: Forgot to free i2c_data...

lib/platform/linux-i2c.c
```diff
@@ -228,41 +244,134 @@ static uint8_t i2c_gas_cap_get(struct switchtec_dev *dev)
 		return (rx_buf[0] & GAS_TWI_MODE_MASK);
 }
 
-/* One I2C transaction can write a maximum of 28 bytes */
-#define I2C_MAX_WRITE 28
+/* One I2C transaction can write a maximum of 26 bytes */
+#define I2C_MAX_WRITE 24
```
lsgunth: The difference between the comment and the define is confusing. Please explain in the comment.

lib/platform/linux-i2c.c
```diff
@@ -47,17 +47,33 @@ struct switchtec_i2c {
...
 #define to_switchtec_i2c(d)  \
 	((struct switchtec_i2c *) \
 	 ((char *)d - offsetof(struct switchtec_i2c, dev)))
 
+static uint8_t get_tag(struct switchtec_dev *dev)
+{
+	struct switchtec_i2c *idev = to_switchtec_i2c(dev);
+	/* Valid tag is 0x01 ~ 0xff */
+	idev->tag++;
+	if (0 == idev->tag)
```
lsgunth: Style nit:  

"if (!idev->tag)"

lib/platform/linux-i2c.c
```diff
@@ -228,41 +244,134 @@ static uint8_t i2c_gas_cap_get(struct switchtec_dev *dev)
...
+		.msgs = &msg,
+		.nmsgs = 1,
+	};
+
 	struct {
-		uint32_t gas_addr;
+		uint8_t command_code;
+		uint8_t byte_count;
+		uint8_t tag;
+		uint8_t offset[4];
```
lsgunth: Why is this not a uint32_t? If it's a uint32_t, we don't need to memcpy below and can just do an assign (much nicer).

lib/platform/linux-i2c.c
```diff
@@ -228,41 +244,134 @@ static uint8_t i2c_gas_cap_get(struct switchtec_dev *dev)
...
+
+	do {
+		usleep(2000);
+		ret = ioctl(idev->fd, I2C_RDWR, &rwdata);
+		if (ret < 0) {
+			/*
+			 * For now, this is all we can do. If it's a problem we'll
+			 * have to refactor all GAS access functions to return an
+			 * error code.
+			 */
```
lsgunth: This comment doesn't seem accurate... The calling function does it in a loop. The calling function really needs a comment.

lib/platform/linux-i2c.c
```diff
@@ -228,41 +244,134 @@ static uint8_t i2c_gas_cap_get(struct switchtec_dev *dev)
...
+		tag = get_tag(dev);
+		i2c_gas_data_write(dev, dest, src, n, tag);
+		status = i2c_gas_write_status_get(dev, tag);
+		if (status == 0 || status == GAS_TWI_MRPC_ERR)
+			break;
+		retry_count++;
+	}while (retry_count < MAX_RETRY_COUNT);
+
+	if (retry_count == MAX_RETRY_COUNT) {
+		exit(1);
```
lsgunth: I think it would be more appropriate to do:

raise(SIGBUS);

This allows the calling library to handle this case if it wants. Libraries that might exit the program aren't very friendly to use.

lib/platform/linux-i2c.c
```diff
@@ -54,11 +54,13 @@ struct switchtec_i2c {
 #define CMD_GAS_WRITE  0xEA
 #define CMD_GET_WRITE_STATUS  0xE2
 #define CMD_GAS_WRITE_WITH_STATUS  0xE8
+#define CMD_GAS_READ  0xE9
```
lsgunth: Please apply the same issues in the gas_write commit to this one

lib/platform/linux-i2c.c
JoeyZhang-Microsemi: If we use uint32_t, a uint8_t will be inserted between tag and offset because of the data alinement. When we send this data to I2C bus, it will be an incorrect I2C data flow for GAS TWI.

JoeyZhang-Microsemi:
Thanks for the review and comments. I updated the code.

Joey

lib/platform/linux-i2c.c
JoeyZhang-Microsemi: The Switchtec firmware still supports legacy TWI mode, but it is not suggested. So I do not support the legacy TWI mode in switchtec-user. I wrote it in commit message.

lib/platform/linux-i2c.c
lsgunth: Use "_attribute__((packed))" instead to avoid the padding. It will be clearer.

The compiler will also handle the unaligned access for you.

lib/crc8.c
```diff
@@ -0,0 +1,80 @@
...
+
+uint8_t crc8(uint8_t *msg_ptr, uint32_t byte_cnt, uint32_t oldchksum,
+	      bool init)
+{
+	uint32_t  offset;
+	uint8_t   remainder;
+
+	remainder = ((init == true) ? 0 : oldchksum);
+
+	for (offset = 0; offset < byte_cnt; offset++)
```
lsgunth: Style nit:  the open brace goes on the line with the for().

JoeyZhang-Microsemi:
Updated the code. It is clean with the packed attribute on struct. Thanks again.

Joey

lib/crc8.c
```diff
@@ -0,0 +1,79 @@
...
+ * OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ *
+ */
+
+#ifdef __linux__
+
+#include "crc8.h"
+uint8_t crc8_0107_lut[] =
```
lsgunth: I submitted this patchset to Travis and it failed the sparse check on this line. It should be static:

lib/crc8.c:28:9: error: symbol 'crc8_0107_lut' was not declared. Should it be static?

JoeyZhang-Microsemi: Fixed it. Thanks
Joey

PR #23, opened 2018-06-25
Bug Fix - switcthec_status

JoeyZhang-Microsemi:
BUG Description:
Function switchtec_status() uses SWITCHTEC_MAX_PORTS when sending MRPC command MRPC_LNKSTAT. But the firmware just return the status of valid ports. the number of ports may less than the SWITCHTEC_MAX_PORTS. So the data of ports[SWITCHTEC_MAX_PORTS] has some invalid items. 

To get the correct number of ports, we need add more checks. 
(1) valid item of output are continuous.
(2) if the physical port 0 is valid, it must be first item of output.
(3) physical port number = Port ID + Stack ID * MAX_PORTS_PER_STACK

### DIFF
--- a/lib/switchtec.c
+++ b/lib/switchtec.c
@@ -297,8 +297,10 @@ int switchtec_status(struct switchtec_dev *dev,
 
 
 	for (i = 0; i < SWITCHTEC_MAX_PORTS; i++) {
+		if ((ports[i].phys_port_id == 0) && (i != 0))
+			break;
 		if ((ports[i].stk_id >> 4) > SWITCHTEC_MAX_STACKS)
-			continue;
+			break;
 		nr_ports++;
 	}
 
@@ -307,9 +309,6 @@ int switchtec_status(struct switchtec_dev *dev,
 		return -ENOMEM;
 
 	for (i = 0, p = 0; i < SWITCHTEC_MAX_PORTS && p < nr_ports; i++) {
-		if ((ports[i].stk_id >> 4) > SWITCHTEC_MAX_STACKS)
-			continue;
-
 		s[p].port.partition = ports[i].par_id;
 		s[p].port.stack = ports[i].stk_id >> 4;
 		s[p].port.upstream = ports[i].usp_flag;


lsgunth:
What's the point of this? Is there a bug you are fixing? If so it should be described in the commit message.

Is the firmware really returning such garbage in some situations that this convoluted checking is necessary?

JoeyZhang-Microsemi:
Hi Logan,
Thanks for the review. I have updated my commits. Because the function switchtec_status does not know the output length of MRPC command link status, it use the SWITCHTEC_MAX_PORTS. The invalid data would be read. To get the number of ports, we need to add more checks.

Thanks
joey

lsgunth:
You still haven't answered the question. Is there actually a bug? And what is wrong with the existing check to get the number of ports? And is the firmware actually sending such messed up data that such overkill checks are necessary to determine the end of the list? Maybe we should do something smarter and ensure the firmware sends a list terminator that's easier to distinguish.

JoeyZhang-Microsemi:
There is a bug. When we enable DMA in the driver of switchtec (Switchtec-kernel). The switchtec-user reports incorrect physical port in command status and bw. I added some printf statements in switchtec-user and found the switchtec-user did not get correct number of ports. 

I captured the PCIe trace between host and switchtec. The switchtec did not send such messed up data. It was returned by switchtec driver.
For non-DMA mode of switchtec driver, Switchtec firmware initialized the invalid output space with 0xFF, so original check can work.

So we also can fix this issue on switchtec DMA driver, it need to initialize the DMA space before each MRPC command.

lsgunth:
Ah. Thanks for the explanation.

Yes, it sounds like the switchtec kernel DMA code is sending uninitialized buffers back to userspace. That is the real bug and definitely needs to be fixed there. Sending uninitialized data to userspace is a serious security issue that needs to be addressed properly.

JoeyZhang-Microsemi:
Thanks all, We have fixed this issue on switchtec-kernel. I cancel this pull request.

JoeyZhang-Microsemi:
We found this issue again on switchtec-user devel branch. Reopen it.

wesleywesley:
@JoeyZhang-Microsemi 
As demonstrated, the DMA MRPC driver with output buffer initialized to 0xff, worked correct with status command.
Can we close this request?

PR #24, opened 2018-06-25
Bug fix - output overflow of MRPC Get Bandwidth Counter

JoeyZhang-Microsemi:
Because the MRPC input and output share 1k data buffer. There is an
MRPC output data overflow in this MRPC command. This change fixes it.


wesleywesley:
looks ok.

PR #26, opened 2018-07-02
Fix false alarming when fw toggle command

wesleywesley:
Only output error message when fw toggle execution
failed case.
Before the fix, the output is "unknown error" when
success case.

### DIFF
--- a/lib/platform/linux.c
+++ b/lib/platform/linux.c
@@ -61,7 +61,7 @@ struct switchtec_linux {
 
 const char *platform_strerror(void)
 {
-	return "Unknown Error";
+	return "Success";
 }
 
 static int dev_to_sysfs_path(struct switchtec_linux *ldev, const char *suffix,


lsgunth:
Hmm, this was by design: it should say "success" when you toggle the firmware and it does whenever I try it. Is this in windows or something?

wesleywesley:
It was found when Linux test.
In previous version, when errno = 0, perror was called, which return "success"
But in current version, then errno = 0, platform_strerror was called, which return "unknown error".

lsgunth:
Can we still retain some kind of success message in the patch? 

It may also be good to know what's changed that something is setting errno without there being an error. That should probably be fixed too.

wesleywesley:
Hi, Logan,

Does change the platform_strerror from 

const char *platform_strerror(void) 
{
return "Unknown Error"; 
}
 
to

 const char *platform_strerror(void) 
 {
return "Success"; 
} 

fulfill your original for this function?


Regards,
Wesley

wesleywesley:
Hi, Logan,

const char *platform_strerror(void)
{
return "Unknown Error";
}
to
const char *platform_strerror(void)
{
return "Success";
}

is ok?

lsgunth:
Yes, I already approved the patch that does this...

PR #19, opened 2018-06-12
Add payload bandwidth support on switchtec-user command gui and bw. 

JoeyZhang-Microsemi:
Switchtec firmware will support payload bandwidth for performance monitor. We add an option --bw_type for switchtec-user command gui and bw to support the payload bandwidth.

PR #32, opened 2018-07-27
bugfix  - bandwidth incorrect 

JoeyZhang-Microsemi:
Issue: The bw command does not report correct bandwidth data.

This issue is introduced by pull request #19 (Add payload bandwidth support on switchtec-user command gui and bw). 

The bandwidth counter would be reset in max 250ms after received the Set Bandwidth Counter MRPC command. To finish the bandwidth counter reset, a delay is needed before the Get Bandwidth Couner MRPC command.

### DIFF
--- a/cli/gui.c
+++ b/cli/gui.c
@@ -349,6 +349,9 @@ retry:
 	ret = switchtec_bwcntr_set_many(dev, numports, port_ids, bw_type);
 	if (ret < 0)
 		cleanup_and_error("Set bandwidth type");
+	/* switchtec_bwcntr_set_many will reset bandwidth counter and it needs
+	 * about 1s. */
+	sleep(1);
 
 	if (reset)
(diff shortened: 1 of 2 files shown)


lsgunth:
I think a 1s delay is a bit long and would be quite noticeably slow. I wouldn't go longer than 0.5s, but given you say it should be done in a quarter second 0.3s might make sense. 

Also, please add a comment explaining why you are sleeping.

JoeyZhang-Microsemi:
Hi Logan,

I added comments for it. And I tried to sleep 0.3, 0.5 or 0.8, but the bug can not be fixed. the delay time should be about 1s. 

Thanks
Joey

wesleywesley:
@JoeyZhang-Microsemi 
this reset is needed for inband, but not for low speed link like twi and uart

PR #33, opened 2018-07-31
set errno to 0 in "switchtec_open_by_path" for inband

wesleywesley:
for fw-toggle and evcntr-setup command, the corresponding
function invoke "switchtec_perror" w/o ret value check, which
may lead to switchtec-user utility pop up false alarm, because
errno is no 0.

### DIFF
--- a/lib/platform/linux.c
+++ b/lib/platform/linux.c
@@ -781,6 +781,8 @@ struct switchtec_dev *switchtec_open_by_path(const char *path)
 
 	if (isatty(fd))
 		return switchtec_open_uart(fd);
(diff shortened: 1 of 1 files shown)


lsgunth:
I thought we already resolved this before. It should say "success" if errno is not zero which is what we want. If that's still not the case you should investigate what is setting errno and fix that. Or clear errno before the call. (If you do this you better know exactly what set errno in the first place.)

wesleywesley:
@lsgunth 
root cause the first place set errno to non-zero, and fixed.

PR #34, opened 2018-08-01
if errno not equal to 0, keep its value instead of set to ENODEV

wesleywesley:
In switchtec_open funcion, if no device was found, while errno equal
to 0, this will introduce false "success" prompt, so in this case,
set the errno to ENODEV unconditionly.

But if errno already set by system api, its value should be kept
instead of set to ENODEV.

### DIFF
--- a/lib/switchtec.c
+++ b/lib/switchtec.c
@@ -131,7 +131,7 @@ found:
 	if (ret)
 		snprintf(ret->name, sizeof(ret->name), "%s", device);
 	else
-		errno = ENODEV;
+		errno = errno ? errno : ENODEV;
 
 	return ret;
 }


kelvin-cao:
I'd think the return of ENODEV is acceptable as a general failure errno. A given errno may have different meaning in different functions, so pass through an errno might not be a better way. And for an API, it's better to have determined errno set, like 3 errnos with defined meaning.

lsgunth:
I think you need to explain better what you are trying to solve. Based on what the user sees and what you expect to see.

Just stating that errno "should" be some value is not good enough.

wesleywesley:
@lsgunth 
update the commit message

PR #35, opened 2018-08-01
replace default handler for SIGBUS

wesleywesley:
default handler for SIGBUS print message "Bus error (core dump)".
replace the default handler and print message more friendly.

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -36,7 +36,7 @@
 #include <time.h>
 #include <fcntl.h>
 #include <unistd.h>
-
+#include <signal.h>
 #include <errno.h>
 #include <stdio.h>
 
@@ -1924,9 +1924,34 @@ static struct prog_info prog_info = {
 		"(ex: /dev/switchtec0)",
 };
 
+#ifndef _WIN32
+static void sig_handler(int signum)
+{
+	if (signum == SIGBUS) {
+		fprintf(stderr, "Error communicating with the device. "
+			"Please check your setup.\n");
+		exit(1);
+	}
+}
+
+static void setup_sigbus(void)
+{
+	signal(SIGBUS, sig_handler);
+}
+
+#else /* _WIN32 defined */
+
+static void setup_sigbus(void)
+{
+}
+
+#endif
+
 int main(int argc, char **argv)
 {
 	int ret;
+
+	setup_sigbus();
 
 	ret = commands_handle(argc, argv, &prog_info);
 


lsgunth:
NAK. The library should not be messing with the signals. That's up to the calling code. If this is required it should be done in the client.

However, I think that "Bus error" is more informative than the message you replaced it with.

wesleywesley:
@lsgunth 

Frist:, "bus error (core dump)" message will let user consider the sw crashed, this is a unacceptable phenomenon.

actually, this patch was introduce for because in some special case, the response string is not as the negotiated format (for example, one register, which was reserved in some cfg; while in other cfg, it is with real meaning):
******************** ok case ------
gas_reg_read <0x135000> [2 Byte]
f8 11
CRC: 0xa3
0x00000000:000b>
********************
But
******************** problem case -------
gas_reg_read <0x135c00> [4 Byte]
[PFF] cs addr: 0x0300,not hit
00 00 00 00
CRC: 0xb6
0x00000000:000c>
********************

Because when come across the second string pattern, sw will finally exit with raise signal (because sw parse method only work for the first pattern.)

We have to method in sw side (instead of fw) for this specific case:
1. take the 2nd string format into account, sw should also parse it correctly
2. or raise signal when 2nd string appears, but actually, it is not a bus error.

But even we take 2nd format into account, sw still may come across real bus error case, and in that case, the message "bus error core dump" is not acceptable, this is why we have to change the signal handler.

wesleywesley:
@lsgunth 
NAK. The library should not be messing with the signals. That's up to the calling code. If this is required it should be done in the client. 

--- there are 3 levels: command line interface, core lib and platform lib.
I think what you suggest was move the handler fomr the core/platform lib (now in platform lib) to the upper level, such as command line interface?

lsgunth:
There is only one library. It contains platform specific code but it isn't really another level.

Yes, if we do anything with signals it should be in the CLI code.

And my point was: "Bus Error" is far clearer than "Device returns are unexpected".

wesleywesley:
@lsgunth 
change the signal process from  lib to cli.

BTW, when check by travis-ci, some foundings:
1. for signal process, method 1 "signal()" and method 2 "sigaction()"
the second method could not parse travis-ci, so fallback to the first method.
2. on MINGW check, the "SIGBUS" is a unrecognized definition
so differenciate by __WIN32 for signal process
uart platform also only support on Linux os.

cli/main.c
```diff
@@ -1924,10 +1924,35 @@ static struct prog_info prog_info = {
 		"(ex: /dev/switchtec0)",
 };
 
+#ifndef _WIN32
+static void sig_handler(int signum)
+{
+	if (signum == SIGBUS) {
+		fprintf(stderr, "Device returns are unexpected, "
+			"please check the setup.\n");
```
lsgunth: Can we use a better error message. Something like:

"Error communicating with the device. Please check your setup."

lsgunth:
Note: SIGBUS is also used by the I2C code in the same manner.

wesleywesley:
@lsgunth updated, pls approve if no more to fix.

PR #37, opened 2018-08-02
 fw_wait() return original status when bgstaus in error state

wesleywesley:
when test fw-update on header/data crc error files
the fw_write_fd(file) return is not the original fw download status
from mrpc cmd, but "DLSTAT_HARDWARE_ERR" from fw_wait fun.
This introduce the false message of "Hardware error", instead of
the expected "crc incorrect / header incorrect"

### DIFF
--- a/lib/fw.c
+++ b/lib/fw.c
@@ -107,10 +107,13 @@ int switchtec_fw_wait(struct switchtec_dev *dev,
 		ret = switchtec_fw_dlstatus(dev, status, &bgstatus);
 		if (ret < 0)
 			return ret;
+		if (*status != SWITCHTEC_DLSTAT_INPROGRESS &&
+		    *status != SWITCHTEC_DLSTAT_COMPLETES &&
+		    *status != SWITCHTEC_DLSTAT_SUCCESS_FIRM_ACT &&
+		    *status != SWITCHTEC_DLSTAT_SUCCESS_DATA_ACT)
+			return *status;
 		if (bgstatus == MRPC_BG_STAT_ERROR)
 			return SWITCHTEC_DLSTAT_HARDWARE_ERR;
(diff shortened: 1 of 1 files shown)


lib/fw.c
```diff
@@ -105,8 +105,10 @@ int switchtec_fw_wait(struct switchtec_dev *dev,
 		usleep(5000);
 
 		ret = switchtec_fw_dlstatus(dev, status, &bgstatus);
+
 		if (ret < 0)
 			return ret;
+
```
lsgunth: Don't change unrelated whitespace

lib/fw.c
```diff
@@ -220,7 +222,7 @@ int switchtec_fw_write_fd(struct switchtec_dev *dev, int img_fd,
 
 		ret = switchtec_fw_wait(dev, &status);
 		if (ret != 0)
-		    return ret;
+			return status;
```
lsgunth: This doesn't look correct. We might return the wrong thing when 'ret' actually is the error.

I'd suggest we change switchtec_fw_wait() to return a different value if the status indicates a CRC error.

wesleywesley:
@lsgunth 
confirm with firmware team, when bgstatus is error, the front status is already set to the failure reason, that is 2-6, except timeout.
So change return HARDWARE ERROR, to return the original status in fw_wait function.

wesleywesley:
1.	Once fw dl bgstatus is 0xff, front status could be 0 to 14

Among them:
0, 1, 7-9 are no error state
2-6 are error state
0Xa to 0Xd (10-13) reserved
0xe(14) are timeout state

-- for this reason: change the code to
if status is in error state, return the error state
if status is out of the error state, return the "hardware error"

2.	Opposite:
Once front status in error state: 2 to 6, 0xA to 0Xd , the bgstatus must be 0xff.

PR #38, opened 2018-08-09
Devel

wesleywesley:


wesleywesley:
We have go through all the requests when merge from bug-fix branches to devel branch.
Merge to master from devel now.

PR #39, opened 2018-08-22
Map the entire gas space twice for two different section

wesleywesley:
new F/W GAS_READ_MRPC command could be used to access(r/w)
GAS space which beyond GAS MRPC section, so limit the scope
to the first 4K bytes of GAS space

### DIFF
--- a/lib/platform/linux.c
+++ b/lib/platform/linux.c
@@ -509,6 +509,60 @@ static int linux_port_to_pff(struct switchtec_dev *dev, int partition,
 #define __force
 #endif
 
+static ssize_t resource_size(struct switchtec_linux *ldev, const char *fname)
+{
+	char respath[PATH_MAX];
+	struct stat stat;
+	int fd, ret;
+
+	ret = dev_to_sysfs_path(ldev, fname, respath,
+				sizeof(respath));
+	if (ret) {
+		errno = ret;
+		return -1;
+	}
+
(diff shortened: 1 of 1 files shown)


lsgunth:
Nak -- you may be breaking things that have already been used by users. For example, the 'switchtec gas' commands become completely useless after this change. 

You need to still map the entire GAS which means probably two mmap calls. You also *must* make sure it works in a backwards compatible way: which means it must work with the original kernel and the new kernel with the WC MRPC region.

Logan

wesleywesley:
original kernel with iomap, set the whole gas as UC-(uncached minus), but it worked with user tool by 
/device/resouce0_wc WC.
But it exist a aliasing issue.

lsgunth:
???

 You didn't address any of my concerns.

lsgunth:
Ick, no. Please see ca79cc178cd5e167df720902f631de6b9595ae98 for how this should be done. Note it's only very lightly tested and may need some fallback code or something in some cases.

Logan

wesleywesley:
@lsgunth 
thx.
We will test this patch on both UC- and WC+UC- driver.

lsgunth:
Thanks, also please ensure it passes Travis. I didn't do that.

wesleywesley:
@lsgunth it pass Travis-ci check. pls approve. thx.

wesleywesley:
@lsgunth thx.

PR #42, opened 2018-09-04
Fix port binding commands issues.

JoeyZhang-Microsemi:
The output of command port-bind-info does not display strings for
bind state and port binding commands use wrong condition for error
check.


cli/main.c
```diff
@@ -919,10 +919,32 @@ static int arbitration_set(int argc, char **argv)
 	return 0;
 }
 
+static void print_bind_info(struct switchtec_bind_status_out status)
+{
+	int result = status.bind_state & 0x0F;
+	int state = (status.bind_state & 0xF0) >> 4;
+
+	switch (result) {
+	case 0:
```
lsgunth: We should probably have an enum for these case values somewhere.

cli/main.c
```diff
@@ -940,23 +962,19 @@ static int port_bind_info(int argc, char **argv)
 
 	ret = switchtec_bind_info(cfg.dev, &bind_status, cfg.phy_port);
 
-	if (ret < 0) {
-		printf("bind_info_error: %f", ret);
+	if (ret > 0) {
```
lsgunth: You still need to handle the case where switchtec_bind_info() returns less than zero (indicating an error submitting the command or something like that.

cli/main.c
```diff
@@ -940,23 +962,19 @@ static int port_bind_info(int argc, char **argv)
...
+		printf("bind_info_error: %d\n", ret);
```
lsgunth: Probably should also be using switchtec_perror()

cli/main.c
JoeyZhang-Microsemi: The function switchtec_bind_info()/switchtec_bind()/switchtec_unbind() cannot return a value that is less then zero. If the return value is bigger than zero, it indicates a error.

cli/main.c
JoeyZhang-Microsemi: Because same error codes were used by different MRPC commands and have different meaning. so  we just display the error code now.

cli/main.c
JoeyZhang-Microsemi: I agree it. Thanks

cli/main.c
lsgunth: That's wrong. switchtec_bind_info() calls switchtec_cmd() and returns the result. switchtec_cmd() can always return negative numbers.

cli/main.c
lsgunth: Well returning a number is incredibly user unfriendly. You should fix that and decode the number into a useful error message. If you can use switchtec_perror() that would be great, if not you should create something else. switchtec_bind_perror() or something.

lib/switchtec.c
```diff
@@ -589,26 +616,18 @@ float switchtec_die_temp(struct switchtec_dev *dev)
...
 {
-	int ret;
-
 	struct switchtec_bind_status_in sub_cmd_id = {
 		.sub_cmd = MRPC_PORT_INFO,
 		.phys_port_id = phy_port
 	};
 
-	ret = switchtec_cmd(dev, MRPC_PORTPARTP2P, &sub_cmd_id,
+	return switchtec_cmd(dev, MRPC_PORTPARTP2P, &sub_cmd_id,
```
lsgunth: Nice clean up, but can this be another commit?

JoeyZhang-Microsemi:
Thanks for the review. I has split the commit. 

Thanks
Joey

wesleywesley:
Following enum overlap the system error number:

ERR_PHYC_PORT_ARDY_BIND = 0x00000001, 
ERR_LOGC_PORT_ARDY_BIND = 0x00000002, 
…
ERR_BIND_CHECK_FAIL = 0x0000000d,

Which lead to false prompt in case of non-port-bind related commands by following code in switchtec_strerror function:

	case ERR_PHYC_PORT_ARDY_BIND:
		msg = "Physical port already bound"; break; 
	case ERR_LOGC_PORT_ARDY_BIND: 
		msg = "Logical bridge instance already bound"; break; 
        …
	case ERR_BIND_CHECK_FAIL: 
		msg = "Port bind checking failed"; break;

PR #43, opened 2018-09-12
pff csr gas region support max 255 pcie functions instead of 48

wesleywesley:


wesleywesley:
#47 : Fix false maximum supported PCIe function number issue 
solve the same problem more accurately, so close this one accordingly.

PR #45, opened 2018-10-05
Compiler issues

lsgunth:
Here are a couple fixes for issues in new versions of compilers and C libraries.

Thanks,

Logan

### DIFF
--- a/lib/platform/linux-i2c.c
+++ b/lib/platform/linux-i2c.c
@@ -33,6 +33,7 @@
 #include <sys/stat.h>
 #include <sys/ioctl.h>
 #include <sys/mman.h>
+#include <sys/sysmacros.h>
 #include <fcntl.h>
 #include <unistd.h>
 #include <errno.h>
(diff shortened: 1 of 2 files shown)


lsgunth:
Note: these passed travis on my repo:

https://travis-ci.org/lsgunth/switchtec-user/builds/437717431

wesleywesley:
@lsgunth thx for update

I have 2 questions:
1. by what method, we will use the new versions of compilers and C libraries, and see these issues?
by sudo apt-get install xxx?
2. there are lots of snprintf, seems the second patch add check for truncation for 3 of the snprintf, shall other snprintf also be added the check?
or the patch only target for suppress the warnings appeared, and therefore no need for other snprintf if there is no warning appeared by them?

lsgunth:
1. It's not going to be a simple apt-get. You'd need to upgrade libc and gcc packages to debian unstable or something like that.

2. These are the only snprintfs it pointed out. It has a heuristic to tell if it's possible to truncate the string. It's only necessary where truncation is possible and therefore no further changes should be necessary.

PR #46, opened 2018-10-18
Fix a bug in fw-toggle output

wesleywesley:
The fw-toggle command will show success or failure message by
switchtec_perror() at the end of output. switchtec_perror() takes errno
as the only input for message output.

It's possible that all the toggle operations was done successfully, but
the errno is not zero. Consider the case that a code snippet calls a
function with retries, and the first calls failed and errno set, then
one retry succeeded. In this case errno is set, and operations are
succeeded. The problem is the call to switchtec_perror() is
unconditional. This would result in the switchtec_perror() show failure
message on success.

This patch fix this by set correct errno before the call to
switchtec_perror().

### DIFF
--- a/lib/platform/linux-i2c.c
+++ b/lib/platform/linux-i2c.c
@@ -504,6 +504,8 @@ static void i2c_gas_read(struct switchtec_dev *dev, void *dest,
 
 	if (retry_count == MAX_RETRY_COUNT)
 		raise(SIGBUS);
+
+	errno = 0;
 }
 
 static void i2c_memcpy_from_gas(struct switchtec_dev *dev, void *dest,


lsgunth:
It would be better to fix the errno leak at the source, not at the end...

kelvin-cao:
> It would be better to fix the errno leak at the source, not at the end...

My understanding is that it's ok for a API like ioctl to keep a errno unchanged when its return code is 0. And that's what happened. In below case, if errno is set before the ioctlI() call, a successful execution of this ioctl() will return zero and keep the errno unchanged. I think it's correct behavior.

```
                ret = ioctl(idev->fd, I2C_RDWR, &rwdata);
                if (ret < 0)
                        goto i2c_read_fail;
```

In the TWI case, some gas read would fail for the first time and succeed for the second try. So the errno is kept set. And when the code runs to switchtec_perror(), it output errors...

lsgunth:
You should fix this in the gas read then. If a retry fails, clear the errno if it's going to do another try.

Logan

kelvin-cao:
> You should fix this in the gas read then. If a retry fails, clear the errno if it's going to do another try.
> 
> Logan

Errno is only valid/meaningful in a failed API call. So I'd think it's ok for the i2c_gas_data_read() to return with zero and errno set, and the caller of the API should only check the return value and ignore errno. 

For a switchtec_API(), it seems we don't enforce the rule that errno must be cleared when it returns with zero. User code need to check return value of switchtec_API(), and only check errno when non-zero returned. So, for the case to use switchtec_perror() to show 'success', it's up to the user code to update errno for switchtec_perror();

Thought?

lsgunth:
Disagree. It's better to fix it at the source than in one implementation. Then other people don't run into the same issue.

kelvin-cao:
> Disagree. It's better to fix it at the source than in one implementation. Then other people don't run into the same issue.

Take below code snippet for example, I don't think it's the ioctl's responsibility to ensure errno is zero on success. The user code should protect this. And the actual behavior is the ioctl() doesn't update errno on success, so the errno could be non-zero. User should not exepct this code snippet behave correctly.
```
ioctl(idev->fd, I2C_RDWR, &rwdata);
perror("ioctl");
```

Our API use case is quite similar.
```
switchtec_fw_toggle_active_partition(cfg.dev, cfg.firmware, cfg.config);
switchtec_perror("firmware toggle");
```
The user code should do what it should do.

lsgunth:
No, you're wrong. It is definitely the ioctl's responsibility to _not_ set errno if no error occurred.

Logan

kelvin-cao:
> 
> 
> No, you're wrong. It is definitely the ioctl's responsibility to _not_ set errno if no error occurred.
> 
> Logan

It is reasonable to not set errno for an API if no error occurred. 
If errno is already set before a call to ioctl, do you think it's ioctl's responsibility to clear it if no error occurred?

lsgunth:
It is not reasonable to set errno for an API call if no error occurred.

If it's set before the API is called it's not the API's problem. The bug there is in the code that set the errno  when no error occured or failed to clear errno if a specific error was ignored.

Logan

kelvin-cao:
> It is not reasonable to set errno for an API call if no error occurred.
This makes sense to me. I'll generate another patch to clear the errno after a succeeded retry. This patch will also be updated.

PR #47, opened 2018-10-23
Fix false maximum supported PCIe function number issue

wesleywesley:
The maximum supported PCIe function number is set to 255 instead
of the false 48.
To be backward compatible for linux inband mode, a new ioctl and
corresponding data structure are created, while keep the deprecated
one.
Userspace try the new ioctl first, if it fails, use the old ioctl.


wesleywesley:
This patch is correlate with the one at the kernel repo:
[https://github.com/Microsemi/switchtec-kernel/pull/43](url)
Both are fix the false maximum supported PCIe function number in PFF CSR region of GAS.

The Travis CI build passed.

Following combinations have been tested for the aim of compatibility:
Old driver <--> new user tool
New driver <--> old user tool
New driver <--> new user tool

lib/platform/linux.c
```diff
@@ -792,7 +793,23 @@ static void event_summary_copy(struct switchtec_event_summary *dst,
 	for (i = 0; i < SWITCHTEC_MAX_PARTS; i++)
 		dst->part[i] = src->part[i];
 
-	for (i = 0; i < SWITCHTEC_MAX_PORTS; i++)
+	for (i = 0; i < SWITCHTEC_MAX_PFF_CSR && i < ARRAY_SIZE(src->pff); i++)
+		dst->pff[i] = src->pff[i];
+}
+
+static void event_summary_copy_v2(struct switchtec_event_summary *dst,
+			       struct switchtec_ioctl_event_summary_v2 *src)
```
lsgunth: Don't copy code. Just pass in the size variable as an argument.

wesleywesley:
@lsgunth 
thx for comments.
event_summary_copy function has been updated.

lsgunth:
We really need to sort out the kernel changes first before I can really review this.

wesleywesley:
@lsgunth 
Pull request: switchtec: Fix false maximum supported PCIe function number issue #43 cooperate with the #47 , fixing in identical bug but in kernel code

PR #52, opened 2018-11-19
In-band GAS read & GAS dump output sanity check

wesleywesley:
To distinguish all 1 returns of in-band GAS read & GAS dump valid
or not, a light way rely on check of device id is added.
Raise the SIGABRT signal when the output is abnormal.

For the reason of Windows system has no define of SIGBUS in
SIGNAL.H, SIGABRT is choosed.

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -2016,7 +2016,6 @@ static struct prog_info prog_info = {
 		"(ex: /dev/switchtec0)",
 };
 
-#ifndef _WIN32
 static void sig_handler(int signum)
 {
 	if (signum == SIGBUS) {
@@ -2030,14 +2029,6 @@ static void setup_sigbus(void)
 {
 	signal(SIGBUS, sig_handler);
 }
-
-#else /* _WIN32 defined */
-
-static void setup_sigbus(void)
-{
-}
-
-#endif
 
 int main(int argc, char **argv)
 {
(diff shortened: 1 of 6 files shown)


wesleywesley:
@lsgunth 
Before the next step of implementation for sanity check

1. Necessity
Sanity check is necessary for gas operation, especially when user use the gas sub commands for debugging, to distinguish the all 1 read back is valid or not in some unknown hardware state instantly

2. Place holder for the sanity check code
Yes, current implementation have the check every time of for gas read access: gasread8/16/32/64 or memcpy_from_gas/write_from_gas, it is a penalty of performance

For one sanity check when map policy:
1. It confuse these two things, even GAS access is insanity, the map should also success
For example: once hard-reset command executed, the GAS access read will return all 1, that's insanity, it is because the MSE bit is cleared, but this phenomenon should not failed the GAS map operation
2. If hardware enter into insanity state in between of the series of GAS read operation, the once check policy is not applicable
3. sanity check only apply to in-band : Linux and Windows, while twi & uart have its own check method (PEC/CRC and retry), this is the reason for me to choose modify the common file mmap_gas.h dereferenced by Linux.c and Windows.c

lsgunth:
I still don't really see the justification here. If the hardware fails and starts reading ones back it will be pretty obvious it's completely broken and we don't need to check every single access. This is standard for all PCI accesses on Linux. The performance cost is too high for little gain.

wesleywesley:
@lsgunth thx for review.

PR #53, opened 2018-11-28
Fix unreadable output string for list command

wesleywesley:
In system broken state e.g., state after run hard-rest command, all
1 returns from GAS register read access lead to the unreadable
output string by list command.
Fix this issue by output "unknown" string in that case.

Windows platform has no of this issue, for the reason of the list
command fetch required info from registry instead of GAS registers.


wesleywesley:
@kelvin-cao @JoeyZhang-Microsemi @Isgunth
pls help review this patch, because I want to include this patch into recent release, thx.

wesleywesley:
@lsgunth thx for review, now the code change to "If the first character is 0xff then the entire string is bad".

PR #60, opened 2018-12-27
Fix segmentation fault case for UART interface on 32bit system

wesleywesley:
The ap (type of va_list) is undefined after return of function
vsnprintf that uses va_arg(ap, type).
Fix this undefined case by replace vsnprintf with snprintf and
va_arg combination.

Root cause for segmentation fault:
the va_arg after vsnprintf in send_cmd function, return the
value of the first variable argument after the fixed arguments,
instead of the expected second one.
Because the first varialbe arguments is an offset inf GAS, and
the second variable arguments is a memory address; take the
offset(relative low) as memory address and deferencing cause
the segmentation fault.
This issue found on i686, while not on x86_64.

References:
[1] https://linux.die.net/man/3/vsnprintf
The functions vprintf(), vfprintf(), vsprintf(), vsnprintf()
are equivalent to the functions printf(), fprintf(), sprintf(),
snprintf(), respectively, except that they are called with a
va_list instead of a variable number of arguments. These
functions do not call the va_end macro. Because they invoke
the va_arg macro, the value of ap is undefined after the call.
See stdarg(3).
[2] https://linux.die.net/man/3/stdarg
If ap is passed to a function that uses va_arg(ap,type) then
the value of ap is undefined after the return of that function.

### DIFF
--- a/lib/platform/linux-uart.c
+++ b/lib/platform/linux-uart.c
@@ -127,21 +127,26 @@ static int send_cmd(int fd, const char *fmt, int write_bytes, ...)
 	int cnt;
 	char cmd[1024];
 	uint8_t *write_data;
+	uint32_t write_crc;
 	va_list argp;
 
 	va_start(argp, write_bytes);
 
+	if (write_bytes) {
+		write_data = va_arg(argp, uint8_t *);
+		write_crc = va_arg(argp, uint32_t);
+	}
+
 	cnt = vsnprintf(cmd, sizeof(cmd), fmt, argp);
 
 	if (write_bytes) {
-		write_data = va_arg(argp, uint8_t *);
 		for (i = 0; i< write_bytes; i++) {
 			cnt += snprintf(cmd + cnt, sizeof(cmd) - cnt,
 				       "%02x", write_data[write_bytes - 1 - i]);
 		}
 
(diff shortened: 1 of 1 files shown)


wesleywesley:
@lsgunth 
thx for review.
maybe I am not get your point.

I try to generate a new version patch with send_cmd function take 4 fixed arguments, the added 4th arguments is the write_data, like:
static int send_cmd(int fd, const char *fmt, int write_bytes, const void *src, ...)

it worked for cli and gasrd, but for gaswr, the same undefined argp case still exists, when use the va_arg marco to get the "CRC" value, because the argp stay at the addr(offset) after call vsnprintf, instead of the expected "CRC" one.

lsgunth:
You'd have to do the same for the CRC... Also, I wouldn't make the options mandatory, only call the va_args() functions ahead of vsnprintf, if write_data is set. So the prototype should still be:

static int send_cmd(int fd, const char *fmt, int write_bytes, ...)

but if write_bytes is non zero, the first two arguments are data and CRC. The following arguments are passed to vsnprintf.

wesleywesley:
@lsgunth
Pls help review the updated version. Thx.

wesleywesley:
@lsgunth 
Thx for review.

PR #63, opened 2019-01-03
Fix output of event-wait command with false info on Windows platform

wesleywesley:
On Windows platform, the output of event-wait command with false
info when timeout(command such as: switchtec.exe event-wait
switchtec0 -t 1000).
The root cause is: the data structure struct switchtec_event_summary
sum doesn't zeroize on Windows platform like Linux platform, and
in timeout situation, no update of this data structure, then
the output is false and random.
Fix it by zeroize the structure before use it.

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -669,7 +669,7 @@ static int event_wait(int argc, char **argv)
 {
 	const char *desc = "Wait for an event to occur";
 	struct event_list elist[256];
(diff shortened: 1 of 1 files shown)


wesleywesley:
@lsgunth
Thx for review.
The patch is applicable to both Linux and Windows platform before the commit message's modify.

wesleywesley:
@lsgunth 
Thx for review.

PR #65, opened 2019-01-04
Add event check for local partition

wesleywesley:
In switchtec_event_summary_set function, if index type is
SWITCHTEC_EVT_IDX_LOCAL, the waited event bit is set on
variable local_part but not on variable array part[] out
of event summary data structure.

But switchtec_event_check function only check whether the
specified event happened or not by compare variable array
part[].
So there exist one case, the wanted event on the local
partition happened, but finally timeout for waiting.

Fix this false timeout by add event check for local
partition.


wesleywesley:
@lsgunth 
Thx for review.

PR #66, opened 2019-01-07
Event wait type

wesleywesley:
Set the default value of waited event id as -1(SWITCHTEC_EVT_INVALID)
and popup warning when user doesn't specify a valid event type
for event-wait command.


wesleywesley:
@lsgunth 
Thx for review.

PR #68, opened 2019-01-10
Fix a typo in temp_lib.py code

wesleywesley:
In Linux environment, the library should be libswitchtec.so


wesleywesley:
@lsgunth Thx for review.

PR #69, opened 2019-01-11
Reset usage string in case of command belong to named subcommand

wesleywesley:
To avoid the content of usage string with double copies of
programe info(exe) and command(cmd), reset the usage string
in case of command bleong to named subcommand

The double copies of same string output is test with command:
./swtichtec gas read
the output is like:
Usage: ./switchtec gas./switchtec gas read <device> [OPTIONS]


wesleywesley:
@lsgunth Thx for review.

PR #71, opened 2019-01-31
Change port bind info command to list all enabled physical ports

wesleywesley:
If Port Binding Info command input data with physical port ID equal
to 0xff, then the status of all the enabled physical ports will be
queried.
Changed the code accordingly.

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -992,23 +992,28 @@ static int arbitration_set(int argc, char **argv)
 
 static void print_bind_info(struct switchtec_bind_status_out status)
 {
-	enum switchtec_bind_info_result result = status.bind_state & 0x0F;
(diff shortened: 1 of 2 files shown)


cli/main.c
```diff
@@ -1029,7 +1034,10 @@ static int port_bind_info(int argc, char **argv)
 
 	argconfig_parse(argc, argv, desc, opts, &cfg, sizeof(cfg));
 
-	printf("physical port:%d\n", cfg.phy_port);
+	if (cfg.phy_port == 0xff)
```
lsgunth: You've given no way for the user to know about this special un-intuitive 0xff value.  You should make that the default so the user either provides the -f option or not, if they do not it gives all ports.

PR #74, opened 2019-02-21
Valid event id should not be equal to SWITCHTEC_MAX_EVENTS

wesleywesley:
Valid event id check in linux_event_ctl function exclude
SWITCHTEC_MAX_EVENTS


wesleywesley:
@lsgunth 
Thx for review.

PR #78, opened 2019-03-08
Threshold of event counter usage rule rectify

wesleywesley:
Per dev spec: When the counter reaches threshold, a per-port event
notification is reported to the host. Note that only single valid
bit port mask and single valid bit event type mask counter are
able to perform event notification.
Change the 'and' operater to 'or'.


wesleywesley:
@lsgunth 
thx for review.

PR #79, opened 2019-03-11
Enable showing of ltssm.minor status

vermilionink:
The proposed pull contains two parts:
The update switchtec.h is the bug fix.
The update switchtec.c is optional.
Thanks,
Jason


wesleywesley:
@lsgunth thx for review.
@vermilionink thx for the patch.

PR #85, opened 2019-03-20
Rename parameter which collides with C++ keyword "new" operator

wesleywesley:
This change enables C++ compilation and makes the library usable
in C++ projects.

Signed-off-by: Floessie <floessie.mail@gmail.com>


wesleywesley:
@lsgunth 
Thx for review.

The code change is manually.
So for keep the author info, intentionally change the signed-off-by to Floessie manually.

Maybe next time the same case happen, I can generate a patch first, then applying it, that will keep the author info intact. Like what I do for sync up the patches from kernel.org to microsemi/switchtec-kernel repo.

PR #87, opened 2019-03-26
Fall back to resource0 in case resource0_wc doesn't exist

wesleywesley:
If management/nt ep BAR is non-prefetchable, there will be no
resource0_wc device file exists under /sys/devices.
In this case, we fall back to resource0.


wesleywesley:
@lsgunth 
Thx for review.

PR #91, opened 2019-04-16
Display the physical port when list all port informaion

wesleywesley:
When port state in unbound, bind fail, bind in progress, the physical
port information displayed when specify for a single port, but doesn't
appear when list all port information.

Fix this bug by display physical port in all the port states.


wesleywesley:
@lsgunth 
Thx for review.

Regard,
Wesley

PR #92, opened 2019-05-07
Enhance switchtec error handling to make it MRPC aware

kelvin-cao:
A new global variable mprc was introduced for this. On MRPC handling
errors, this variable is set to the corresponding MRPC CMD ID, on other
errors, this variable is set to -1.

### DIFF
--- a/lib/switchtec.c
+++ b/lib/switchtec.c
@@ -388,18 +388,38 @@ void switchtec_status_free(struct switchtec_status *status, int ports)
 }
 
 /**
+ * @brief The MRPC command ID when errno is set.
+ *
+ * If errno is for MRPC (with the SWITCHTEC_ERRNO_MRPC_FLAG_BIT set), this
+ * variable will be set to the corresponding MRPC command ID.
+ */
+int mrpc_error_cmd;
+
+/**
  * @brief Return a message coresponding to the last error
  *
  * This can be called after another switchtec function returned an error
  * to find out what caused the problem.
+ *
+ * For MRPC errors (mrpc_error_cmd is not -1) that are unknown to this function,
+ * the string "Unknown MRPC error" are returned. Otherwise, either proper
+ * system error string or MRPC error string is returned.
  */
 const char *switchtec_strerror(void)
 {
-	const char *msg;
+	const char *msg = "Unknown MRPC error";
+	int err;
 
-	switch (errno) {
-	case 0: msg = platform_strerror(); break;
+	if ((errno & SWITCHTEC_ERRNO_MRPC_FLAG_BIT) == 0) {
+		if (errno)
+			return strerror(errno);
+		else
+			return platform_strerror();
+	}
 
+	err = errno & ~SWITCHTEC_ERRNO_MRPC_FLAG_BIT;
+
+	switch (err) {
 	case ERR_NO_AVAIL_MRPC_THREAD:
 		msg = "No available MRPC handler thread"; break;
 	case ERR_HANDLER_THREAD_NOT_IDLE:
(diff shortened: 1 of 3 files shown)


inc/switchtec/switchtec.h
```diff
@@ -273,6 +274,8 @@ int switchtec_event_wait(struct switchtec_dev *dev, int timeout_ms);
 
 /*********** Generic Accessors ***********/
 
+extern int mrpc;
```
lsgunth: I would suggest that this belongs in switchtec/errors.h

lsgunth: Also it needs a much better name. 'mrpc_error_cmd' or something like that. Global variables should always be verbosely named.

lib/switchtec.c
```diff
@@ -395,7 +403,10 @@ void switchtec_status_free(struct switchtec_status *status, int ports)
  */
 const char *switchtec_strerror(void)
 {
-	const char *msg;
+	const char *msg = "Unknown error";
+
+	if (mrpc == -1)
+		return strerror(errno);
```
lsgunth: I think you need to call platform_strerror() in this path as well...

lib/switchtec.c
```diff
@@ -460,7 +479,14 @@ const char *switchtec_strerror(void)
 void switchtec_perror(const char *str)
 {
 	const char *msg = switchtec_strerror();
-	fprintf(stderr, "%s: %s\n", str, msg);
+
+	if (!strcmp(msg, "Unknown error"))
```
lsgunth: This seems like you should have an if statement on the mrpc global variable. Why would "unknown error" indicate that you'd print with an MRPC CMD prefix?

lib/switchtec.c
```diff
@@ -592,8 +618,8 @@ int switchtec_log_to_file(struct switchtec_dev *dev,
 		return log_b_to_file(dev, MRPC_FWLOGRD_THRD, fd);
 	};
 
-	errno = EINVAL;
-	return -errno;
+	set_errno(EINVAL);
```
lsgunth: This seems like it should be in separate patches. If it's even necessary which I'm not convinced of. There's a large number of places that set errno and we have to be careful with this.

My expectation is seeing 'mrpc' is only set when an mrpc command fails we should be mostly fine seeing all the existing generally stops when there is an error.  However, we may need a clear error function to reset mrpc to -1 when someone sets errno to 0.

lib/switchtec.c
kelvin-cao: I wanted to give more detailed info which includes the CMD ID and error code returned from firmware for the *unknown MRPC errors*. So it seems I should return a more specific string from switchtec_strerror() to make it more clear, like 'Unknown MPRC error'.

For other cases, either proper platform error strings or strings for known MRPC errors are returned in switchtec_strerrro().

The reason to add this for unknown MRPC errors is that we found that it's likely some errors from firmware *may* not be documented in spec, so we cannot decode them properly. And for library user, it would be useful for them to show raw MRPC errors if they do not want the strings.

lib/switchtec.c
kelvin-cao: I considered these two solutions. The PROs of current one is that for upcoming APIs we won't need to worry about to reset the mrpc to -1, and we also do not have to stop after an error. The Cons is we need to go through all the places that set errno.

If we introduce another error function to clear error, we will need to be very careful to figure out if the path prior to the error has updated 'mrpc' or not. I think although we have two variables 'mrpc' and 'errno', we should treat them as one, and the operation should be atomic. It would make the concept more clear and less restriciton in the coding.

I'd think that going through all the places that set errno is the cost we need to pay.

lib/switchtec.c
kelvin-cao: Ok, I found current Linux platform_strerror() simply returns "Success", I'll make it wrapping the strerror().

inc/switchtec/switchtec.h
kelvin-cao: Will update.

kelvin-cao:
In this updated patch. 
1, move 'extern int mrpc' to 'switchtec/errors.h'
2, rename 'mrpc' into 'mrpc_error_cmd'
3, remove the reference to set_errno()
4, update the default MRPC error string in switchtec_strerror() from 'Unknown error' to 'Unknown MRPC error'
5, update Linux version platform_strerror() to make it wrapping 'strerror()'

lib/switchtec.c
lsgunth: Well all our APIs support not stopping, it's just that we don't have much code that does this.

It's not that hard to go through all the cases where we set the errno. Except the problem is you can't change all the other system calls that do this. So if the user gets a failed MRPC command followed by a failed malloc (for example) then it will look like an MRPC error when they call switchtec_perror().

The downside of the clear errno function is that users of our libraries probably won't realise they need to do this. But I think it's the only correct way to do it even if it's hard on the users.

Another thought I had was to use a couple high bits in errno to indicate that the mrpc_error is valid. That's probably the best choice unless I'm missing something. That way when errno is cleared we know to ignore mrpc_error.

lib/platform/linux.c
```diff
@@ -64,7 +64,7 @@ struct switchtec_linux {
 
 const char *platform_strerror(void)
 {
-	return "Success";
+	return strerror(errno);
```
lsgunth: This isn't quite valid. strerror() also applies on windows, it just doesn't cover all the cases and we need to fall back on the windows platform errors.

lib/switchtec.c
kelvin-cao: > So if the user gets a failed MRPC command followed by a failed malloc (for example) then it will look like an MRPC error when they call switchtec_perror().

I'm thinking it's a inproper use of our APIs. User should call switchtec_perror() immediately after a switchtec API. And for a system call error output, user should use the perror() instead of swtichtec_perror(). We only need to ensure that the error of each switchtec API function be properly returned.
```
ret = switchtec_XXX();
if (ret)
        switchtec_perror();
p = malloc();
if (p == NULL)
        switchtec_perror(); --> perror();
```
> Another thought I had was to use a couple high bits in errno to indicate that the mrpc_error is valid. 

Because the 'MRPC Command Return Value' takes 4 bytes, and the errno is of type int, I don't think it's a good idea to encode anything into errno, which may break the MRPC interface.

lib/platform/linux.c
kelvin-cao: Sorry, I didn't get your point. There's a platform_strerror() implementation on the windows platform, but it's using Windows system APIs, not strerror().

lib/switchtec.c
lsgunth: No, that doesn't change anything. What if instead of malloc() they call one of our API's that fails because of an internal malloc() call?  We simply can not go through all our code and clear mrpc_error for every system call that might set errno.

So I still think that using a single high bit to say that the mrpc_errno is the correct approach. That way anything else that overwrites errno automatically clears the bit and then we don't check mrpc_errno. I don't think there's any error code that uses more than 30 bits so using the 31st bit (with the 32nd bit as the sign) should be acceptable.

lib/platform/linux.c
lsgunth: You still need to call strerror() on windows and all other platforms. strerror() is standard to C not to linux. Malloc calls, for example, always set errno. Windows also needs to call it's platform specific error handler because all the syscalls the interface with the operating system have a different error facility.

So moving the strerror into the linux specific platform is wrong. It needs to remain in the general case.

lib/switchtec.c
kelvin-cao: Unluckily, below error code is using all bits. Even though we can mask off the high bit in a MRPC call, as it's the only one, but there's still chance firmware add other error code like this to make this workaround invalid. 

`ERR_ACCESS_REFUSED = 0xFFFF0001`

If we want to ensure each switchtec API funciton returns (mrpc_error_cmd, errno) pair correctly, we don't actually need to go through every system call, but just clear the errno after a non stopped switchtec function call in the switchtec function implementation (for a void returning switchtec function, we may need to do the same thing).  So the convention turns out to be that if a switchtec function calls another switchtec function, either stop on error or set_errno(0) (mrpc_error_cmd = -1, errno = 0), if If all switchtec functions return (mrpc_error_cmd, errno) correctly, users do not need to clear errno themselves. What do you think? 
```
int switchtec_functionA()
{
        switchtec_funtionB();
        set_errno(0);        // mrpc_error_cmd = -1, errno = 0
        p = malloc();
        if (p == NULL)
                return -1;
}
```

kelvin-cao: No, user code need to follow the same convention.

```
switchtec_functionA();
set_errno(0);
switchtec_functionB();
```

lsgunth: The point is that adding this set_error() function to the api requirements is complicated and you can't expect your callers to get it correct (or even us to get it correct inside the library). It'll take a bunch of nasty code audits to ensure everything is right.

Allowing for the last setter of errno to override the mrpc_errno is a huge win in terms of the complexity we have to consider when using errno.

I don't buy that we have to plan for the firmware to set the high bit for three reasons:

1) If the firmware happens to set the high bit it's fine because we already need it set.

2) We have control over the firmware. Go tell them to limit the error number to 30 bits or push back if they add an error code that high. 30 bits is more than enough for any of their future purposes.

3) If we absolutely had to, we could map the crazy error with the high bit set into a sane lower numbered error.

The only concern is if some upper layer software wants to use that bit for something. This is pretty rare though and in this case using switchtec_perror() would be invalid.

lib/platform/linux.c
wesleywesley: Sorry, seems strerror() (return string describing error number) standard to Linux, its counterpart on Windows here is "FormatMessage function", (Formats a message string, while FORMAT_MESSAGE_FROM_SYSTEM is specified, an an application can pass the result of the GetLastError function to retrieve the message text for a system-defined error)

lib/switchtec.c
kelvin-cao: I've aligned with firmware, they agreed to remove the 0xffff0001 return code from Gen4, and make sure the highest two bits are not used (cleared). I'll update with the bit 31st as the flag and keep 32nd bit zero as the sign.

lib/platform/linux.c
lsgunth: No. strerror() is *not* standard to Linux. It's standard to C and applies to all standard C functions. Linux also uses it heavily for system calls. Windows does not use it for system calls and instead uses FormatMessage but windows still uses it for all standard C functions (like malloc). 

So, once again, moving the strerror into the linux specific platform is wrong. Don't do it.

kelvin-cao:
Updated.

1. use 31st bit as the flag.
2. move strerror() out of platform_strerror().

kelvin-cao:
Updated per comment.

PR #94, opened 2019-05-14
Cfg positive handler fix

wesleywesley:


kelvin-cao:
Looks good.

wesleywesley:
> It's gross that strtoul() behaves that way. But this looks like a good solution.

Yes, strtoul() convert a negative number to a huge positive number.

wesleywesley:
> It's gross that strtoul() behaves that way. But this looks like a good solution.
> 
> Yes, strtoul() convert a negative number to a huge positive number.

man strtoul:
Negative values are considered valid input and are silently converted to the equivalent unsigned long int value.

PR #97, opened 2019-05-16
Gas fix

wesleywesley:


cli/gas.c
```diff
@@ -287,7 +287,7 @@ static int gas_dump(int argc, char **argv)
 	} cfg = {};
 	const struct argconfig_options opts[] = {
 		DEVICE_OPTION,
-		{"count", 'n', "NUM", CFG_LONG_SUFFIX, &cfg.count, required_argument,
+		{"count", 'n', "NUM", CFG_POSITIVE, &cfg.count, required_argument,
```
lsgunth: This patch is completely wrong. CFG_LONG_SUFFIX accepts size arguments with a suffix. For example '4k' would parse as 4096. Positive, Non-negative, etc don't do this. I don't see a lot of value in changing this.

wesleywesley: Yes, I know, a suffix such as "Pi" will auto-complete the zeros after the size number.

But what we come across was some one input a negative value (example: for gas read "--addr/-a" argument), finally this negative is convert to a huge positive (CFG_LONG_SUFFIX type handler return value was "int", while "addr" variant is type of "unsigned long"), this is not what we expected.

I don't want to change CFG_LONG_SUFFIX handler to same as CFG_POSITIVE/CFG_NONNEGATIVE handler, namely treat it as error when negative value input.
On the hand, "trick" like "suffix" is unnecessary for these gas operations.
These are the reaons for me to change these argument type.

lsgunth: No the 'suffix' is not a trick, it's a feature. I don't know what you mean by "Pi" or auto-complete zeros... that's really not what's happening. 

If there's a problem with LONG_SUFFIX, then fix it. But removing the suffix feature is not acceptable.

cli/gas.c
```diff
@@ -402,8 +402,8 @@ static int gas_read(int argc, char **argv)
 
 	static struct {
 		struct switchtec_dev *dev;
-		unsigned long addr;
-		unsigned long count;
```
lsgunth: No, I don't think this is a good idea either. If these stay 'long' instead of 'unsigned long' then you can detect a negative number and do something about it. With 'unsigned long' a negative number becomes a positive.

If you're going to do anything, put an if statement to check for a negative and error out. Or add a CFG_SIZE_SUFFIX or something along those lines.

wesleywesley: Yes, this patch was modified the previous "unsinged long" to "long", to align with the return value type of handler for CFG_LONG_SUFFIX. After this modified, we can determine whether the number input by user is negative or not. 


Do you feel it is necessary to add two more type/handler pairs:
CFG_POSITIVE_SUFFIX & CFG_NONNEGATIVE_SUFFIX for the arguments here used in gas operations.
gas dump: count (CFG_POSITIVE_SUFFIX)
gas_read: addr (CFG_NONNEGATIVE_SUFFIX) / count (CFG_POSITIVE_SUFFIX)
gas_write: addr (CFG_NONNEGATIVE_SUFFIX)
By these two handler, the addr/count check will be moved to handler, instead of here in gas.c.

Or keep as it is now, the check was in gas.c.

lsgunth: That would probably be sufficient. I suggested size_suffix, because this will probably be exclusively used for sizes and you could store the value in a size_t.

wesleywesley:
@lsgunth 
Thx for review.

Regard,
Wesley

PR #98, opened 2019-05-16
Ums reviewed patches

kelvin-cao:
These patches have been reviewed with Logan.


kelvin-cao:
Two tweeks to pass travis.

PR #104, opened 2019-07-01
Mrpc gas x2

kelvin-cao:
- Fix a typo 'create_gas_write'
- Introduce local_pax_id to make it possible to test if MRPC is targeting local the PAX switch

### DIFF
--- a/inc/switchtec/mrpc.h
+++ b/inc/switchtec/mrpc.h
@@ -60,6 +60,8 @@ enum mrpc_cmd {
 	MRPC_MULTI_CFG = 29,
 	MRPC_SES = 30,
 	MRPC_RD_FLASH = 31,
+	MRPC_GAS_READ = 41,
+	MRPC_GAS_WRITE = 52,
 	MRPC_ECHO = 65,
 	MRPC_GET_PAX_ID = 129,
 };
--- a/inc/switchtec/mrpc.h
+++ b/inc/switchtec/mrpc.h
@@ -61,6 +61,7 @@ enum mrpc_cmd {
 	MRPC_SES = 30,
 	MRPC_RD_FLASH = 31,
 	MRPC_ECHO = 65,
+	MRPC_GET_PAX_ID = 129,
 };
 
 enum mrpc_bg_status {
--- a/lib/switchtec.c
+++ b/lib/switchtec.c
@@ -147,6 +147,25 @@ static int set_gen_variant(struct switchtec_dev * dev)
 	return 0;
 }
 
+static int set_local_pax_id(struct switchtec_dev *dev)
+{
+	unsigned char local_pax_id;
+	int ret;
+
+	dev->local_pax_id = -1;
+
+	if (!switchtec_is_pax(dev))
+		return 0;
+
+	ret = switchtec_cmd(dev, MRPC_GET_PAX_ID, NULL, 0,
+			    &local_pax_id, sizeof(local_pax_id));
+	if (ret)
+		return -1;
+
+	dev->local_pax_id = local_pax_id;
+	return 0;
+}
+
 /**
  * @brief Open a Switchtec device by string
  * @param[in] device A string representing the device to open
@@ -226,6 +245,9 @@ found:
 		errno = ENODEV;
 
 	if (set_gen_variant(ret))
+		return NULL;
+
+	if (set_local_pax_id(ret))
 		return NULL;
 
 	return ret;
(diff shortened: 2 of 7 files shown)


inc/switchtec/mrpc.h
```diff
@@ -60,8 +60,10 @@ enum mrpc_cmd {
 	MRPC_MULTI_CFG = 29,
 	MRPC_SES = 30,
 	MRPC_RD_FLASH = 31,
+	MRPC_GAS_READ = 41,
 	MRPC_ECHO = 65,
 	MRPC_GET_PAX_ID = 129,
+	MRPC_GAS_WRITE = 135,  // CHECK ME
```
lsgunth: Please check and remove comment before merging.

There's also another check me comment in the same file that should be resolved and removed.

lib/switchtec.c
```diff
@@ -228,6 +247,9 @@ struct switchtec_dev *switchtec_open(const char *device)
 	if (set_gen_variant(ret))
 		return NULL;
 
+	if (set_local_pax_id(ret))
```
lsgunth: This is going to break any hardware that does not support MRPC_GET_PAX_ID.

lib/switchtec.c
```diff
@@ -147,6 +147,25 @@ static int set_gen_variant(struct switchtec_dev * dev)
...
+
+	if (!switchtec_is_pax(dev))
+		return 0;
+
+	ret = switchtec_cmd(dev, MRPC_GET_PAX_ID, NULL, 0,
+			    &local_pax_id, sizeof(local_pax_id));
+	if (ret)
+		return -1;
+
+	dev->local_pax_id = local_pax_id;
```
lsgunth: Honestly, I don't like this. I'd rather just stick with the constant SWITCHTEC_PAX_ID_LOCAL. If the user specifies the local pax_id on the command line then it will still use MRPC commands. It will only fall back to the default behavior if the pax id is not actually specified.

lib/switchtec.c
kelvin-cao: Should return 1 on non-pax switch, will update.

lib/switchtec.c
kelvin-cao: I think there's still some benefit if we take '-x local_pax_id' the same as '-x' is not specified. 

- When do scripting, user maybe not aware of what the pax id of local pax is, but just specifying all the valid pax id in the script. I think it would be more reasonable to check for local/remote, rather then '-x' parameter.
- The new local_pax_id is also meaningful for some cases that user want to know on which pax switch they are locally operating on and therefore apply some policy (the API to get it is not available now). 

User would expect the behavior is consistent if the target is local, no matter if the '-x' is specified or not. Therefore the rule is also clear, when accessing local, use default ops and when accessing remote, use MRPC ones.

lsgunth: I don't know. I see it as more consistent to behave one way if -x is specified and another way if it's not. It's weird to have some -x *values* do one thing and others and others do another.

inc/switchtec/mrpc.h
kelvin-cao: Firmware decided to use current values in the most recent release and update the values (to final ones, currently the firmware is still under development) in the next release. So I think we can keep the comment here till firmware change it to final ones.

lib/switchtec.c
kelvin-cao: Ooh, you mean the MRPC_GET_PAX_ID. I think, for PAX, the firmware need to have a way to report the pax id of the locally connected PAX switch, otherwise, we need to quit. It's like we cannot get the partition of a PSX switch.

lib/switchtec.c
kelvin-cao: It seems to me that the more fundamental difference is local and remote. The users should have the sense that for a PAX fabric, they see local switch and remote switches, and therefore the ops for local and ops for remote. So I'd think current policy is more consistent with the concept that the firmware wants to deliver to users.

inc/switchtec/mrpc.h
lsgunth: In that case, I'd say don't bother merging it until we have the proper values in place.

lib/switchtec.c
lsgunth: Whatever, they *should* be consistent anyway.

kelvin-cao:
I've updated the MRPC_GAS_WRITE define with final value. For the other checkme comment, I'll update in other commit.

PR #110, opened 2019-09-04
GEN4 FW Image/Partition Handling

lsgunth:
This is a bit of a rush job, I'd probably clean it up more if I had more time. However, this should be complete.

I've tested all the gen3 changes as best as I can. The gen4 changes are fairly simple, but need testing. 

Please review and test. As we are pressed by your deadline, please make any changes as needed to get it to work and let me know what they are. If everything looks good, then just merge it.

### DIFF
--- a/inc/switchtec/mrpc.h
+++ b/inc/switchtec/mrpc.h
@@ -61,6 +61,7 @@ enum mrpc_cmd {
 	MRPC_SES = 30,
 	MRPC_RD_FLASH = 31,
 	MRPC_GAS_READ = 41,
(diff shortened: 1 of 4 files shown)


kelvin-cao:
Thanks for the code. I've udpated based on this PR and make gen4 work, it's been merged to devel. #111 

Generally I added a gen4 specific MRPC to get the fw partition info like active/read_only. Since these info are partition related, not fw image related. So I also introduced functions to do it and also calling the metadata functions to provide the complete part information. 

Some other minor change for bugfix and function completion.

PR #111, opened 2019-09-06
Fw gen4 x2

kelvin-cao:
The purpose of this patch series is to add Gen4 support of the fw related operations.

This is based on the Logan's work with few changes to make Gen4 work.


lib/fw.c
```diff
@@ -729,6 +705,40 @@ static int switchtec_fw_info_metadata_gen3(struct switchtec_dev *dev,
...
+			ret = switchtec_fw_map_get_active(dev, inf);
+			break;
+		default:
+			ret = switchtec_flash_part(dev, inf, inf->part_id);
+	}
+
+	if (ret)
+		return ret;
+
+	if (inf->part_id == SWITCHTEC_FW_PART_ID_G3_NVLOG)
```
lsgunth: Small thing but this check was duplicated: it's in switchtec_fw_info_metadata_gen3().

lib/fw.c
```diff
@@ -773,6 +783,12 @@ static int switchtec_fw_info_metadata_gen4(struct switchtec_dev *dev,
 	return -1;
 }
 
+static int switchtec_fw_part_info_gen4(struct switchtec_dev *dev,
+				       struct switchtec_fw_image_info *inf)
+{
+	return switchtec_fw_info_metadata_gen4(dev, inf);
```
lsgunth: Not really sure the point of this helper.... it's a wrapper that does nothing, extra noise.

lsgunth: Oh never mind, I should have looked at the full series before commenting.

lib/fw.c
```diff
@@ -783,9 +783,92 @@ static int switchtec_fw_info_metadata_gen4(struct switchtec_dev *dev,
...
 				       struct switchtec_fw_image_info *inf)
 {
+	struct switchtec_flash_part_all_info_gen4 all_info;
+	struct switchtec_flash_part_info_gen4 *part_info;
+
+	int ret;
+	uint8_t subcmd = MRPC_PART_INFO_GET_ALL_INFO;
+
+	ret = switchtec_cmd(dev, MRPC_PART_INFO, &subcmd, sizeof(subcmd),
+			    &all_info, sizeof(all_info));
```
lsgunth: Hmm, kind of tricky, but in the future it might be nice to find a way to call this only once when getting all the partitions. Instead of once for every partition and then only using the one partition's info. Just a future thought for a cleanup.

lib/fw.c
```diff
@@ -1292,6 +1294,11 @@ int switchtec_fw_is_boot_ro(struct switchtec_dev *dev)
 
 	int ret;
 
+	if (!switchtec_is_gen3(dev)) {
+		errno = ENOTSUP;
+		return -1;
```
lsgunth: Minor fix: Should probably just return 0 saying the partition is not read only when it's not supported.

lsgunth:
Just put some minor thoughts for some small fix ups but your changes largely look good. Thanks!

lib/fw.c
kelvin-cao: I agree.

lib/fw.c
kelvin-cao: F/W is still updating this part of MRPC interface, in gen4, it's more likely that we will have no MRPC command to set RW to a partition and then allow for programming it. So let's wait and see what will be final interface from the F/W and tuning this part in the software accordingly.

lib/fw.c
kelvin-cao: Yes, should be removed.

kelvin-cao:
> 
> 
> Just put some minor thoughts for some small fix ups but your changes largely look good. Thanks!

THANK YOU for the code and review!

PR #114, opened 2019-10-15
Fixed fw-update '-A' option not working issue.

xyan264:


cli/main.c
```diff
@@ -1405,6 +1409,27 @@ static int fw_update(int argc, char **argv)
 	progress_finish();
 	printf("\n");
 
+	if ((phase_id == SWITCHTEC_BOOT_PHASE_BL2) && !cfg.dont_activate) {
```
lsgunth: NAK

'dont_activate' is handled within switchtec_fw_write_file() and should remain that way. This looks like a bug in switchtec_fw_write_file_ex() which isn't in master or devel.

kelvin-cao: Hi Logan,

It's some code that not ready to merge into devel/master. I'll handle it offline.

Thanks,
Kelvin

PR #115, opened 2019-10-19
Added warning messages for any recovery routines that will change OTP

yungwesl:
Added the following warning block to the tools
  config-set      Set the device security settings (BL1 and Main Firmware only)
  kmsk-entry-add  Add a KMSK entry (BL1 and Main Firmware only)
  state-set       Set the secure state (BL1 and Main Firmware only)
  dport-lock-update Update secure unlock version

---
        fprintf(stderr, "WARNING: This operation makes changes to the device OTP memory and is not reversable. ");
        ret = ask_if_sure(0);
        if(ret)
                return ret;
---

### DIFF
--- a/cli/recovery.c
+++ b/cli/recovery.c
@@ -680,6 +680,10 @@ static int security_config_set(int argc, char **argv)
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


kelvin-cao:
Duplicate with #120, close this.

PR #116, opened 2019-10-22
Fixed fw-info fail when some partitions are missing

xyan264:
Check valid flag in partition to make sure only the valid partitions are processed.


kelvin-cao:
The fw-info doesn't work on Gen3.

PR #119, opened 2019-10-28
Changed behavior of -y option in recovery commands 

xyan264:
Changed behavior of -y option in recovery commands so that it is consistent with other main menu commands.

### DIFF
--- a/cli/recovery.c
+++ b/cli/recovery.c
@@ -495,7 +495,7 @@ static int fw_transfer(int argc, char **argv)
 		struct switchtec_dev *dev;
 		FILE *fimg;
 		const char *img_filename;
-		int confirm;
+		int assume_yes;
 		int force;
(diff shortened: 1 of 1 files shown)


cli/recovery.c
```diff
@@ -544,7 +544,7 @@ static int fw_transfer(int argc, char **argv)
 
 	print_fw_image_info(cfg.img_filename, &finfo);
 
-	ret = ask_if_sure(!cfg.confirm);
+	ret = ask_if_sure(cfg.confirm);
```
kelvin-cao: I'd suggest to use the variable name 'assume_yes' to be consistent with the name used in fw-update CLI, which would be more meaningful.

xyan264: Code updated according to your suggestion. Please review.

PR #120, opened 2019-10-28
Added warning to recovery commands that modify OTP

xyan264:


kelvin-cao:
Currently the fw-update/fw-info doesn't work on Gen3. And since 00000000 is valid secure version number, we should not display it for Gen3 chip which doesn't actualy support secure version, use 'N/A' is more reasonable.

PR #122, opened 2019-11-01
Add partition redundancy flag support

kelvin-cao:
A new redundancy flag for the partitions was introduced in firmware. With this flag set for a given partition type, like main firmware partitions, when the active partition failed to boot for some reason, the other valid partition of the same type will be chosen as the active partition to boot.

Two new commands fw-redund-set/fw-redund-clear are added for the flag set/clear. And the fw-info command is also enhanced to display this new flag.

Given we already have two flags (Ro, Running) to show and adding the third one makes the display of them looks disordered. I also tuned the display to make the flags be shown at given offset like below.

```
sudo ./switchtec fw-info /dev/switchtec1
Currently Running:
  IMG   Version: 3.40 B034
Active Partitions:
  MAP   Version: 3.40 B034      CRC: 53486e39         (Running)
  KEY   Version: 3.40 B034      CRC: 96cb6e33                    (Redundant)
  BL2   Version: 3.40 B034      CRC: 61f30035   (RO)  (Running)  (Redundant)
  IMG   Version: 3.40 B034      CRC: 96cb6e33   (RO)  (Running)  (Redundant)
  CFG   Version: 3.40 B034      CRC: f1c7fc63   (RO)  (Running)  (Redundant)
Inactive Partitions:
  MAP   Version: 3.40 B034      CRC: 53486e39
  KEY   Version: 3.40 B034      CRC: 96cb6e33                    (Redundant)
  BL2   Version: 3.40 B034      CRC: 61f30035                    (Redundant)
  IMG   Version: 3.40 B034      CRC: 96cb6e33                    (Redundant)
  CFG   Version: 3.40 B034      CRC: f1c7fc63                    (Redundant)
```


cli/main.c
```diff
@@ -1433,6 +1433,108 @@ static int fw_toggle(int argc, char **argv)
...
+	}
+
+	ret = switchtec_fw_setup_redundancy(cfg.dev, 1, cfg.type);
+	if (ret)
+		switchtec_perror("set fw redundancy");
+
+	return ret;
+}
+
+static int fw_redund_clear(int argc, char **argv)
```
lsgunth: This function is nearly identical to the one above it. I'm sure one way or another you can make them the same function....

kelvin-cao: Rebased and updated the CLI with adding only one command.

kelvin-cao:
> 
> 
> Mostly looks good except for one copied function that should be made common.
> 
> Also, in the future please try to separate library changes and CLI changes into different commits. It's an obvious place to make a distinction and smaller commits are much easier to follow.

I thought we want to have the added library function with the CLI change in one commit, like add a new function together with its caller in one commit.

lsgunth:
> > Mostly looks good except for one copied function that should be made common.
> > Also, in the future please try to separate library changes and CLI changes into different commits. It's an obvious place to make a distinction and smaller commits are much easier to follow.
> 
> I thought we want to have the added library function with the CLI change in one commit, like add a new function together with its caller in one commit.

No, we want to add a new function together with it's caller in the same **PR**. If you need to do that in a single commit your commits get way too big.

PR #124, opened 2019-11-05
Print error code hex for fw-update related commands.

xyan264:


kelvin-cao:
Paul, I created another fix commit for it and merged it directly into devel branch.

PR #125, opened 2019-11-05
Fix a bug of fw-read on gen4 switch

kelvin-cao:
The fw image layout on the fw partition changed in Gen4, the metadata is
placed at the very beginning of the partition, rather than at the foot
in Gen3. The Gen4 fw image file layout is also changed, the image body is
placed at offset 0x1000, rather than immediatly after the image header.

This fix makes sure for a fw-read on Gen4 switch that
1, The image body is read from correct offset from a fw partition.
2, The image body is written to the correct offset of a image file.

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -1497,8 +1497,8 @@ static int fw_read(int argc, char **argv)
 	}
 
 	progress_start();
-	ret = switchtec_fw_read_fd(cfg.dev, cfg.out_fd, inf->part_addr,
-				   inf->image_len, progress_update);
+	ret = switchtec_fw_body_read_fd(cfg.dev, cfg.out_fd,
+					inf, progress_update);
 	progress_finish();
 
 	if (ret < 0)
--- a/lib/fw.c
+++ b/lib/fw.c
@@ -699,6 +699,7 @@ static int switchtec_fw_info_metadata_gen3(struct switchtec_dev *dev,
 
 	version_to_string(metadata->version, inf->version,
 			  sizeof(inf->version));
+	inf->part_body_offset = 0;
 	inf->image_crc = metadata->image_crc;
 	inf->image_len = metadata->image_len;
 	inf->metadata = metadata;
@@ -780,6 +781,7 @@ static int switchtec_fw_info_metadata_gen4(struct switchtec_dev *dev,
 
 	version_to_string(metadata->version, inf->version,
 			  sizeof(inf->version));
+	inf->part_body_offset = metadata->header_len;
 	inf->image_crc = metadata->image_crc;
 	inf->image_len = metadata->image_len;
 	inf->metadata = metadata;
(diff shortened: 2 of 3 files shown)


cli/main.c
```diff
@@ -1497,7 +1497,8 @@ static int fw_read(int argc, char **argv)
 	}
 
 	progress_start();
-	ret = switchtec_fw_read_fd(cfg.dev, cfg.out_fd, inf->part_addr,
+	ret = switchtec_fw_read_fd(cfg.dev, cfg.out_fd,
+				   inf->part_addr + inf->part_body_offset,
```
lsgunth: I'd suggest it would be better to pass 'inf' directly to switchtec_fw_read_fd() instead of requiring the caller to know the exact formula for the arguments.

lib/fw.c
```diff
@@ -1260,16 +1262,35 @@ static int switchtec_fw_img_write_hdr_gen4(int fd,
...
+
 	switch (info->gen) {
-	case SWITCHTEC_GEN3: return switchtec_fw_img_write_hdr_gen3(fd, info);
-	case SWITCHTEC_GEN4: return switchtec_fw_img_write_hdr_gen4(fd, info);
+	case SWITCHTEC_GEN3:
+		ret = switchtec_fw_img_write_hdr_gen3(fd, info);
+		break;
+	case SWITCHTEC_GEN4:
+		ret = switchtec_fw_img_write_hdr_gen4(fd, info);
+		lseek(fd, info->part_body_offset, SEEK_SET);
```
lsgunth: Why not just do this inside of switchtec_fw_img_write_hdr_gen4()? Looks like it would be a lot cleaner.

cli/main.c
kelvin-cao: I suppose this API is to read at any offset of any bytes from a partition, so I think we'd better leave the flexibility to user rather than forcing it to read from the beginning of the image body. Or we can provide another API basing on this just to read the body as a counterpart of the write_hdr().

lib/fw.c
```diff
@@ -780,6 +781,7 @@ static int switchtec_fw_info_metadata_gen4(struct switchtec_dev *dev,
 
 	version_to_string(metadata->version, inf->version,
 			  sizeof(inf->version));
+	inf->part_body_offset = metadata->header_len;;
```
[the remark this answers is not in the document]
kelvin-cao: Will update.

lib/fw.c
kelvin-cao: I thought it would make the code more clear that we see hdr handling difference of gen3 and gen4 at this level. On the other hand, yes it would be cleaner to get it into write_hdr_gen4(). Will update.

cli/main.c
lsgunth: Well it's questionable why anyone would want to read the firmware partially. So, I'd probably just say we don't need it (until someone actually does have a use-case).

But yes, if there's another API function that just takes the inf struct and applies the necessary math that would be ok too.

kelvin-cao:
Updated.

cli/main.c
```diff
@@ -1497,8 +1497,8 @@ static int fw_read(int argc, char **argv)
 	}
 
 	progress_start();
-	ret = switchtec_fw_read_fd(cfg.dev, cfg.out_fd, inf->part_addr,
-				   inf->image_len, progress_update);
+	ret = switchtec_fw_img_write_body(cfg.dev, cfg.out_fd,
```
lsgunth: Ick, I very much prefer a name that reflects the fact that it's reading it from the device.

lib/fw.c
```diff
@@ -1223,6 +1225,25 @@ int switchtec_fw_read_fd(struct switchtec_dev *dev, int fd,
...
+/**
+ * @brief Write the body for a Switchtec firmware image file
+ * @param[in]  dev     Switchtec device handle
+ * @param[in]  fd      File descriptor for image file to write
+ * @param[in]  info    Partition information structure
+ * @param[in]  progress_callback This function is called periodically to
+ *     indicate the progress of the read. May be NULL.
+ * @return number of bytes written on success, error code on failure
+ */
+
```
lsgunth: Nit: unnecessary blank line.

lib/fw.c
```diff
@@ -1223,6 +1225,25 @@ int switchtec_fw_read_fd(struct switchtec_dev *dev, int fd,
 	return read;
 }
 
+/**
+ * @brief Write the body for a Switchtec firmware image file
```
lsgunth: Definitely need to describe the fact that this function is reading the firmware from the device and writing it to a file...

kelvin-cao:
Updated.

PR #126, opened 2019-11-13
Fixed seg fault in fw-read if key manifest not exist

xyan264:
Fixed segmentation fault in fw-read when key manifest does not exist.

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -1618,7 +1618,7 @@ static int fw_read(int argc, char **argv)
 
 	if (!inf || !inf->valid) {
 		ret = -1;
-		fprintf(stderr, "fw_part_summary: No valid image for this partition!\n");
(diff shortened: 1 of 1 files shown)


cli/main.c
```diff
@@ -1616,6 +1616,12 @@ static int fw_read(int argc, char **argv)
 	else
 		inf = cfg.inactive ? sum->img.inactive : sum->img.active;
 
+	if (!inf || !inf->valid) {
+		ret = -1;
+		fprintf(stderr, "fw_part_summary: No valid image for this partition!\n");
```
kelvin-cao: Please remove the fw_part_summary from the message, usually we specify the function name in switchtec_perror() in error. It's not necessary in this case.

xyan264: Updated, please review.

PR #127, opened 2019-11-19
Secure boot kmsk missing param fix

xyan264:
Added parameter checking for 'kmsk-entry-add', 'dport-unlock' and 'dport-lock-update' commands.

### DIFF
--- a/cli/recovery.c
+++ b/cli/recovery.c
@@ -977,7 +977,22 @@ static int dport_unlock(int argc, char **argv)
 		{NULL}
 	};
 
+	cfg.serial = 0;
+	cfg.unlock_version = 0xffff;
+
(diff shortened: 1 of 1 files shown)


cli/recovery.c
```diff
@@ -977,8 +977,23 @@ static int dport_unlock(int argc, char **argv)
 		{NULL}
 	};
 
+	cfg.serial = 0;
+	cfg.unlock_version = 0xffff;
```
kelvin-cao: It'd be much cleaner to initialize these in the structure variable definition. Like 

cfg = {
        .unlock_version = 0xffff;
};

In this way, no need to set serial to zero as it's done automatically.

xyan264: Updated. Please review.

PR #128, opened 2019-11-20
Secure boot latency ingress port 60 fix

xyan264:
switchtec_lat_setup() could return a positive error code. Just check if return value is non-zero.

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -561,7 +561,7 @@ static int latency(int argc, char **argv)
 	}
 
 	ret = switchtec_lat_setup(cfg.dev, cfg.egress, cfg.ingress, 1);
-	if (ret < 0) {
+	if (ret) {
 		switchtec_perror("latency");
 		return -1;
 	}


cli/main.c
```diff
@@ -561,7 +561,7 @@ static int latency(int argc, char **argv)
 	}
 
 	ret = switchtec_lat_setup(cfg.dev, cfg.egress, cfg.ingress, 1);
-	if (ret < 0) {
+	if (ret) {
```
kelvin-cao: switchtec_lat_setup() may return positive value on success. What is the return value you see in this case?

xyan264: It's PMON_ERR_LTCY_SRC_PRT_INV (0x00100004) or PMON_ERR_PRT_INV (0x00100002).

kelvin-cao: This doesn't really fix the issue. switchtec_lat_setup() may return positive value on success in current code.

xyan264:
Closed due to out of date. 

Fix updated in PR #171 .

PR #129, opened 2019-11-21
Fix evcntr-wait timeout segfault in Windows

brian-mckenzie:
The evcntr-wait command will segfault in Windows when the timeout expires. This is because switchtec_evcntr_wait() passes NULL to switchtec_event_wait_for()'s "res" parameter, which makes its way to gasop_event_summary(), which then tries to memset it.

This can be fixed by passing a switchtec_event_summary struct instead of NULL.

### DIFF
--- a/lib/platform/gasops.c
+++ b/lib/platform/gasops.c
@@ -270,6 +270,9 @@ int gasop_event_summary(struct switchtec_dev *dev,
 	int i;
 	uint32_t reg;
 
+	if (!sum)
+		return 0;
+
 	memset(sum, 0, sizeof(*sum));
 
 	sum->global = gas_reg_read32(dev, sw_event.global_summary);
--- a/lib/platform/linux.c
+++ b/lib/platform/linux.c
@@ -873,7 +873,7 @@ static int linux_event_summary(struct switchtec_dev *dev,
 	struct switchtec_linux *ldev = to_switchtec_linux(dev);
 
 	if (!sum)
-		return -EINVAL;
+		return 0;
 
 	ret = ioctl(ldev->fd, SWITCHTEC_IOCTL_EVENT_SUMMARY, &isum);
 	if (!ret) {


lsgunth:
I think it would be better to fix this in the other way and put a NULL check where it segfaults in windows.

lib/events.c
```diff
@@ -420,8 +420,12 @@ int switchtec_event_wait_for(struct switchtec_dev *dev,
 		now = ((tv.tv_sec) * 1000 + tv.tv_usec / 1000);
 
 		if (timeout_ms > 0 && now - start >= timeout_ms) {
-			ret = switchtec_event_summary(dev, res);
-			return ret;
+			if (res != NULL) {
+				ret = switchtec_event_summary(dev, res);
```
kelvin-cao: I think it would be more cleaner and more reasonable to make the check of this parameter in the swtichtec_event_summary().

lsgunth: Ditto. There's already a check in linux_event_summary()... The same check is missing in gasop_event_summary().

brian-mckenzie: If we have the NULL check in gasop_event_summary() and have it return a negative value (e.g. -EINVAL) like linux_event_summary(), then switchtec_event_wait_for() will return this value to evcntr_wait(), which will call perror() and exit rather than printing the "timeout" message. I think the same would be true in the Linux case. As per the function descriptions, switchtec_evcntr_wait() and switchtec_event_wait_for() are supposed to return 0 on timeout.

brian-mckenzie: gasop_event_summary() should still probably have a NULL check, but I don't think we should call that function at all if we know we're not giving it the right arguments.

lsgunth: No, the NULL check should simple do nothing, it should not return an error. This is to support other functions that might only optionally set the event summary. 

It's quite common for functions to only set their output if the output is not-null.

brian-mckenzie: So just return 0 in gasop_event_summary() if sum is NULL? Should the same be done in the Linux case too then? Currently linux_event_summary() returns -EINVAL.

lsgunth: Oh, the error is probably ignored in cases where it might be non-zero. Probably easiest to just copy what's in the Linux code. 

If you want to do more work, change it so that a NULL pointer does nothing and returns no error, and add error checking to the places it is missing in.

brian-mckenzie: If we just copy what's in the Linux code, that would be to return -EINVAL when the parameter is NULL, which would lead to evcntr_wait() calling perror() and exiting rather than show the timeout message. I don't think that's what we'd want.

If we make it so a NULL pointer does nothing returns 0, it looks like there aren't any places that would have missing error checks, so this would be fine. We should probably do the same in linux_event_summary() too though for consistency (and so that an evcntr_wait timeout prints the "timeout" message in Linux rather than calling perror).

lsgunth: Sounds good.

brian-mckenzie: Hi @lsgunth -- please review the latest changes when you have a chance.

PR #132, opened 2019-11-26
Switchtec cmd fix

kelvin-cao:
The switchtec_cmd() API returns zero on success, positive and negative value on errors, this was the original definition of it. As code changes, some of the rules become not so clear and there was some breaks in the code.

This patch series ensures that the return value of switchtec_cmd() be
1. zero on success
2. negative on system error with proper errno set
3. positive on MRPC error with proper mrpc_error_cmd/errno set

And it also fixed some the caller in the library to ensure all callers in the library check non-zero return of switchtec_cmd() for errors, rather then check negative value only.


kelvin-cao:
thanks.

PR #134, opened 2019-12-05
Move MRPC_PART_INFO out of for loop to improve code efficiency

xyan264:
Move MRPC_PART_INFO command up one level: from switchtec_fw_part_info_gen4() to switchtec_fw_part_info(), out of the for loop.

### DIFF
--- a/lib/fw.c
+++ b/lib/fw.c
@@ -853,7 +853,7 @@ err_out:
 	return -1;
 }
 
-struct switchtec_flash_part_all_info_gen4 {
+struct switchtec_flash_info_gen4 {
 	uint32_t firmware_version;
 	uint32_t flash_size;
 	uint16_t device_id;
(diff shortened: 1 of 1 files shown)


lib/fw.c
```diff
@@ -970,9 +962,20 @@ static int switchtec_fw_part_info(struct switchtec_dev *dev, int nr_info,
...
+	struct switchtec_flash_info_gen4 all_info;
+
 	if (info == NULL || nr_info == 0)
 		return -EINVAL;
 
+	if (dev->gen == SWITCHTEC_GEN4) {
+		ret = switchtec_cmd(dev, MRPC_PART_INFO, &subcmd,
+				    sizeof(subcmd), &all_info,
+				    sizeof(all_info));
+		if (ret < 0)
```
kelvin-cao: Any non-zero value should be taken as error.

lib/fw.c
```diff
@@ -970,9 +962,20 @@ static int switchtec_fw_part_info(struct switchtec_dev *dev, int nr_info,
 	int ret;
 	int i;
 
+	uint8_t subcmd = MRPC_PART_INFO_GET_ALL_INFO;
```
kelvin-cao: Unnecessary blank line.

lib/fw.c
xyan264: Updated, please review.

lib/fw.c
xyan264: Updated, please review.

PR #135, opened 2019-12-09
Fix fw-info command errors when device is in BL2 boot phase

xyan264:
Use PING command during initialization to determine device generation, skip showing 'current running firmware version' in BL2 boot phase, and set multicfg to none when device is in BL2 phase. 

Also need to change I2C write function to allow stable communication.


lib/recovery.c
```diff
@@ -0,0 +1,103 @@
...
+			sizeof(ping_dw_le),
+			&r, sizeof(r));
+	if (ret == 0) {
+		*reply_dw = le32toh(r.reply);
+		hw_stat = le32toh(r.stat);
+		*phase = hw_stat & 0xff;
+		*rev = (hw_stat >> 8) & 0x0f;
+		*gen = map_to_gen((hw_stat >> 12) & 0x0f);
+	} else {
+		if ((errno & (~SWITCHTEC_ERRNO_MRPC_FLAG_BIT))
```
lsgunth: Can we introduce an ERRNO_ MRPC() macro (or something like that) so this line can be a bit less nasty and not need the odd wrapping?

lib/recovery.c
```diff
@@ -0,0 +1,103 @@
...
+		    == ERR_MPRC_UNSUPPORTED) {
+			*reply_dw = ~ping_dw;
+			*phase = SWITCHTEC_BOOT_PHASE_FW;
+			*gen = SWITCHTEC_GEN3;
+			*rev = SWITCHTEC_REVA;
+
+			errno = 0;
+		} else
```
lsgunth: Both else branches should either have braces or not. Never mixed. Please add braces to the else condition.

lib/switchtec.c
```diff
@@ -60,90 +60,100 @@
...
+
+	ret = switchtec_ping(dev, 0, &r, &phase, &gen, &rev);
+	if (ret)
+		return -1;
+
+	dev->gen = gen;
+
+	if (phase != SWITCHTEC_BOOT_PHASE_FW) {
+		return 0;
+	}
```
lsgunth: Nit: no braces on a single line if.

lib/switchtec.c
```diff
@@ -60,90 +60,100 @@
...
+	{0x4168, SWITCHTEC_PSX},   //PSX 68XG4
+	{0x4152, SWITCHTEC_PSX},   //PSX 52XG4
+	{0x4136, SWITCHTEC_PSX},   //PSX 36XG4
+	{0x4128, SWITCHTEC_PSX},   //PSX 28XG4
+	{0x4200, SWITCHTEC_PAX},   //PAX 100XG4
+	{0x4284, SWITCHTEC_PAX},   //PAX 84XG4
+	{0x4268, SWITCHTEC_PAX},   //PAX 68XG4
+	{0x4252, SWITCHTEC_PAX},   //PAX 52XG4
+	{0x4236, SWITCHTEC_PAX},   //PAX 36XG4
+	{0x4228, SWITCHTEC_PAX},   //PAX 28XG4
```
lsgunth: I don't see the point of this. I prefer using the device ID table unless we can't actually get the device ID; then fall back on generation. But then we'd also need to do something similar for the variant....

lib/recovery.c
```diff
@@ -101,3 +101,24 @@ int switchtec_ping(struct switchtec_dev *dev, uint32_t ping_dw,
...
+int switchtec_get_boot_phase(struct switchtec_dev *dev,
+			     enum switchtec_boot_phase *phase_id)
+{
+	int ret;
+	enum switchtec_gen gen;
+	uint32_t t = 0;
+	uint32_t r = 0;
+	uint32_t rev;
+
+	ret = switchtec_ping(dev, t, &r, phase_id, &gen, &rev);
```
lsgunth: This is such a weird crazy idea to put this in a command called ping.... The firmware folks can do their crazy naming but I think our function should reflect what it actually does (get device info?) -- not some legacy "ping" name.

lib/platform/linux-i2c.c
```diff
@@ -374,15 +375,26 @@ static void i2c_gas_write(struct switchtec_dev *dev, void __gas *dest,
...
+	uint8_t status_retry_count = 0;
 
 	do {
+		status_retry_count = 0;
 		tag = get_tag(idev);
 		i2c_gas_data_write(dev, dest, src, n, tag);
-		status = i2c_gas_write_status_get(dev, tag);
-		if (status == 0 || status == GAS_TWI_MRPC_ERR)
+		do {
+			usleep(5000);
```
lsgunth: A 5ms delay for every write seems like it would slow it down a ton.... Is this really necessary? We only sleep for 1ms below if it fails.... I2C is supposed to have clock stretching if the device we are communicating with isn't fast enough... does that not work for this hardware?

lib/platform/linux-i2c.c
```diff
@@ -374,15 +375,26 @@ static void i2c_gas_write(struct switchtec_dev *dev, void __gas *dest,
...
+			status = i2c_gas_write_status_get(dev, tag);
+			if (status == 0 || status == GAS_TWI_MRPC_ERR)
+				break;
+			status_retry_count++;
+		} while (status_retry_count < MAX_STATUS_GET_RETRY);
```
lsgunth: A retry loop inside a retry loop is a bit much....

lib/switchtec.c
xyan264: Previously our device only has one boot phase -- main firmware. Now we introduce the new boot phase BL1 and BL2. At any time, the device boot phase can be one of (BL1), BL2, or main firmware. 

Unfortunately, some of the information previously available in main firmware are not available in BL1/2 phase (device not fully initialized yet):
- The PING command is available in all phases, and can be used reliably to get the device generation
- Device ID is only available in main firmware phase
- Variant is determined from device ID, and can only be determined in main firmware as well

lsgunth: Ok, well lets only fall back on the ping command if we can't get a sensible device ID ...

xyan264: If device is not in main firmware phase, function 'get_device_id()' would fail miserably. We have to avoid calling 'get_device_id()' in this case. That's why we want to get current boot phase first. 

I agree this function name is not accurate, something like get_info might be a better fit.

lsgunth: I don't see how it will fail "miserably"... it returns a value from a register. Worst case the value is not correct, and it's easy to detect that sucha value is not in our table.

If the firmware is crashing or anything like that due to reading that register that is a huge problem that needs to be fixed in the firmware.

xyan264: I see what you mean. I will change the code to get device ID first, and fall back to ping if it fails.

PR #136, opened 2019-12-10
Fix initialization error while device in BL2 boot phase

xyan264:
Use 'switchtec_get_device_info' to determine device generation in BL2.


lib/recovery.c
```diff
@@ -0,0 +1,107 @@
+/*
+ * Microsemi Switchtec(tm) PCIe Management Library
+ * Copyright (c) 2017, Microsemi Corporation
```
kelvin-cao: 2019?

lib/recovery.c
```diff
@@ -0,0 +1,107 @@
...
+#include "switchtec/mrpc.h"
+#include "switchtec/errors.h"
+#include <unistd.h>
+
+#include <errno.h>
+#include <stdio.h>
+#include <string.h>
+
+#define IS_UNSUPPORTED_MRPC(x) ((x & (~SWITCHTEC_ERRNO_MRPC_FLAG_BIT)) == \
+				ERR_MPRC_UNSUPPORTED)
```
kelvin-cao: I agree with Logan that we have a common macro like ERRNO_ MRPC() to get error, with which we can also test other errors. And it seems it's a common macro which better be placed in a .h file.

lib/recovery.c
```diff
@@ -0,0 +1,107 @@
...
+			      uint32_t *reply_dw,
+			      enum switchtec_boot_phase *phase,
+			      enum switchtec_gen *gen,
+			      enum switchtec_rev *rev)
+{
+	int ret;
+	struct ping_reply {
+		uint32_t stat;
+		uint32_t reply;
+	} r;
```
kelvin-cao: Maybe we can have a more meaningful name?

lib/recovery.c
```diff
@@ -0,0 +1,107 @@
...
+ * @param[in]  dev	Switchtec device handle
+ * @param[in]  ping_dw	Ping data
+ * @param[out] reply_dw	Ping reply data (2's complement of ping data)
+ * @param[out] phase	The current boot phase
+ * @param[out] gen	Device generation
+ * @param[out] rev	Device revision
+ * @return 0 on success, error code on failure
+ */
+int switchtec_get_device_info(struct switchtec_dev *dev, uint32_t ping_dw,
+			      uint32_t *reply_dw,
```
kelvin-cao: It seems like the ping_dw and reply_dw is not necessary as parameters if this API is to get the hw device info.

lib/recovery.c
```diff
@@ -0,0 +1,107 @@
...
+		hw_stat = le32toh(r.stat);
+		*phase = hw_stat & 0xff;
+		*rev = (hw_stat >> 8) & 0x0f;
+		*gen = map_to_gen((hw_stat >> 12) & 0x0f);
+	} else {
+		if (IS_UNSUPPORTED_MRPC(errno)) {
+			*reply_dw = ~ping_dw;
+			*phase = SWITCHTEC_BOOT_PHASE_FW;
+			*gen = SWITCHTEC_GEN3;
+			*rev = SWITCHTEC_REVA;
```
kelvin-cao: SWITCHTEC_REV_UNKNOWN?

lib/recovery.c
```diff
@@ -0,0 +1,107 @@
...
+			    sizeof(ping_dw_le),
+			    &r, sizeof(r));
+	if (ret == 0) {
+		*reply_dw = le32toh(r.reply);
+		hw_stat = le32toh(r.stat);
+		*phase = hw_stat & 0xff;
+		*rev = (hw_stat >> 8) & 0x0f;
+		*gen = map_to_gen((hw_stat >> 12) & 0x0f);
+	} else {
+		if (IS_UNSUPPORTED_MRPC(errno)) {
```
kelvin-cao: Use 'else if...else' to avoid embeded 'if...else...'?

lib/recovery.c
```diff
@@ -0,0 +1,107 @@
...
+			*reply_dw = ~ping_dw;
+			*phase = SWITCHTEC_BOOT_PHASE_FW;
+			*gen = SWITCHTEC_GEN3;
+			*rev = SWITCHTEC_REVA;
+
+			errno = 0;
+		} else {
+			return -1;
```
kelvin-cao: We need to set errno on error.

kelvin-cao: I cannot think of how we can reach this condition since even gen3 should return unsupported. I just thought we should use a common error code to all cases that neither return 0 nor return unsupported, but it should be ok to pass the errno derived from swtichtec_cmd. So now I'd think it's ok to leave it alone.

lib/recovery.c
```diff
@@ -0,0 +1,107 @@
...
+ * @param[out] phase	The current boot phase
+ * @param[out] gen	Device generation
+ * @param[out] rev	Device revision
+ * @return 0 on success, error code on failure
+ */
+int switchtec_get_device_info(struct switchtec_dev *dev, uint32_t ping_dw,
+			      uint32_t *reply_dw,
+			      enum switchtec_boot_phase *phase,
+			      enum switchtec_gen *gen,
+			      enum switchtec_rev *rev)
```
kelvin-cao: We need check for NULL on input parameter.

lib/switchtec.c
```diff
@@ -122,27 +122,33 @@ static const struct switchtec_device_id switchtec_device_id_tbl[] = {
 static int set_gen_variant(struct switchtec_dev * dev)
 {
 	const struct switchtec_device_id *id = switchtec_device_id_tbl;
+	int ret;
+	enum switchtec_boot_phase phase;
+	enum switchtec_gen gen;
+	enum switchtec_rev rev;
```
kelvin-cao: It seems it's not used.

lib/recovery.c
kelvin-cao: This API is a general one, I think we should move it to lib/switchtec.c

lsgunth:
I agree with all of Kelvin's comments. Besides those it looks good to me.

xyan264:
Updated based on the comments above. Please review again.

lib/switchtec.c
```diff
@@ -718,6 +721,70 @@ int switchtec_log_to_file(struct switchtec_dev *dev,
...
+	if (ret == 0) {
+		dev_info = le32toh(reply.dev_info);
+		if (phase)
+			*phase = dev_info & 0xff;
+		if (rev)
+			*rev = (dev_info >> 8) & 0x0f;
+		if (gen)
+			*gen = map_to_gen((dev_info >> 12) & 0x0f);
+	} else if (ERRNO_MRPC(errno) == ERR_MPRC_UNSUPPORTED) {
+			if (phase)
```
kelvin-cao: indent

PR #137, opened 2019-12-10
Fix I2C write timing for BL2 boot phase.

xyan264:
With the introduction of BL1/2 boot phase, the I2C write timing needs to be changed as well.

Unlike main firmware which is based on RTOS, the BL1/2 firmware is implemented in a single-thread, poll-based fashion. This increases system response delay when firmware is taking lengthy operations (such as writing/erase flash).

A direct result of this timing difference is that i2c_gas_write_status_get() has potential to timeout and needs to be called repeatedly, and the delays needs to be increased as well. This check in reflect these changes.

There should be margin for optimizing the delays and number of retries, which can be explored in later revisions.

### DIFF
--- a/lib/platform/linux-i2c.c
+++ b/lib/platform/linux-i2c.c
@@ -59,6 +59,7 @@ struct switchtec_i2c {
 #define CMD_GAS_READ  0xE9
 
 #define MAX_RETRY_COUNT  100
+#define MAX_STATUS_GET_RETRY  200
 #define PEC_BYTE_COUNT  1
 #define TWI_ENHANCED_MODE  0x80
 #define GAS_TWI_MRPC_ERR  0x20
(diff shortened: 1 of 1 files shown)


lib/platform/linux-i2c.c
```diff
@@ -373,18 +374,29 @@ static void i2c_gas_write(struct switchtec_dev *dev, void __gas *dest,
...
+	uint8_t status_retry_count = 0;
 
 	do {
+		status_retry_count = 0;
 		tag = get_tag(idev);
 		i2c_gas_data_write(dev, dest, src, n, tag);
-		status = i2c_gas_write_status_get(dev, tag);
-		if (status == 0 || status == GAS_TWI_MRPC_ERR)
+		do {
+			usleep(5000);
```
lsgunth: Again, a 5ms sleep seems like it will slow things down significantly....

lib/platform/linux-i2c.c
```diff
@@ -373,18 +374,29 @@ static void i2c_gas_write(struct switchtec_dev *dev, void __gas *dest,
...
+			status = i2c_gas_write_status_get(dev, tag);
+			if (status == 0 || status == GAS_TWI_MRPC_ERR)
+				break;
+			status_retry_count++;
+		} while (status_retry_count < MAX_STATUS_GET_RETRY);
```
lsgunth: The nested retry loop seems nasty.... At the very *least* push it into it's own function.

xyan264:
The existing code works fine with Main Firmware, but with BL2 firmware, we found function i2c_gas_write_status_get() times out while firmware taking a lengthy operation (e.g., reading a relative large amount of data from a slow flash). The inner for-loop is a way to give firmware more time to process the command and return status.

lib/platform/linux-i2c.c
xyan264: Agree.

lsgunth:
Ok,so push the inner loop into it's own function and make sure it only delays after the first failure.

Please add a comment that the extra retry is only typically necessary for the BL2 firmware.

Also, please add the description to the commit message so we get a description if/when someone does a git blame.

xyan264:
Updated, please review again.

lib/platform/linux-i2c.c
```diff
@@ -367,24 +368,49 @@ static uint8_t i2c_gas_write_status_get(struct switchtec_dev *dev,
...
+static int i2c_gas_write_status_wait(struct switchtec_dev *dev,
+				     uint8_t tag)
+{
+	int status_retry_count = 0;
+	uint8_t status;
+
+	do {
+		status = i2c_gas_write_status_get(dev, tag);
+		if (status == 0 || status == GAS_TWI_MRPC_ERR)
+			break;
```
lsgunth: Hold on a second... I just noticed there's already a retry loop inside i2c_gas_write_status_get() so we have three levels of retry loops... One must be redundant. Maybe we just need to catch another error inside that retry loop...

In any case, you could return here instead of breaking and deciding whether to return an error, but you'd also probably want to return an error in the case of GAS_TWI_MRPC_ERR....

xyan264:
Closed due to my internal reasons. Will continue the discussion in another PR.

PR #139, opened 2019-12-11
Fix I2C write timing for BL2 boot phase.

xyan264:
BL2 firmware is implemented as a single thread, so extra delay is required to get write status while firmware is performing a lengthy operation (e.g., accessing flash memory).

### DIFF
--- a/lib/platform/linux-i2c.c
+++ b/lib/platform/linux-i2c.c
@@ -59,6 +59,7 @@ struct switchtec_i2c {
 #define CMD_GAS_READ  0xE9
 
 #define MAX_RETRY_COUNT  100
+#define MAX_STATUS_GET_RETRY  50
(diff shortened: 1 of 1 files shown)


lib/platform/linux-i2c.c
```diff
@@ -343,27 +344,27 @@ static uint8_t i2c_gas_write_status_get(struct switchtec_dev *dev,
 	msgs[1].buf = rx_buf;
 
 	do {
-		usleep(2000);
 		ret = ioctl(idev->fd, I2C_RDWR, &rwdata);
 		if (ret < 0)
-			goto i2c_ioctl_fail;
+		{
```
lsgunth: Nit: braces for if's go on the if line.

xyan264: Updated.

PR #141, opened 2019-12-12
Add mfg submenu and ping command

xyan264:


lib/mfg.c
```diff
@@ -0,0 +1,73 @@
...
+#include "config.h"
+
+/**
+ * @brief Ping a device with a 32-bit data.
+ * @param[in]  dev	Switchtec device handle
+ * @param[in]  ping_dw	Ping data
+ * @param[out] reply_dw	Ping reply data (2's complement of ping data)
+ * @return 0 on success, error code on failure
+ */
+int switchtec_ping(struct switchtec_dev *dev, uint32_t ping_dw,
```
kelvin-cao: It seems it's not necessary to have this function as we already have switchtec_get_device_info(), in which we should put the test of ping_dw internally, which means it should only succeed with a successful test of internally generated ping_dw.

lsgunth: That sounds like a good idea!

cli/mfg.c
```diff
@@ -0,0 +1,112 @@
...
+
+	return 0;
+}
+
+static const struct cmd commands[] = {
+	{"ping", ping, "Ping firmware and get current boot phase"},
+};
+
+static struct subcommand subcmd = {
+	.name = "mfg",
```
lsgunth: Are there plans to add more mfg commands? I find it a bit distasteful to have a submenu for one command...

xyan264: There are about 10 commands under this submenu, those will be added later.

lib/mfg.c
xyan264: Updated based on Kelvin's suggestions.

cli/common.h
```diff
@@ -55,4 +55,19 @@ int pax_handler(const char *optarg, void *value_addr,
...
+			.complete="/dev/switchtec*", \
+			.env="SWITCHTEC_DEV", \
+			.help="switchtec device to operate on. can be any of:\n" \
+			" * a device path (/dev/switchtec0)\n" \
+			" * an index (0, 1, 2)\n" \
+			" * a pci address (3:00.1)\n" \
+			" * An I2C path with slave address (/dev/i2c-1@0x20)\n" \
+			" * A UART path (/dev/ttyUSB0)\n" \
+	}
+
```
lsgunth: You shouldn't need to duplicate this...

```
#define DEVICE_OPTION    DEVICE_OPTION_NO_PAX, \
         { \
			"pax", 'x', .cfg_type=CFG_CUSTOM, \
			.value_addr=&cfg.dev, \
			.argument_type=required_argument, \
			.custom_handler=pax_handler, \
			.env="SWITCHTEC_PAX", \
			.help="PAX ID within a PAX fabric. Only valid on " \
			"Switchtec PAX devices" \
	}

```

xyan264: Good idea, updated.

PR #143, opened 2019-12-14
Command description updates

brian-mckenzie:
Various command descriptions, command argument descriptions, and other text outputs are incorrect, have spelling/grammar errors, or are inconsistent. Also, command descriptions are duplicated in both the commands[] arrays and the command functions themselves (for passing the description to argconfig_parse()).

The following changes are being applied to the main, GAS, and fabric commands:
-statically define all command descriptions and use these for the commands[] array and argconfig_parse() calls
-update command and argument descriptions as appropriate for correctness, spelling/grammar, and consistency

The following changes are also being made:
-fix some argv[] usages in main.c error outputs
-fix a typo in the AER_IN_VEP event description


cli/fabric.c
```diff
@@ -37,9 +37,19 @@
...
+static const char * const cmd_desc_topo_info = "Show topology information";
+static const char * const cmd_desc_gfms_bind = "Bind the EP (function) to the specified host";
+static const char * const cmd_desc_gfms_unbind = "Unbind the EP (function) from the specified host";
+static const char * const cmd_desc_gfms_dump = "Dump the GFMS database";
+static const char * const cmd_desc_route = "Show routing information";
+static const char * const cmd_desc_port_control = "Control a port";
+static const char * const cmd_desc_portcfg_show = "Display a port's configuration information";
+static const char * const cmd_desc_portcfg_set = "Configure a port";
+static const char * const cmd_desc_gfms_events = "Display GFMS event information";
+
```
lsgunth: Nak -- please keep the descriptions with the commands. Then people will be more likely to see them if they need to update them.

cli/fabric.c
```diff
@@ -37,9 +37,19 @@
...
 static int gfms_bind(int argc, char **argv)
 {
-	const char *desc = "Unbind the EP(function) to the specified host";
+	const char *desc = cmd_desc_gfms_bind;
```
lsgunth: If you want to move the description outside the function (but still right above it) that's fine. But ]then you wouldn't need the 'desc' variable, just use the 'cmd_desc_x' variable directly.

cli/fabric.c
```diff
@@ -209,7 +219,7 @@ static int portcfg_set(int argc, char **argv)
 		DEVICE_OPTION,
 		{"phys_port_id", 'p', "NUM", CFG_INT,
 		 &cfg.phys_port_id, required_argument,
-		 "physical port id", .require_in_usage = 1},
+		 "Physical port ID", .require_in_usage = 1},
```
lsgunth: In general, help text is not a sentence and is not capitalized. Most aren't so I wouldn't change it to this.

cli/main.c
```diff
@@ -449,9 +476,9 @@ static int bw(int argc, char **argv)
 		DEVICE_OPTION,
 		{"time", 't', "NUM", CFG_POSITIVE, &cfg.meas_time,
 		  required_argument,
-		 "measurement time, in seconds"},
+		 "measurement time in seconds"},
 		{"verbose", 'v', "", CFG_NONE, &cfg.verbose, no_argument,
-		 "print posted, non-posted and completion results"},
+		 "print posted, non-posted, and completion results"},
```
lsgunth: I don't think the oxford comma adds value here.

cli/fabric.c
brian-mckenzie: Note I'll be changing these from variables to #defines since this compiles fine in Windows but not Linux. Apparently certain versions of certain compilers have a limitation where they don't recognize such variable declarations as constant (even though they're declared with const), so the compiler errors out when assigning them to struct elements that expect const.

Do you still think it's better to keep the definitions right above the associated functions? Typically #define's are grouped together at the top of the file, before any function definitions, as this keeps things more organized. It seems like that's the switchtec-user convention too.

cli/fabric.c
brian-mckenzie: Good point, I'll remove all the unnecessary 'desc' variables. See my question above about moving the description.

cli/fabric.c
brian-mckenzie: I agree it would probably be better lowercase, but all other help text is capitalized on the first letter, so I did this to be consistent.

cli/main.c
brian-mckenzie: Sure, I'll remove it.

cli/fabric.c
lsgunth: What? That's mad. I know GCC can do this fine. #define's are not really appropriate here.

This works fine for me on any version of GCC/clang, I try: 
https://godbolt.org/z/9iUEAD

You might have an unnecessary second const there that doesn't really make sense in that context; but it should still compile.

And yes, still keep it by the function.

cli/fabric.c
lsgunth: Looking at this patch it looks like most of it is not capitalised... It's probably just same newer help texts that were added that are capitalised. I'd standardise on lower case.

cli/fabric.c
brian-mckenzie: It fails when used like this: https://godbolt.org/z/y62_5o
That's how we're using it here. GCC 7.3 fails, and GCC 8.1+ succeeds.

lsgunth: Gross. I'd just leave it the way it was then instead of dealing with #define mess. 

Technically the desc in the function is supposed to be more detailed than the short one-line description in the list. They were never supposed to be the same.

cli/fabric.c
brian-mckenzie: Ok so it looks like the top-level command descriptions are uppercase for main, fabric, "version", and "help", but not gas. The subcommand descriptions are capitalized for fabric but not main and gas. And subcommand choices (e.g. for evcntr-setup events) seem to all be capitalized. I'll change everything to lowercase. Does that sound good?

lsgunth: Yes

cli/fabric.c
brian-mckenzie: They were all the same the way things were before these changes, except for minor things like a missing "the" or a spelling mistake. And some of the descriptions were entirely incorrect in one of the instances but not the other as they were pasted from other commands. If I revert these back then I should at least fix any errors, but that would basically duplicate each description. The duplication is probably what led to certain instances having a correction but others not, which is why duplication should typically be avoided.

If the descriptions in the function were mostly different than the commands[] descriptions as originally intended then I agree we should leave these as they were, but since they're basically all duplicated I think the #defines would be the safer option.

lsgunth: Ok, well do what you feel is best.

PR #146, opened 2019-12-17
Fix 'fw-update' for BL2 boot phase

xyan264:
The existing switchtec_fw_write_file() function does not work in BL2 phase because BL2's fw-update feature uses a different MRPC command ID. 

We fix this by extending the switchtec_fw_write_file() function to take MRPC command ID parameter.


cli/main.c
```diff
@@ -1377,8 +1394,9 @@ static int fw_update(int argc, char **argv)
 	}
 
 	progress_start();
-	ret = switchtec_fw_write_file(cfg.dev, cfg.fimg, cfg.dont_activate,
-				      cfg.force, progress_update);
+	ret = switchtec_fw_write_file_ex(cfg.dev, rpc_cmd,
```
lsgunth: I don't like the structure here... I don't think any users trying to update firmware should have to deal with raw MRPC commands or boot phases.

Instead of creating a bunch of new public functions, with the _ex prefix, create them as static (and instead of naming them with an _ex prefix, it's common to just prepend two underscores to the name).

Then have switchtec_fw_write_file() (and others) check the boot phase before performing the underlying operation. Then you don't have to make fw_update() or other callers, any more complicated; and our overall API isn't made more complicated either.

Also, if we are checking the boot phase often, it *may* make sense just to load this information and cache it on switchtec_open()...

xyan264: There are several constraints at play here:
- existing API should continue to be supported
- new BL2 fw-update feature only differs by command ID, we want minimal duplicate code
- function _switchtec_fw_wait()_ and _switchtec_fw_dlstatus()_ are called from _for_ loops, for performance considerations, we want to avoid checking boot phase in these functions
- from API perspective, it is possible (though less likely) that user open device at one boot phase, and do fw-update in another boot phase, so the cached boot phase becomes stale

I agree that users should not have to deal with MRPC commands and boot phases. We can move the boot phase checking and MRPC command to _switchtec_fw_write_file()_. But the underlying functions _switchtec_fw_wait()_ and _switchtec_fw_dlstatus()_ has to be extended to accept MRPC command ID.

xyan264: Updated to reflect my suggestion above. The function name _switchtec_fw_wait_ex()_ and _switchtec_fw_dlstatus_ex()_ can be changed as appropriate. 

Please take a look again, thanks.

lsgunth: I don't really see any issue here.

Probably one mistake is that switchtec_fw_wait() and switchtec_fw_dlstatus() shouldn't have ever been exported. They aren't used by anything. 

Probably the easiest solution is to cache the boot phase in the device structure. 

Another solution is to just have the public switchtec_fw_wait() and switchtec_fw_dlstatus() call the private version which take the mrpc command. Though in this case, using them will probably be pretty broken.

Just avoid introducing a bunch of new complicated public APIs for this...

xyan264: I have updated code based on the 2nd solution above, considering that a phase change is still possible between _switchtec_open()_ and fw-update (to confirm by Kelvin). 

No new API function is introduced. The public switchtec_fw_wait() and switchtec_fw_dlstatus() just calls the private version with known MRPC command. 

If this update is OK, I'll wait for comments from Kelvin.

cli/main.c
```diff
@@ -1347,6 +1348,19 @@ static int fw_update(int argc, char **argv)
...
+	ret = switchtec_get_device_info(cfg.dev, &phase_id, NULL, NULL);
+	if (ret < 0) {
+		switchtec_fw_perror("firmware update", ret);
+		return ret;
+	}
+	if (phase_id == SWITCHTEC_BOOT_PHASE_BL1) {
+		fprintf(stderr,
+			"This command is only available in BL2 or Main Firmware!\n");
+		fprintf(stderr,
+			"Use 'fw-transfer' instead to download a BL2 image.\n");
```
kelvin-cao: 'mfg fw-transfer', to 'transfer' instead of download?

lib/fw.c
```diff
@@ -151,16 +145,22 @@ int switchtec_fw_dlstatus(struct switchtec_dev *dev,
...
- * indicates it's INPROGRESS. Sleeps 5ms between each poll.
  */
-int switchtec_fw_wait(struct switchtec_dev *dev,
-		      enum switchtec_fw_dlstatus *status)
+int switchtec_fw_dlstatus(struct switchtec_dev *dev,
+			  enum switchtec_fw_dlstatus *status,
+			  enum mrpc_bg_status *bgstatus)
+{
+	return __switchtec_fw_dlstatus(dev, MRPC_FWDNLD, status, bgstatus);
+}
```
kelvin-cao: This way, the switchtec_fw_dlstatus() API won't work in BL2.

lib/fw.c
```diff
@@ -151,16 +145,22 @@ int switchtec_fw_dlstatus(struct switchtec_dev *dev,
...
+
+static int __switchtec_fw_wait(struct switchtec_dev *dev,
+			       enum mrpc_cmd rpc_cmd,
```
kelvin-cao: It looks a bit odd to use 'rpc' in this context, as we use 'mrpc' more, maybe cmd is good enough.

lib/fw.c
```diff
@@ -185,6 +185,21 @@ int switchtec_fw_wait(struct switchtec_dev *dev,
...
+ * @return 0 on success, error code on failure
+ *
+ * Polls the firmware download status waiting until it no longer
+ * indicates it's INPROGRESS. Sleeps 5ms between each poll.
+ */
+int switchtec_fw_wait(struct switchtec_dev *dev,
+		      enum switchtec_fw_dlstatus *status)
+{
+	return __switchtec_fw_wait(dev, MRPC_FWDNLD, status);
+}
```
kelvin-cao: Same as switchtec_fw_dlstatus().

cli/main.c
kelvin-cao: @lsgunth We will have another mfg command/API boot-resume which makes the firwmare step into another boot phase (BL1->BL2->Main). As different boot phase runs different F/W image with different command set, so the run time environment across this API will be different. 

I think we have two options, either forcing the fd be invalid after a boot-resume (so user has to do close and re-open after the boot phase switch), and caching the boot phase on open; or get the boot phase locally to any function that may branch for different phase. I guess option 1 is probably better as other stuff like device node name and pax id may also change over a resume, continue using the old fd may cause unknown side effect.

cli/main.c
lsgunth: Wait why can't fw_update just do the right thing for a BL2 image? Why do we need to create a new command for it?

xyan264: Oh this message is printed in BL1 mode. 

BL1 (Bootloader 1) is implemented in an on-chip boot ROM and has limited capabilities, hence a different command for transferring image in BL1.

cli/main.c
xyan264: My vote goes to option 1 as well. If we can make sure user re-open the dev after switching boot phase, then our code can be much cleaner.

cli/main.c
lsgunth: Ok, so you require a different MRPC command/procedure... this does not at all justify a different user API and command line interface. This just means everyone using it, be it with a shell script or via the library have to have an if (BL1) case for no good reason.

cli/main.c
lsgunth: I think the complaint is now the unused switchtec_fw_wait() and switchtec_fw_dlstatus() are pretty broken. We could probably safely drop them though.

I think operations like a boot phase switch will need to have an exclusive lock on the device... Sounds like it will be problematic if anything else is using the device while it occurs. So on switchtec_open() we will need to get a shared flock, then on exclusive operations like a boot phase change we'd switch to an exclusive flock which ensures no other processes are using it. Thus caching the boot phase is now an option because it can't change while you have the device open unless your own process changed it.

I'm actually a bit surprised the firmware can change phase without requiring a host to be rebooted. It sounds like a pretty drastic thing.

cli/main.c
xyan264: The BL1 'fw-transfer' command has a quite different behavior from 'fw-update' in BL2 or MAIN. 

With 'fw-transfer', you can only send BL2 image to device RAM, no writing to flash or anything. 
'fw-update' supports downloading any (BL2, MAIN, CFG) image to flash.

This line was previously added at request from our APPS team, so users trying to update FW image in BL1 will get a bit more information about what else they can try, instead of just a cold error "command not available."

The 'fw-transfer' command in 'mfg' submenu will be checked in later.

If this line creates more confusion, then maybe we can remove it.

cli/main.c
xyan264: Code updated based on Kelvin's option 2. @lsgunth please take a look.

cli/main.c
lsgunth: I still don't see why the existing fw update API can't just see that we're in BL1 and do the right thing.... If the CLI prints some warnings or something saying we're in BL1 and therefore the new image is only stored in ram, then that seems acceptable. But the API should still be simple and take care of these details for the user.

cli/main.c
lsgunth: Yes, this looks much cleaner, but I'm still against creating a second firmware update API and CLI command.

cli/main.c
xyan264: A little background info:
_One use case for mfg commands is recovering from boot failure due to corrupt partitions. In this case, user can set a jumper pin on the board, and the device will boot into BL1 (always available since it's in on-chip ROM). 

From BL1, user can do 'fw-transfer' to send a BL2 image to RAM, then do 'fw-execute' to run the image ('fw-transfer' does not automatically execute the image). The device will boot into BL2 phase, and then user can do 'fw-update' to update corrupted partitions._

Previously we have 'mfg fw-transfer' for BL1, and 'fw-update' for BL2/MAIN. From the above discussion, it seems we can move 'fw-transfer' and 'fw-execute' command from **mfg** submenu to main menu, and merge 'fw-transfer' into 'fw-update': print a message that this command is running in BL1 so the image is sent to RAM instead of flash. 

Note that in BL1, user still need to send 'fw-execute' command to execute the image.

@lsgunth let me know what you think.

lsgunth: Yes, merge fw-transfer and fw-update...

I think fw-execute could still be in the mfg menu... Though I don't feel that strongly about it. You must be very clear that fw-execute only applies to uploads during BL1, unless you have a way to support having it work after a regular firmware update.

Also, fw-execute sounds super dangerous... How does this interact with the host OS? Seems like it would need to reinitialize everything...

kelvin-cao: The fw-transfer transfers a BL2 image into the chip ram, and the following fw-execute jumps to the tranfered BL2 image and executes. Thereafter, the BL2 fw starts to run in a while(1) loop fashion with limited commands supported aiming to do some recovery related stuff that user may ask it to do by MRPC commands. 

I think the fw-transfer and fw-execute are two steps of one operation (run a BL2 fw), therefore it makes more sense to have them grouped in the same mfg sub-menu. For the API, I'm good if we have them implemented in the current fw update APIs.

So I'd prefer we have the fw-transfer as a seperate mfg command with only needed option (none?), without being mixed with the fw-update which have many other options which are meaningless to the transfer operation. Meanwhile we have the same API supporting both download and transfer.

lsgunth: I disagree. fw-transfer and fw-update do exactly the same thing (push firmware to the device). Having two different commands is confusing to the user. It's fine to ignore or warn if they specify arguments that don't apply in the situation. The API will be almost exactly the same and would also be confusing on which is used when. Better to avoid the extra code and documentation burden and just adapt the existing command.

xyan264: Here is a brief summary, based on the current implementation:
- at API level, there is only one function _switchtec_fw_write_file()_, both _fw-transfer_ and _fw-update_ use this function to send FW image to device
- at CLI command level, @lsgunth prefer combining _fw-transfer_ into _fw-update_, while @kelvin-cao wants to keep them separate.

Based on what I see so far we all like the idea of keeping the same API function _switchtec_fw_write_file()_ for BL1/2 & MAIN. So at this point we can say this API definition is decided. 

At CLI command level, I agree with Kelvin that the _fw-transfer_ command should be grouped with _fw-execute_, in mfg submenu. From a user perspective, it would be weird to do FW transfer using command from Main Menu, then do _fw-execute_ from mfg submenu. 

Keeping these 2 commands separate would mean some duplicate code at CLI level, but the _fw-transfer_ command is much simpler so there should not be that much of duplicate code.

@lsgunth what do you think?

lsgunth: If you absolutely must have a separate CLI command I would expect *very* good documentation in the 'switchtec help' for both commands. It must explain when each command is the correct one to use and the difference between them. You must also define your terms. Saying 'fw-transfer' can only be used in "BL1" will not be sufficient because most people will not know what BL1 is or when it applies.  

This would be a case where the description in the command should not match the description in the list because the former will need to be quite long and detailed.

xyan264: @lsgunth code updated with more information in 'fw-update' descriptions. Please take a look.

cli/main.c
```diff
@@ -1354,6 +1358,14 @@ static int fw_update(int argc, char **argv)
 	if (type < 0)
 		return type;
 
+	if (switchtec_boot_phase(cfg.dev) == SWITCHTEC_BOOT_PHASE_BL1) {
```
kelvin-cao: I'd suggest we have helper functions like switchtec_in_bl1/bl2/main() or similar to deal with the boot phase check.

cli/main.c
```diff
@@ -1317,8 +1317,12 @@ static int fw_update(int argc, char **argv)
 {
 	int ret;
 	int type;
-	const char *desc = "Flash the firmware with a new image";
-
+	const char *desc = "Flash device with a new firmware image\n\n"
+			   "NOTE - A device can be in one of these three boot phases: \n"
+			   "Bootloader1 (BL1), Bootloader2 (BL2) or MAIN FIRMWARE (FW).\n\n"
```
kelvin-cao: MAIN FIRMWARE (FW) -> MAIN FIRMWARE (MAIN)?

cli/main.c
kelvin-cao: I agree we can drop the switchtec_fw_dlstatus() and switchtec_fw_wait() from the API list, with a major version update. I don't see any chance user should use it.

cli/main.c
lsgunth: I disagree. That just sounds like a lot of unnecessary boiler plate code that doesn't add any additional clarity.

cli/main.c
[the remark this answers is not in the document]
xyan264: Here is my understanding of the mfg submenu:
- Unlike main menu, whose commands are more or less intuitive to most users, the mfg submenu commands are used by advanced users during manufacturing process. Users are directed to these commands by a Knowledge-Base page (e.g., How to Recover Device from Corrupt Partitions, or How to Generate Secure Setting Files and Load to Device). These KB pages are available online and gives background on what BL1/2 are, how to boot to BL1/2, and the procedure and commands to use at each step.
- So the help information in mfg submenu should provide enough information to help user if he/she got stuck with a command in particular (e.g., missing parameters, wrong boot phase, wrong file format), but leave the background knowledge to the KB page.
- This is similar to GAS submenu, where we did not detail which register/bit controls what. User need to find those info elsewhere, before they can use the commands there. Of course GAS is a more extreme case where you basically cannot use the commands without background knowledge, while mfg submenu command is more in the middle: you can play with it but background knowledge is still required.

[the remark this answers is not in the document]
lsgunth: No, that's a horrible idea. All commands should be largely understandable from the help text regardless of whether the user got send there from more detailed documentation. Saying you don't have to document something because it's documented somewhere non-obvious is wrong.

The analogy with the GAS sub-menu is flawed. It provides an interface to raw registers that no one should ever really use and is documented as such.

If you want to document the mfg submenu as more advanced commands, fine; but that does not mean you don't need proper documentation to say what the commands are used for and how to use this. This is especially so because we've made things more confusing with having a second way to send firmware to the device and thus the documentation must be *crystal* clear on the differences between them without assuming the user's read some random knowledge base article somewhere else.

Additionally, just because there will be documentation somewhere in some non-obvious knowledge- base doesn't mean making the procedure simpler with a fw-recover command would not be valuable.

cli/main.c
kelvin-cao: I'm not feeling very strong to have it. However, I see a check of this like with current code of one level indent consumes 72 character in a line. With a helper function like switchtec_in_bl1() we will have less worry about the 80-character issue. And with helper like switchtec_in_bl1_main(), which is the check for some MRPC/CMD, we will further reduce the code and improve the readability.

cli/main.c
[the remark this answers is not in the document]
xyan264: Agree on adding more detailed descriptions of boot phases. I have put the descriptions in common.h so they can be easily included in other commands if needed. Please take a look.

Renaming 'fw-transfer' to 'fw-transfer-bl2' is a good idea, will be applied when we check in the corresponding code. 

About having one commands that does transferring BL2, updating BL2/MAIN and all: there are many branches in recovery process: after transferring BL2, user can choose to switch to UART interface for future commands; also user have the option to update BL2/CFG/MAIN to active/non-active partition, and choose to boot from active/non-active partitions using CFG from active/non-active partition. Using one command means user have to set many command options for this command. It would be much easier if we offer them as individual commands instead, so user can do mix and match to suit their needs.

About keeping a copy of baseline BL2 image that works for most cases, as firmware and utility code are both constantly evolving, it is hard to find a version of BL2 that support most cases, for different hardware revisions. Over time, a baseline BL2 image will slowly become obsolete, if we don't update them along with the utility code.

@lsgunth let me know what you think.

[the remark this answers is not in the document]
lsgunth: Well my comment for a fw-recover command would have been in addition to the fw-transfer command. Just a thought to make it easier on people.

cli/main.c
lsgunth: The goal isn't to have the fewest characters in a line. If there were a number of cases where we had to do the compare that weren't fitting on the line then maybe, but currently we have none.

PR #147, opened 2019-12-18
Add 'config-set' command to mfg submenu

xyan264:


lib/crc32.h
```diff
@@ -0,0 +1,44 @@
...
+/*
+* Standard CRC-32 Polynomial of:
+*      x^32+x^26+x^23+x^22+x^16+x^12+x^11+x^10+x^8+x^7+x^5+x^4+x^2+x^1+1
+*/
+#define CRC32_POLYNOMIAL        (0x04C11DB7)
+#define CRC32_INIT_REMAIDER     (0xFFFFFFFF)
+#define CRC32_FINAL_XOR_VALUE   (0xFFFFFFFF)
+
+uint32_t pmc_crc32(const uint8_t *msg_ptr, uint32_t byte_cnt,
+		   uint32_t oldchksum, int init, int last);
```
lsgunth: As this is the only thing for this header, I'd suggest we add it to an existing header. switchtec_priv.h or switchtec.h maybe.

Also, please add the CRC in a separate commit with it's own description.

lib/crc32.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include <stdint.h>
+#include <stdlib.h>
+
+#define CRC32_WIDTH             (8 * sizeof(uint32_t))    /* in bits  */
+
+/*
+* Standard CRC-32 Polynomial of:
+*      x^32+x^26+x^23+x^22+x^16+x^12+x^11+x^10+x^8+x^7+x^5+x^4+x^2+x^1+1
+*/
```
lsgunth: We appear to be implementing "CRC-32/CKSUM" as documented [here](http://reveng.sourceforge.net/crc-catalogue/all.htm)... We should note that in a comment and perhaps as part of the function name... "Standard CRC-32" isn't really a description of it as there are a bunch of CRC-32 algorithms that could be CRC-32.

lib/mfg.c
```diff
@@ -145,6 +147,51 @@ int switchtec_security_config_get(struct switchtec_dev *dev,
...
+	struct setting_data {
+		uint64_t cfg;
+		uint32_t pub_key_exponent;
+		uint8_t rsvd[4];
+	} sd;
+	uint64_t L64 = 0;
+
+	memset(&sd, 0, sizeof(sd));
+
+	sd.cfg |= setting->jtag_lock_after_reset? 0x40 : 0;
```
lsgunth: 0x40, etc should probably be an enum somewhere.

lib/mfg.c
```diff
@@ -226,3 +273,70 @@ int switchtec_active_image_index_set(struct switchtec_dev *dev,
...
+		uint64_t cfg;
+		uint32_t pub_key_exponent;
+		uint8_t rsvd[36];
+	};
+	struct setting_file {
+		struct setting_file_header header;
+		struct setting_file_data data;
+	} file_data;
+
+	rlen = fread(&file_data, 1, sizeof(file_data), setting_file);
```
lsgunth: NAK. I don't think introducing this custom binary file is a good idea in this day and age. What's going to be creating said file? It's just an extra giant pain for the users.

If it must be configured with a file, we should definitely use a plain text file of some kind of format. I think I'd prefer a simple key/value (ini/config like file) --  yaml or json would be the next best thing, but they aren't that great at being human writable.

xyan264: OK this needs a bit of background:
We do have a Chiplink tool for generating this binary file. The idea of forcing user to generate the file from Chiplink is that these settings will be written to chip's one-time programming (OTP) memory, so we want to make sure the settings are generated in a controlled way. 

Also to ensure integrity of the file, we add signature and CRC32 to the header.

So for users, they should already generated the file using Chiplink tool, and this function is provided to load the file content into a data structure which they can pass to secure_setting_set function.

lib/mfg.c
```diff
@@ -145,6 +147,51 @@ int switchtec_security_config_get(struct switchtec_dev *dev,
...
+int switchtec_security_config_set(struct switchtec_dev *dev,
+				  struct switchtec_security_cfg_set *setting)
+{
+	int ret;
+	struct setting_data {
+		uint64_t cfg;
+		uint32_t pub_key_exponent;
+		uint8_t rsvd[4];
+	} sd;
+	uint64_t L64 = 0;
```
kelvin-cao: We should probably avoid using capitalized var name if no special reason.

lib/mfg.c
```diff
@@ -226,3 +273,70 @@ int switchtec_active_image_index_set(struct switchtec_dev *dev,
...
+	};
+	struct setting_file {
+		struct setting_file_header header;
+		struct setting_file_data data;
+	} file_data;
+
+	rlen = fread(&file_data, 1, sizeof(file_data), setting_file);
+
+	if (rlen < sizeof(file_data))
+		return SWITCHTEC_SETTING_FILE_ERROR_LEN;
```
kelvin-cao: In this utility context, we use return value and errno for the error returning. Generally, we have two types of errors, system error and mrpc command errror. The system error is like what you are handling here, some check failure or system API failure, please use the standard errno values (like ENOMEM, EINVAL) for them instead of creating function specific ones; mrpc command error is what we get from the MRPC execution. For more detail, pls check this PR #132.

lib/crc32.h
xyan264: I can put this in common.h but since we already have a 'crc8.h' which is the same situation, I would like to maintain consistency for now.

All other suggestions are adopted and code updated. @lsgunth please have a look.

xyan264:
@lsgunth could you review the changes and comment?

lib/mfg.c
xyan264: @lsgunth 
- By 'this should be clear everywhere ... etc.', did you mean adding some comments to the function loading the chiplink file?
- By 'actual user friendly interface', did you mean you prefer using INI, YAML, JSON files?

Some background information: Chiplink is a Java-based GUI tool that generates configuration files for Switchtec device. From the GUI user can select the desired settings and click 'Generate' to generate a configuration file. The configuration file is a binary file with signature and CRC32 to ensure data integrity. 

The original intention was to ensure that user always generate configuration files using Chiplink tool, as these are critical settings written to device One-Time Programming (OTP) memory. We didn't expect user to read/modify these files in other ways. 

The use of binary format does pose one issue at least: in the case when user has multiple configuration files, there is no easy way to tell which configuration file has what settings. The human-readable INI, YAML or JSON file format is a great way to solve this problem, and is definitely easier to work with.

Unfortunately at this time the format of files generated by Chiplink has already been defined. For future format definitions, we will certainly push for more human-readable file formats, such as INI, YAML or JSON.

lsgunth: > @lsgunth
> 
>     * By 'this should be clear everywhere ... etc.', did you mean adding some comments to the function that loads the chiplink file?

I mean it should be clear in the function name:

switchtec_read_sec_cfg_chiplink_file()
 
>     * By 'actual user friendly interface', did you mean you prefer using INI, YAML, JSON files?

Yes. Even an interface that just lets you set these flags with command line options would be infinitely better.

> 
> Some background information: Chiplink is a Java-based GUI tool that generates configuration files for Switchtec device. From the GUI user can select the desired settings and click 'Generate' to generate a configuration file. The configuration file is a binary file with signature and CRC32 to ensure data integrity.

I'm aware of chiplink and I hate it. It's such a gigantic pain in the ass if you need to change a config file and managing the config files is awful. Not to mention the agony of trying to describe to someone else they need to change their configuration options and that they need this monstrosity software to do it. I've honestly just stopped using it a long time ago. I have a couple old config files with some very old firmware I use. Upgrading is too painful.

For config files, we were stuck with that legacy mistake by the firmware people -- however, I'd rather not perpetuate that pain with new features where we aren't forced to use a binary format .

> Unfortunately at this time the format of files generated by Chiplink has already been defined. For future format definitions, we will certainly push for more human-readable file formats, such as INI, YAML or JSON.

I'd push back against that. It's actively hostile to your users.

xyan264: I see your point, and agree with most of the statement above. 

At this time, all design have been finalized and we are at pre-release stage already, so changing file definitions requires coordination among different groups and could impact the release schedule. The (binary) file format has been defined thus far: configuration file, KMSK file (to be checked in later), signature data file (to be checked in later). 

I do agree that we should move away from binary format. We will do our best to make that happen for any future file format definitions.

lsgunth: Well  you can help by just implementing a command line version that takes options in addition to the command that takes the chiplink file.

xyan264: I talked to the team and it turns out the command line version had been proposed during design phase (which predates me). Considerations was given to making sure Chiplink is used to generate the files, as these settings are written to OTP and any error in the data could potentially brick the device. Chiplink, with its GUI, can perform more advanced checking on input data, ensuring the generated configuration is valid. Also, the binary data format was defined to discourage user from tempering with the setting files. 

At this stage the decision has been made and we can move forward with the binary format and without the command line version.

lib/crc32.h
```diff
@@ -0,0 +1,37 @@
...
+#ifndef CRC32_H
+#define CRC32_H
+
+#include <stdint.h>
+#include <stdlib.h>
+
+/*
+* This function implements CRC-32/CKSUM
+*/
+uint32_t crc32(const uint8_t *msg_ptr, uint32_t byte_cnt,
```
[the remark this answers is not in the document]
xyan264: Updated, please review.

lib/mfg.c
lsgunth: That's ridiculous. If there are checks to be done they should be done in the software that is doing the programming or sending the MRPC command. The switchtec  library can easily do any checks the chiplink software can -- and probably should. This run around just makes it worse for users... 

What if some user just uses the library to set these values? Sounds like a recipe for them to brick their device because apparently there are some undocumented and unchecked for checks that need some other software to be applied. This plan gets a big NAK from me.

Creating some insane binary format to force the user to use some other piece of software so that software can do some special checks is just a huge pain for everyone involved and far more dangerous.

lib/crc32.c
```diff
@@ -0,0 +1,100 @@
+/*
```
lsgunth: Similar to the header, I don't see why we can't just put this in a common crc file.

lib/crc8.c
```diff
@@ -24,7 +24,7 @@
 
 #ifdef __linux__
 
-#include "crc8.h"
```
lsgunth: Better to do a file rename in a different commit.

lib/mfg.c
lsgunth: Or better yet the firmware should do these unsepecified checks when it receives the MRPC command....

xyan264: Talked to the team again, our concerns and suggestions are noted. However, the decision to use binary format and skip command line options stays. It is a decision based on the overall picture, from design to implementation and customer support. It does mean we have to give up some usability of switchtec-user, but it's a sacrifice they are will to make. 

So let's move on with our current implementation.

lib/crc8.c
xyan264: Updated, please review.

xyan264:
Apologies for the Travis CI build error. CRC functions are defined for Linux build only. 

Should I open another PR or revert this merge?

lsgunth:
Probably revert... why isn't the repo setup to not allow merges until CI passes?

xyan264:
Maybe it's a setting issue. 'Revert' opens a new PR#153 above.

PR #148, opened 2019-12-24
Add 'mailbox' command to 'mfg' submenu.

xyan264:


lib/mfg.c
```diff
@@ -145,6 +147,42 @@ int switchtec_security_config_get(struct switchtec_dev *dev,
...
+			return ret;
+
+		reply.num_remaining = le32toh(reply.num_remaining);
+		reply.num_returned = le32toh(reply.num_returned);
+
+		ret = write(fd, reply.data,
+			    (reply.num_returned) * SWITCHTEC_MB_LOG_LEN);
+		if (ret < 0)
+			return ret;
+	} while(reply.num_remaining > 0);
```
kelvin-cao: A whitespace needs to be between while and (.

lib/mfg.c
```diff
@@ -145,6 +147,42 @@ int switchtec_security_config_get(struct switchtec_dev *dev,
 	return 0;
 }
 
+/**
+ * @brief Retrieve mailbox entries
+ * @param[in]  dev	Switchtec device handle
+ * @param[in]  fd	File handle to write the log data
+ * @return 0 on success, error code on failure
+ */
+int switchtec_mailbox_get(struct switchtec_dev *dev, int fd)
```
kelvin-cao: It might be better to name it like switchtec_mailbox_to_fd() or similar to indicate that this API eventually write the mailbox log into a file.

xyan264:
Updated and merged.

cli/mfg.c
```diff
@@ -229,6 +229,41 @@ static int info(int argc, char **argv)
 	return 0;
 }
 
+static int mailbox(int argc, char **argv)
+{
+	const char *desc = "Retrieve mailbox logs";
```
lsgunth: What are mailbox logs? How are they different from the regular logs?

PR #149, opened 2019-12-24
Add 'state-set' command to 'mfg' submenu

xyan264:


cli/mfg.c
```diff
@@ -44,6 +44,14 @@
 #include <stdio.h>
 #include <string.h>
 
+static const struct argconfig_choice secure_state_choices[] = {
+	{"INITIALIZED_UNSECURED", SWITCHTEC_INITIALIZED_UNSECURED,
+	 	"Unsecured"},
```
kelvin-cao: Indent issue and leading white space. There're other indent issues in this patch.

cli/mfg.c
```diff
@@ -44,6 +44,14 @@
...
+	{"INITIALIZED_SECURED", SWITCHTEC_INITIALIZED_SECURED,
```
lsgunth: These are confusing names... Why is everything INITIALIZED?

cli/mfg.c
```diff
@@ -404,12 +412,87 @@ static int image_select(int argc, char **argv)
...
 
+static int secure_state_set(int argc, char **argv)
+{
+	int ret;
+	enum switchtec_boot_phase phase_id;
+	struct switchtec_security_cfg_stat state = {};
+
+	const char *desc = "Set device secure state (BL1 and Main Firmware only)\n\n"
+			   "This command is can only be used when device "
+			   "secure state is UNINITIALIZED_UNSECURED.";
```
lsgunth: Definitely need more here. What does this actually do? Why would anyone care about this flag?

I'm also worried about a trend that all these new commands are just thin wrappers around basic MRPC commands. You need to consider what your users want to do or get and the use cases, not just provide thin wrappers for everything the hardware does. 

Without understanding what these commands are doing it's hard for me to review anything, but the sense I'm getting is that there should be a higher level interface. Certainly you shouldn't be setting a secure state without actually initializing some kind of security feature... Changing a flag on the device is pretty useless.

cli/mfg.c
kelvin-cao: @lsgunth The chip now has three state regarding security, UNINITIALIZED_UNSECURED, INITIALIZED_UNSECURED AND INITIALIZED_SECURED. With default shipping F/W, the chip is in the UNINITIALIZED_UNSECURED state, in which it can be changed into INITIALIZED_UNSECURED or INITIALIZED_SECURED state per user's need of the security features.

Once a chip is changed into a initialzed state, INITIALIZED_UNSECURED or INITIALIZED_SECURED, its state cannot be updated any more.

cli/mfg.c
```diff
@@ -404,12 +412,87 @@ static int image_select(int argc, char **argv)
...
+	argconfig_parse(argc, argv, desc, opts, &cfg, sizeof(cfg));
+
+	if (cfg.state == SWITCHTEC_SECURE_STATE_UNKNOWN) {
+		fprintf(stderr,
+			"Secure state must be set in this command!\n");
+		return -1;
+	}
+
+	ret = switchtec_get_device_info(cfg.dev, &phase_id, NULL, NULL);
+	if (ret < 0) {
```
kelvin-cao: We already have the phase cached to dev in open. Probably need a rebase to have those.

cli/mfg.c
```diff
@@ -404,12 +412,87 @@ static int image_select(int argc, char **argv)
...
+		return ret;
+	}
+	if (phase_id == SWITCHTEC_BOOT_PHASE_BL2) {
+		fprintf(stderr,
+			"This command is only available in BL1 or Main Firmware!\n");
+		return -2;
+	}
+
+	ret = switchtec_security_config_get(cfg.dev, &state);
+	if (ret < 0) {
```
kelvin-cao: Again, for any API which is built on top of the MRPC commands, we should not just check the negative return value of it, but also the positive case. And within the API itself, we also need to check both negative and positive return of the switchtec_cmd() for error.  #132 
```
        ret = switchtec_cmd(dev, MRPC_SECURITY_CONFIG_GET, NULL, 0,
                            &reply, sizeof(reply));
        if (ret < 0)
                return ret;
```

cli/mfg.c
lsgunth: Hmm, those don't sound like very intuitive names. This all needs to be documented better.

cli/mfg.c
xyan264: Added more detailed descriptions, please check if sufficient. @lsgunth

cli/mfg.c
xyan264: Will update once for all, after all commands checked in.

cli/mfg.c
xyan264: Some background information (excerpts from command descriptions):

A device can be in one of these three secure states:
UNINITIALIZED_UNSECURED: the default state when chip is shipped. All
security-related settings are **uninitialized**.
INITIALIZED_UNSECURED: this is the state when security-related settings are
'initialized', and the chip is set to **unsecured** state.
INITIALIZED_SECURED: this is the state when security-related settings are
'initialized', and the chip is set to **secured** state.

When device is shipped, it comes in _uninitialized-unsecured_ mode. From this mode, user can use 'config-set' and/or other mfg commands to initialize security settings. When user is satisfied with the settings, they use this command to switch chip to either INITIALIZED_SECURED or INITIALIZED_UNSECURED state, and the chip stays in that state till end of life. 

This command is the 'ENABLE' button for security-related settings.

We provide this (and a few other) commands as thin wrappers of basic MRPC commands to offer user better flexibility. As a usecase, user can program the configurations during manufacturing time, and before deployment, decide to enable security only on half of the units, leaving the other half in unsecured mode.

cli/mfg.c
```diff
@@ -404,12 +412,108 @@ static int image_select(int argc, char **argv)
...
+		switchtec_perror("mfg state-set");
+		return ret;
+	}
+	if (state.secure_state != SWITCHTEC_UNINITIALIZED_UNSECURED) {
+		fprintf(stderr,
+			"This command is only valid when secure state is UNINITIALIZED_UNSECURED!\n");
+		return -3;
+	}
+
+	if (!cfg.assume_yes)
```
lsgunth: Tabbing? Looks like it also needs braces?

xyan264:
Updated and merged.

PR #150, opened 2019-12-24
Add 'boot-resume' to 'mfg' submenu

xyan264:


cli/mfg.c
```diff
@@ -404,12 +404,50 @@ static int image_select(int argc, char **argv)
 	return ret;
 }
 
+static int boot_resume(int argc, char **argv)
+{
+	const char *desc = "Resume device boot process (BL1 and BL2 only)";
```
lsgunth: Please define "Resume" as this is not at all clear.

cli/mfg.c
```diff
@@ -404,12 +404,50 @@ static int image_select(int argc, char **argv)
...
+}
+
 static const struct cmd commands[] = {
 	{"ping", ping, "Ping firmware and get current boot phase"},
 	{"info", info, "Display security settings"},
 	{"mailbox", mailbox, "Retrieve mailbox logs"},
 	{"image_list", image_list, "Display active image list (BL1 only)"},
 	{"image_select", image_select, "Select active image index (BL1 only)"},
+	{"boot_resume", boot_resume,
+		"Resume device boot process (BL1 and BL2 only)"},
```
lsgunth: Also a nit I didn't notice before: these should all use the CMD macro given the command line name and function name are the same.

cli/mfg.c
xyan264: More descriptions added. Please take a look.

cli/mfg.c
xyan264: OK. For now I plan to keep this as-is and change all of them in one PR later, when all commands are checked in. Makes it easy for review as well.

cli/mfg.c
```diff
@@ -404,12 +404,56 @@ static int image_select(int argc, char **argv)
...
 
+static int boot_resume(int argc, char **argv)
+{
+	const char *desc = "Resume device boot process (BL1 and BL2 only)\n\n"
+			   "A normal device boot process includes BL1, "
+			   "BL2 and Main Firmware boot phases. In the case "
+			   "when boot process is paused at BL1 or BL2 phase "
+			   "(due to boot failure or BOOT_RECOVERY PIN[0:1] "
+			   "being set to LOW), sending this command requests "
+			   "device to try resuming normal boot process.";
```
lsgunth: So after this command the device will be back into normal firmware? That's probably the important thing to mention.

How does this interact with the OS? Does it not need to reboot? Keep in mind if your system does hotplug well, you might not need to; but other systems will. I would expect we should be handling it in much the same way we handle hard_reset().

xyan264: Good point. Note added to description, also added a warning with confirmation, same as hard_reset() command.

lib/mfg.c
```diff
@@ -264,3 +264,18 @@ int switchtec_active_image_index_set(struct switchtec_dev *dev,
 			    sizeof(set), NULL, 0);
 	return ret;
 }
+
+/**
+ * @brief Resume device boot.
+ *        Note that after calling this function, the current
+ *        'dev' pointer is no longer valid. Before making further
+ *        calls to switchtec library functions, be sure to close
+ *        this pointer and get a new one by calling switchtec_open().
```
lsgunth: Can we also rework this comment to take into account the device might not exist after calling it?

xyan264: Good idea. Updated.

PR #151, opened 2019-12-30
Add 'fw-transfer' and 'fw-execute' commands to mfg submenu

xyan264:
Add 'fw-transfer' and 'fw-execute' commands to mfg submenu.


cli/mfg.c
```diff
@@ -404,12 +412,149 @@ static int image_select(int argc, char **argv)
 	return ret;
 }
 
+static int fw_transfer(int argc, char **argv)
+{
+	const char *desc = "Transfer a firmware image to device (BL1 phase only)\n\n"
+			   "This command only supports transferring firmware "
+			   "image when device is in BL1 boot phase. To update "
```
lsgunth: Please mention that this should be followed by a fw-execute command to activate the firmware and that the transferred firmware is stored in RAM and not preserved after a reboot.

lib/mfg.c
```diff
@@ -264,3 +264,27 @@ int switchtec_active_image_index_set(struct switchtec_dev *dev,
...
+		uint8_t subcmd;
+		uint8_t recovery_mode;
+		uint8_t rsvd[2];
+	} cmd;
+
+	memset(&cmd, 0, sizeof(cmd));
+	cmd.subcmd = MRPC_FW_TX_EXEC;
+	cmd.recovery_mode = recovery_mode;
+
+	ret = switchtec_cmd(dev, MRPC_FW_TX, &cmd, sizeof(cmd), NULL, 0);
```
lsgunth: No need for the temporary ret variable... just 'return switchtec_cmd()'

cli/mfg.c
```diff
@@ -404,12 +412,149 @@ static int image_select(int argc, char **argv)
...
+
+	progress_finish();
+	printf("\n");
+
+	return 0;
+}
+
+static int fw_execute(int argc, char **argv)
+{
+	const char *desc = "Execute previously transferred firmware image (BL1 phase only)\n\n"
```
lsgunth: Please reference the fw_transfer command.

cli/mfg.c
```diff
@@ -404,12 +412,149 @@ static int image_select(int argc, char **argv)
...
+	}
+
+	if (!cfg.assume_yes)
+		printf("This command will execute the previously transferred image.\n");
+	ret = ask_if_sure(cfg.assume_yes);
+	if (ret) {
+		return ret;
+	}
+
+	ret = switchtec_fw_exec(cfg.dev, cfg.bl2_rec_mode);
```
lsgunth: What happens if no image has been transferred yet? Presumably an error and I'd want that error message to be very clear but I don't see how that will happen here.

xyan264:
Updated based on comments above, @lsgunth please review.

inc/switchtec/mfg.h
```diff
@@ -102,6 +102,13 @@ struct switchtec_active_index {
 	enum switchtec_active_index_id keyman;
 };
 
+enum switchtec_bl2_recovery_mode {
+	SWITCHTEC_BL2_RECOVERY_I2C = 1,
+	SWITCHTEC_BL2_RECOVERY_XMODEM = 2,
+	SWITCHTEC_BL2_RECOVERY_I2C_AND_XMODEM = 3,
+	SWITCHTEC_BL2_RECOVERY_NOT_SET = 0xff
```
lsgunth: I don't think SWITCHTEC_BL2_RECOVERY_NOT_SET makes any sense here... This is for the library interface and you seem to be abusing it for the CLI...

cli/mfg.c
```diff
@@ -404,12 +412,153 @@ static int image_select(int argc, char **argv)
...
+		.bl2_rec_mode = SWITCHTEC_BL2_RECOVERY_NOT_SET
+	};
+	const struct argconfig_options opts[] = {
+		DEVICE_OPTION_NO_PAX,
+		{"yes", 'y', "", CFG_NONE, &cfg.assume_yes, no_argument,
+			"assume yes when prompted"},
+		{"bl2_recovery_mode", 'm', "MODE",
+			CFG_CHOICES, &cfg.bl2_rec_mode,
+			required_argument, "BL2 recovery mode",
+			.choices = recovery_mode_choices},
```
lsgunth: Definitely need some documentation... I have no idea what this does, but it seems like having it just throwing an error by default is not very user friendly. Certainly this could have a sane default? Maybe just "BOTH"?

xyan264: Make sense. Added descriptions and default value.

inc/switchtec/mfg.h
xyan264: Good idea. Removed definition and updated affected code.

PR #152, opened 2020-01-09
Move mfg commands to linux build only.

xyan264:


lsgunth:
Why?

xyan264:
CRC functions are defined for Linux build only. 

Also we plan to check in command code that calls OpenSSL library, which will cause Travis MinGW build to fail.

lsgunth:
Why are CRCs linux only? That seems odd... lets fix that? 

Not sure why that was done... If they don't build on windows then fix that.... Otherwise just drop the linux only tag from that file.

xyan264:
Forgot to add these in description:
--------
Many of the mfg commands run on top of I2C device driver which is currently only available in Linux. So it is decided that mfg commands are only available for Linux build. 

Also there are commands (to be checked in) that uses OpenSSL library, which can cause Travis MinGW build errors. It is much easier to make mfg commands Linux-build only, since they are intended for use under Linux anyways.

xyan264:
CRC functions have been made available for Linux and Windows build in PR#154.

lsgunth:
I don't think restricting it to linux only is the way to go. It just confuses things if someone adds I2C support to windows. If something requires I2C it should fail if it's not I2C or something like that... Add the fail code that actually describes the condition you want to prevent. Don't reject non-Linux when you really want to reject Non-I2C... (Or what about the UART?).

xyan264:
Yes we can keep the MinGW build, but the OpenSSL library will be a real hassle. 

There is no pre-built OpenSSL binaries for MinGW, so for Travis build, the only option I see is to cross-compile OpenSSL first, then link our code against it. It could take quite some time to find a solution for MinGW build. Do you have other suggestions?

lsgunth:
The configure script should check for the openssl library and disable any features that require it. (Preferably in a user visible way: so that when the user tries to do something that required it it says that switchtec was not built with the openssl library).

You should not disable all the mfg menu just because you don't have openssl. You should also not assume the user had openssl available when they built the program. Just make it optional.

xyan264:
Good idea. I will try that.

PR #155, opened 2020-01-10
Add 'kmsk-entry-add' command to mfg submenu.

xyan264:
User provide public key file (.PEM format), KMSK file (generated from Chiplink) and Signature file (generated using third-party tool) to add a KMSK entry to chip OTP memory. The public key file and Signature file are used for authentication purpose, and the KMSK file contains the data to be written to OTP.

The command will print error message with directions if OpenSSL development library is not installed. This avoids Travis CI compile errors and is more user-friendly than link errors.


lib/mfg.c
```diff
@@ -463,3 +485,140 @@ int switchtec_read_sec_cfg_file(FILE *setting_file,
...
+	} cmd2;
+
+	struct kmsk_cmd3 {
+		uint8_t subcmd;
+		uint8_t num_entries;
+		uint8_t reserved[2];
+		uint8_t kmsk[SWITCHTEC_KMSK_LEN];
+	} cmd3;
+
+	if (public_key) {
```
lsgunth: It kind of seems like this would be better as three functions: set_public_key, set signature and set_kmsmk... They don't really seem related and pushing them all into the same function just makes this more complicated.... At the very list split them into three static functions.

lib/mfg.c
```diff
@@ -463,3 +485,140 @@ int switchtec_read_sec_cfg_file(FILE *setting_file,
...
+#if HAVE_LIBCRYPTO
+/**
+ * @brief Read public key from public key file
+ * @param[in]  pubk_file Public key file
+ * @param[out] pubk	 Public key
+ * @param[out] exp	 Public key exponent
+ * @return 0 on success, error code on failure
+ */
+int switchtec_read_pubk_file(FILE *pubk_file, uint8_t *pubk,
+			     uint32_t *exp)
```
lsgunth: Nit: the indent doesn't look right here

lib/mfg.c
```diff
@@ -463,3 +485,140 @@ int switchtec_read_sec_cfg_file(FILE *setting_file,
...
+
+#if HAVE_LIBCRYPTO
+/**
+ * @brief Read public key from public key file
+ * @param[in]  pubk_file Public key file
+ * @param[out] pubk	 Public key
+ * @param[out] exp	 Public key exponent
+ * @return 0 on success, error code on failure
+ */
+int switchtec_read_pubk_file(FILE *pubk_file, uint8_t *pubk,
```
lsgunth: Maybe the better API would be to read the key from the file and send it to the hardware? Are there any use cases that can get the key from somewhere else?

Having an API that reads a random binary blob and expecting another function to get that blob when their just non-descriptive uint8_t is pretty ugly. If you need two functions I'd make it a specific pre-sized structure. Or a dynamically allocated private structure. Or, have the set_public_key function take RSAKey directly? 

Oh, except how is the user supposed to know how large the pubk and exp need to be? What happens if they make them too small? Not a great API choice.

lib/mfg.c
```diff
@@ -463,3 +485,140 @@ int switchtec_read_sec_cfg_file(FILE *setting_file,
...
+}
+#endif
+
+/**
+ * @brief Read KMSK data from KMSK file
+ * @param[in]  kmsk_file KMSK file
+ * @param[out] kmsk   	 KMSK entry data
+ * @return 0 on success, error code on failure
+ */
+int switchtec_read_kmsk_file(FILE *kmsk_file, uint8_t *kmsk)
```
lsgunth: Same problems as above... how big is the kmsk data? What is the user supposed to pass here?

cli/mfg.c
```diff
@@ -786,6 +788,191 @@ static int security_config_set(int argc, char **argv)
 	return 0;
 }
 
+#if HAVE_LIBCRYPTO
+static int load_sig_from_file(FILE *sig_fimg, uint8_t *sig)
```
lsgunth: Same problems as with the other load functions. Also, why is this one special in that it's only in the CLI and the others get to be in the library?

cli/mfg.c
```diff
@@ -786,6 +788,191 @@ static int security_config_set(int argc, char **argv)
...
+static int kmsk_already_exist(uint8_t *kmsk,
+			      struct switchtec_security_cfg_stat *state)
+{
+	int i;
+	int key_idx;
+	int match;
+
+	for(key_idx = 0; key_idx < state->public_key_num; key_idx++) {
+		match = 1;
+		for(i = 0; i < SWITCHTEC_KMSK_LEN; i++) {
```
lsgunth: looks like memcmp....

cli/mfg.c
```diff
@@ -786,6 +788,191 @@ static int security_config_set(int argc, char **argv)
...
+			return 1;
+	}
+
+	return 0;
+}
+#endif 
+
+static int kmsk_add(int argc, char **argv)
+{
+#if HAVE_LIBCRYPTO
```
lsgunth: Better to put the #if outside the function and replicate the prototype. We don't like pre-processor directives mixed in code.

cli/mfg.c
```diff
@@ -786,6 +788,191 @@ static int security_config_set(int argc, char **argv)
...
+
+	rlen = fread(sig, 1, SWITCHTEC_SIG_LEN, sig_fimg);
+
+	if (rlen < SWITCHTEC_SIG_LEN)
+		return -1;
+
+	return 0;
+}
+
+static int kmsk_already_exist(uint8_t *kmsk,
```
lsgunth: Seems like this could be in the library. switchtec_security_state_has_key()... I'd put the state as the first argument.

cli/mfg.c
```diff
@@ -786,6 +788,191 @@ static int security_config_set(int argc, char **argv)
...
+			      struct switchtec_security_cfg_stat *state)
```
lsgunth: This might be too late.... but 'stat' is a horrible abbreviation for 'state' because it can be confused with "statistics".... maybe you can clean that up at some point.

lib/mfg.c
xyan264: This is a function which you can use in two scenarios: 
- when the chip is in UNSECURED mode, only KMSK entry data is need.
- when the chip is in SECURED mode, all parameters are needed.

The raw uint8_t* parameter type is not a good choice. I will change them to structure with defined length.

cli/mfg.c
xyan264: This function is only used for this command so is kept here for now. A little hesitant to add it to API, which means one more (probably unnecessary) function to support.

cli/mfg.c
xyan264: OK will update later.

inc/switchtec/mfg.h
```diff
@@ -136,9 +139,17 @@ int switchtec_active_image_index_set(struct switchtec_dev *dev,
 int switchtec_fw_exec(struct switchtec_dev *dev,
 		      enum switchtec_bl2_recovery_mode recovery_mode);
 int switchtec_boot_resume(struct switchtec_dev *dev);
+int switchtec_kmsk_set(struct switchtec_dev *dev,
```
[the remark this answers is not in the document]
xyan264: KMSK stands for key-manifest security key, so we will keep the KMSK name for now.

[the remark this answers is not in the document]
lsgunth: What does that mean? Is it like a root key? That would be a more standard name...
In any case... add some comments and mention the abbrieviation in the commit message. Also, any user facing description should use the full name.

lib/mfg.c
lsgunth: All the more reason to separate it into multiple API functions.... Unless you actually want to check for SECURED mode and ensure that all parameters are set. But that's still a bit ugly IMO.

cli/mfg.c
lsgunth: All the functions are only used for that command.... Looks like a useful function to me to support into the future.... Better than having users open code it themselves because then you have to support the data structure for ever into the future which is much worse.

xyan264: Make sense. I will add it to API function.

lib/mfg.c
xyan264: Separating them into 3 functions means user could call them in different sequence, which is not readily supported by firmware. We would like to make sure the data are sent in the sequence as is done in the function. So keeping it in one function is a better option at the moment.

lsgunth: Well I'd still separate them into three static functions then. All in one is a bit much.

inc/switchtec/mfg.h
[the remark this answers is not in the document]
xyan264: It is like a root key that is used verify a Key-Manifest partition, which stores the public keys for all other partitions. The use of 'KMSK' abbreviation is consistent with DevSpec documentation and all other documentations. 

I can add full name and descriptions to function header and command help text to give more context for the command and API function.

xyan264:
Updated, please review.

xyan264:
@lsgunth could you comment on this? This is more urgent...

configure.ac
```diff
@@ -19,6 +25,13 @@ AS_IF([test "x$with_curses" != xno &&
 
 AC_CHECK_HEADERS([curses.h ncurses/curses.h])
 
+AS_IF([test "x$with_ssl" != xno],
+       [AC_CHECK_HEADERS([openssl/pem.h])
+        AC_CHECK_LIB([crypto], [RSA_free], [], [], [])])
```
lsgunth: Still looking to see that you check the openssl header here...

xyan264:
Updated, verified and merged.

PR #156, opened 2020-01-11
Add fw-type and secure version check to 'fw-update' command.

xyan264:
Add the following checks in 'fw-update':
- FW secure version check: remind user that updating the image will update the chip secure version as well.
- FW type check: if partition map is updated, remind user to update all other partitions as well.
- Device boot phase check: if boot phase is BL2, remind user that the image is not automatically activated, be sure to use 'fw-toggle' to activate the image.

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -1179,12 +1179,14 @@ enum switchtec_fw_type check_and_print_fw_image(int img_fd,
 		return ret;
 	}
 
-	printf("File:     %s\n", get_basename(img_filename));
-	printf("Gen:      %s\n", switchtec_fw_image_gen_str(&info));
-	printf("Type:     %s\n", switchtec_fw_image_type(&info));
-	printf("Version:  %s\n", info.version);
-	printf("Img Len:  0x%" FMT_SIZE_T_x "\n", info.image_len);
-	printf("CRC:      0x%08lx\n", info.image_crc);
+	printf("File:           %s\n", get_basename(img_filename));
+	printf("Gen:            %s\n", switchtec_fw_image_gen_str(&info));
+	printf("Type:           %s\n", switchtec_fw_image_type(&info));
+	printf("Version:        %s\n", info.version);
+	printf("Img Len:        0x%" FMT_SIZE_T_x "\n", info.image_len);
+	printf("CRC:            0x%08lx\n", info.image_crc);
+	if (info.gen != SWITCHTEC_GEN3)
+		printf("Secure version: 0x%08lx\n", info.secure_version);
 
 	return info.type;
 }
@@ -1317,6 +1319,7 @@ static int fw_update(int argc, char **argv)
 {
 	int ret;
 	int type;
+	struct switchtec_fw_image_info info;
 	const char *desc = "Flash device with a new firmware image\n\n"
 			   "This command only supports flashing firmware "
 			   "when device is in BL2 or MAIN boot phase. To "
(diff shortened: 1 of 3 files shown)


cli/main.c
```diff
@@ -1179,13 +1180,15 @@ enum switchtec_fw_type check_and_print_fw_image(int img_fd,
...
-	printf("CRC:      0x%08lx\n", info.image_crc);
-
+	printf("File:           %s\n", get_basename(img_filename));
+	printf("Gen:            %s\n", switchtec_fw_image_gen_str(&info));
+	printf("Type:           %s\n", switchtec_fw_image_type(&info));
+	printf("Version:        %s\n", info.version);
+	printf("Img Len:        0x%" FMT_SIZE_T_x "\n", info.image_len);
+	printf("CRC:            0x%08lx\n", info.image_crc);
+	if (info.gen != SWITCHTEC_GEN3)
+		printf("Secure version: 0x%08x\n",
```
kelvin-cao: why not use 'lx' instead of the casting?

cli/main.c
```diff
@@ -1313,6 +1316,66 @@ static int fw_info(int argc, char **argv)
...
+		}
+		break;
+	default:
+		break;
+	}
+
+	if ((prompt == 0) || assume_yes_ver)
+		return 0;
+
+	fprintf(stderr, "\n\nWARNING:\n"
```
lsgunth: I'd put the check outside the static function. This function, based on it's name, just does the compare... That also means you don't have to pass it an extra variable.

xyan264:
Updated based on the above suggestions. Please review.

cli/main.c
```diff
@@ -1316,18 +1316,14 @@ static int fw_info(int argc, char **argv)
 	return 0;
 }
 
-static int compare_device_secure_version(struct switchtec_dev *dev, int img_fd,
-				         int assume_yes_ver)
+static int check_device_secure_version(struct switchtec_dev *dev, int img_fd)
```
lsgunth: This patch should be squashed... It's making significant changes to a patch in this series and I'd prefer to see the result.

cli/main.c
lsgunth: I'd still put this outside the function in the CLI's function. It's significantly more reusable to have a function that returns a property than a static function prints a specific message for its specific call site.

lsgunth: Also, the warning should really be printed regardless of whether assume_yes is set.... assume_yes is really just for not asking to continue, -- it's not "don't print warnings"...

cli/main.c
```diff
@@ -1451,11 +1447,12 @@ static int fw_update(int argc, char **argv)
 		}
 	}
 
-	ret = compare_device_secure_version(cfg.dev, fileno(cfg.fimg),
-					    cfg.assume_yes);
-	if (ret) {
-		fclose(cfg.fimg);
-		return ret;
+	if (!cfg.assume_yes && !switchtec_is_gen3(cfg.dev)) {
```
lsgunth: I'd still keep the switchtec_is_gen3() inside the static function....

And in fact it sure looks like this could be a useful library function if restructured properly.... It's much better in the long run to add simple interfaces like that do complicated things based on internal state. Otherwise people will just copy the non-library code into their project and you break their project if you ever want to change it.

```
ret = switchtec_secure_version_matches_file(cfg.dev, fileno(cfg.fimg));
if (!ret) {
        /* print warning */
        ret = ask_if_sure(cfg.assume_yes);
        if (ret) {
                fclose(cfg.fimg);
                return ret;
        }
}
```

cli/main.c
xyan264: Apologies. Had meant to do a rebase and commit --amend. Will update this later.

cli/main.c
xyan264: The warning message needs to contain _device_secure_version_ and _file_secure_version_. The function needs to return these fields. The code looks like:
`ret = switchtec_secure_version_matches_file(cfg.dev, fileno(cfg.fimg), &dev_sec_ver, &file_sec_ver);
if (!ret) {
        printf("Warning: updating this file updates device secure version from %x to %x\n", dev_sec_ver, file_sec_ver);
        ret = ask_if_sure(cfg.assume_yes);
        if (ret) {
                fclose(cfg.fimg);
                return ret;
        }
}`

lsgunth: Then two functions that return the sec_ver for a device and a file would be more appropriate...

xyan264: Function is now named "switchtec_fw_file_secure_version_newer()".

xyan264:
Code updated, @lsgunth please take a look.

PR #157, opened 2020-01-14
Add 'debug-unlock' and 'debug-lock-update' commands

xyan264:
Firmware supports debugging through EJTAG, CLI (command line console) and GAS (global address space) access. By default these features are disabled when chip is set to 'SECURED' mode. 

'debug-unlock' command unlocks access to these debug features.
'debug-lock-update' updates the secure unlock version number stored in chip OTP memory.


lib/mfg.c
```diff
@@ -411,6 +411,99 @@ int switchtec_secure_state_set(struct switchtec_dev *dev,
...
+	} cmd1 = {};
+
+	struct unlock_cmd2 {
+		uint32_t subcmd;
+		uint32_t serial;
+		uint32_t unlock_ver;
+		uint8_t signature[SWITCHTEC_SIG_LEN];
+	} cmd2 = {};
+
+	cmd1.subcmd = htole32(MRPC_DBG_UNLOCK_PKEY);
```
lsgunth: Seems like MRPC_DBG_UNLOCK_PKEY should be it's own static function so it can be reused in both of these functions.

cli/mfg.c
```diff
@@ -937,8 +1130,27 @@ static int kmsk_entry_add(int argc, char **argv)
...
+static int debug_unlock(int argc, char **argv)
+{
+	fprintf(stderr,
+		"This command is only available when OpenSSL development library is installed.\n"
+		"Try installing OpenSSL development library in your system and rebuild this\n"
+		"program by doing 'configure' and then 'make'.\n");
+	return -1;
+}
+
+static int debug_lock_update(int argc, char **argv)
```
lsgunth: Seeing we're repeating this so much, how about something like:

```
static int no_openssl(int argc, char **argv)
{
	fprintf(stderr,
		"This command is only available when OpenSSL development library is installed.\n"
		"Try installing OpenSSL development library in your system and rebuild this\n"
		"program by doing 'configure' and then 'make'.\n");
	return -1;
}
#define kmsk_entry_add     no_openssl
#define debug_unlock       no_openssl
#define debug_lock_update  no_openssl
```

xyan264:
Updated based on the above suggestions.

PR #159, opened 2020-01-15
Use new MRPC command ID for 'fw-toggle' and 'fw-update'.

xyan264:
In BL2 boot phase:
- the MRPC command and sub command ID for 'fw-toggle' command has changed
- the MRPC sub command ID for 'fw-update' has changed


xyan264:
Good point. Commit message updated to include what has changed.

PR #160, opened 2020-01-16
Fix i2c_gas_write_status_get() errno handling.

xyan264:
In the do..while loop, when ioctl fails, _errno_ will be set. Later if retry of ioctl succeeds, we should restore _errno_ to its previous value.

After this fix, the net effect is i2c_gas_write_status_get() does not change _errno_ when it succeeds.

### DIFF
--- a/lib/platform/linux-i2c.c
+++ b/lib/platform/linux-i2c.c
@@ -356,8 +356,10 @@ static uint8_t i2c_gas_write_status_get(struct switchtec_dev *dev,
 		pec = i2c_msg_pec(&msgs[1], msgs[1].len - PEC_BYTE_COUNT,
 				  msg_0_pec, false);
 		if (rx_buf[0] == tag && rx_buf[2] == pec &&
-		    (rx_buf[1] == 0 || rx_buf[1] == GAS_TWI_MRPC_ERR))
+		    (rx_buf[1] == 0 || rx_buf[1] == GAS_TWI_MRPC_ERR)) {
+			errno = 0;
 			return rx_buf[1];
+		}
 
 		/* Extra delay is typically only needed for BL1/2 phase */
 		usleep(2000);


lib/platform/linux-i2c.c
```diff
@@ -343,6 +344,8 @@ static uint8_t i2c_gas_write_status_get(struct switchtec_dev *dev,
 	msgs[1].len = 3;
 	msgs[1].buf = rx_buf;
 
+	err = errno;
```
lsgunth: No, this is excessive. All you need to do is clear errno if you ignore an error (or retry an error).  Which should be done before you retry.

xyan264: What if previously errno = X before i2c_gas_write_status_get():
- ioctl failed, we clear errno = 0, and retry
- ioctl succeeded, we return, and errno = 0

So the net effect will be: 
- errno = X
- after calling i2c_gas_write_status_get(), errno = 0
So errno is changed even when function succeeds.

kelvin-cao: I'd think with this fix we'll achieve below statement. So I agree with Paul.

> Valid error numbers are all nonzero; errno is never set to zero by any system call or library function. -- from errno manpage

lsgunth: My version of the man page is:

> The  value  in  errno  is significant only when the return value of the call indicated an error
> ... a function that succeeds is allowed to  change errno.  

> A common mistake is to do
> 
>            if (somecall() == -1) {
>                printf("somecall() failed\n");
>                if (errno == ...) { ... }
>            }
> 
> where errno no longer needs to have the value it had upon  return  from
> somecall()  (i.e.,  it may have been changed by the printf(3)).  If the
> value of errno should be preserved across a library call,  it  must  be
> saved:
> 
>            if (somecall() == -1) {
>                int errsv = errno;
>                printf("somecall() failed\n");
>                if (errsv == ...) { ... }
>            }

xyan264: Agree with @lsgunth . The man page is here:
   http://man7.org/linux/man-pages/man3/errno.3.html

The description does allow different interpretations, but the _Note_ section at the end of page made it clear. 

I have updated code to reset _errno_ when the function succeeds. 

@kelvin-cao please confirm.

lsgunth: Well actually we are all learning here as my original assessment, and the way we've handled this in the past  is not in line with the man page. It seems we should be fixing up the call sites that expect errno to be something meaningful after calling any other function....

lsgunth: errno sucks by the way. For many of these reasons, but we are stuck with it due to the standard library.

kelvin-cao: To my understanding, the man page statements does not conflict with each other, since valid errno is non-zero, so no function needs to reset it to zero. A zero errno makes no different to a successful return, and for an error return, the errno should never be zero too. So it's allowed to change the errno in a success function, but the value has no reason to be reset to zero. 

I agree that we didn't handle this in line with the man page in the past. A proper fix should like the one that Logan quoted. 

Let me know your thoughts.

xyan264:
PR closed due to a scope change. Discussion will continue in another PR.

PR #168, opened 2020-01-21
Disable UART access for MFG submenu commands

xyan264:
Rename 'DEVICE_OPTION_NO_PAX' to 'DEVICE_OPTION_MFG'. 

Also the 'mfg' submenu commands do not support UART access, so the help text in device option is changed accordingly.


cli/common.h
```diff
@@ -62,10 +62,22 @@ enum switchtec_fw_type check_and_print_fw_image(int img_fd,
 			" * an index (0, 1, 2)\n" \
 			" * a PCI address (3:00.1)\n" \
 			" * an I2C path with slave address (/dev/i2c-1@0x20)\n" \
-			" * a UART path (/dev/ttyUSB0)\n" \
```
lsgunth: No UART in MFG? What happens if the user specifies a UART?

cli/common.h
```diff
@@ -62,10 +62,22 @@ enum switchtec_fw_type check_and_print_fw_image(int img_fd,
...
+			.custom_handler=switchtec_handler, \
+			.complete="/dev/switchtec*", \
+			.env="SWITCHTEC_DEV", \
+			.help="Switchtec device to operate on. Can be any of:\n" \
+			" * a device path (/dev/switchtec0)\n" \
+			" * an index (0, 1, 2)\n" \
+			" * a PCI address (3:00.1)\n" \
+			" * an I2C path with slave address (/dev/i2c-1@0x20)\n" \
+			" * a UART path (/dev/ttyUSB0)\n" \
+	}, \
```
lsgunth: Seems to me like we just cleaned this up so it's not duplicating some much junk...

If the only thing that's different is the help message, you should still be able to use some macros to not replicate it.

cli/common.h
xyan264: The command will fail with error code if UART is specified.

cli/common.h
xyan264: Updated to clear up the duplicates.

cli/common.h
lsgunth: What causes that error code? And is the error message actually informative? I don't remember seeing any code to that effect...

cli/common.h
```diff
@@ -62,10 +62,12 @@ enum switchtec_fw_type check_and_print_fw_image(int img_fd,
 			" * an index (0, 1, 2)\n" \
 			" * a PCI address (3:00.1)\n" \
 			" * an I2C path with slave address (/dev/i2c-1@0x20)\n" \
-			" * a UART path (/dev/ttyUSB0)\n" \
+			UART \
 	}
```
lsgunth: I'd probably label the argument EXTRA_HELP, or something to that effect. It could be used for other things.

cli/common.h
xyan264: Since BL1/2 does not support GAS register access, the command will fail during switchtec_open(), with error message: 'Error communicating with the device. Please check your setup.'

In MAIN, the command runs with no problem. However, we don't recommend user using UART for mfg commands.

cli/common.h
xyan264: Updated.

cli/common.h
lsgunth: Well that's hardly informative. It sounds like they just have a bad connection.

It should error out with the message: "Error: manufacturing commands are not supported over UART connections". (or something like that).

xyan264: Good idea. Updated. 

The way to determine if a path is _tty_ might be clumsy. Suggestions welcome!

cli/common.h
xyan264: @lsgunth could you review and approve if it looks OK...

cli/main.c
```diff
@@ -90,6 +90,23 @@ int switchtec_handler(const char *optarg, void *value_addr,
 	return 0;
 }
 
+int mfg_handler(const char *optarg, void *value_addr,
+		const struct argconfig_options *opt)
+{
+	int fd;
+
+	fd = open(optarg, O_RDWR);
+	if (fd >= 0 && isatty(fd)) {
```
lsgunth: Yeah, no I really don't like this... For starters it's a terrible way to tell if you are working with a UART, secondly, it does nothing for library users.

Here's a rough untested diff that would be better. (Preferably done in a couple commits):

```Diff

diff --git a/inc/switchtec/errors.h b/inc/switchtec/errors.h
index c020bfcaf497..d37361a96331 100755
--- a/inc/switchtec/errors.h
+++ b/inc/switchtec/errors.h
@@ -59,6 +59,8 @@ enum {
 	ERR_EVENT_INVALID = 0x100003,
 	ERR_RST_RULE_FAILED = 0x100005,
 	ERR_ACCESS_REFUSED = 0xFFFF0001,
+
+	ERR_UART_NOT_SUPPORTED = 0x10000001,
 };
 
 #endif
diff --git a/lib/mfg.c b/lib/mfg.c
index 8b0420377059..90df281745bc 100755
--- a/lib/mfg.c
+++ b/lib/mfg.c
@@ -89,6 +89,19 @@ static void RSA_get0_key(const RSA *r, const BIGNUM **n,
 }
 #endif
 
+static int switchtec_mfg_cmd(struct switchtec_dev *dev, uint32_t cmd,
+			     const void *payload, size_t payload_len,
+			     void *resp, size_t resp_len);
+{
+	if (dev->ops->flags & SWITCHTEC_OPS_FLAG_NO_MFG) {
+		errno = ERR_UART_NOT_SUPPORTED | SWITCHTEC_ERRNO_MRPC_FLAG_BIT;
+		return -1;
+	}
+
+	return switchtec_cmd(dev, cmd, payload, payload_len,
+			     resp, resp_len);
+}
+
 /**
  * @brief Get serial number and security version
  * @param[in]  dev	Switchtec device handle
@@ -100,8 +113,8 @@ int switchtec_sn_ver_get(struct switchtec_dev *dev,
 {
 	int ret;
 
-	ret = switchtec_cmd(dev, MRPC_SN_VER_GET, NULL, 0, info,
-			    sizeof(struct switchtec_sn_ver_info));
+	ret = switchtec_mfg_cmd(dev, MRPC_SN_VER_GET, NULL, 0, info,
+				sizeof(struct switchtec_sn_ver_info));
 	if (ret)
 		return ret;
 
@@ -137,8 +150,8 @@ int switchtec_security_config_get(struct switchtec_dev *dev,
 		uint8_t  rsvd4[32];
 	} reply;
 
-	ret = switchtec_cmd(dev, MRPC_SECURITY_CONFIG_GET, NULL, 0,
-			    &reply, sizeof(reply));
+	ret = switchtec_mfg_cmd(dev, MRPC_SECURITY_CONFIG_GET, NULL, 0,
+				&reply, sizeof(reply));
 	if (ret)
 		return ret;
 
@@ -200,8 +213,8 @@ int switchtec_mailbox_to_file(struct switchtec_dev *dev, int fd)
 	} reply;
 
 	do {
-		ret = switchtec_cmd(dev, MRPC_MAILBOX_GET, &num_to_read,
-				    sizeof(int), &reply,  sizeof(reply));
+		ret = switchtec_mfg_cmd(dev, MRPC_MAILBOX_GET, &num_to_read,
+					sizeof(int), &reply,  sizeof(reply));
 		if (ret)
 			return ret;
 
@@ -263,8 +276,8 @@ int switchtec_security_config_set(struct switchtec_dev *dev,
 
 	sd.pub_key_exponent = htole32(setting->public_key_exponent);
 
-	ret = switchtec_cmd(dev, MRPC_SECURITY_CONFIG_SET, &sd, sizeof(sd),
-			    NULL, 0);
+	ret = switchtec_mfg_cmd(dev, MRPC_SECURITY_CONFIG_SET, &sd, sizeof(sd),
+				NULL, 0);
 	return ret;
 }
 
@@ -282,8 +295,8 @@ int switchtec_active_image_index_get(struct switchtec_dev *dev,
 		uint8_t index[SWITCHTEC_ACTV_IDX_MAX_ENTRIES];
 	} reply;
 
-	ret = switchtec_cmd(dev, MRPC_ACT_IMG_IDX_GET, NULL,
-			    0, &reply, sizeof(reply));
+	ret = switchtec_mfg_cmd(dev, MRPC_ACT_IMG_IDX_GET, NULL,
+				0, &reply, sizeof(reply));
 	if (ret)
 		return ret;
 
@@ -345,8 +358,8 @@ int switchtec_active_image_index_set(struct switchtec_dev *dev,
 
 	set.count = htole32(i);
 
-	ret = switchtec_cmd(dev, MRPC_ACT_IMG_IDX_SET, &set,
-			    sizeof(set), NULL, 0);
+	ret = switchtec_mfg_cmd(dev, MRPC_ACT_IMG_IDX_SET, &set,
+				sizeof(set), NULL, 0);
 	return ret;
 }
 
@@ -369,7 +382,7 @@ int switchtec_fw_exec(struct switchtec_dev *dev,
 	cmd.subcmd = MRPC_FW_TX_EXEC;
 	cmd.recovery_mode = recovery_mode;
 
-	return switchtec_cmd(dev, MRPC_FW_TX, &cmd, sizeof(cmd), NULL, 0);
+	return switchtec_mfg_cmd(dev, MRPC_FW_TX, &cmd, sizeof(cmd), NULL, 0);
 }
 
 /**
@@ -379,15 +392,15 @@ int switchtec_fw_exec(struct switchtec_dev *dev,
  *        calls to switchtec library functions, be sure to close
  *        this pointer and get a new one by calling switchtec_open().
  * 	  Also be sure to check the return value of switchtec_open()
- * 	  for error, as the device might not be immediately 
+ * 	  for error, as the device might not be immediately
  * 	  accessible after normal boot process.
  * @param[in]  dev	Switchtec device handle
  * @return 0 on success, error code on failure
  */
 int switchtec_boot_resume(struct switchtec_dev *dev)
 {
-	return switchtec_cmd(dev, MRPC_BOOTUP_RESUME, NULL, 0,
-			     NULL, 0);
+	return switchtec_mfg_cmd(dev, MRPC_BOOTUP_RESUME, NULL, 0,
+				 NULL, 0);
 }
 
 /**
@@ -407,8 +420,8 @@ int switchtec_secure_state_set(struct switchtec_dev *dev,
 	}
 	data = htole32(state);
 
-	return switchtec_cmd(dev, MRPC_SECURE_STATE_SET, &data, sizeof(data),
-			     NULL, 0);
+	return switchtec_mfg_cmd(dev, MRPC_SECURE_STATE_SET, &data,
+				 sizeof(data), NULL, 0);
 }
 
 static int dbg_unlock_send_pubkey(struct switchtec_dev *dev,
@@ -425,8 +438,8 @@ static int dbg_unlock_send_pubkey(struct switchtec_dev *dev,
 	memcpy(cmd.pub_key, public_key->pubkey, SWITCHTEC_PUB_KEY_LEN);
 	cmd.pub_key_exp = htole32(public_key->pubkey_exp);
 
-	return switchtec_cmd(dev, MRPC_DBG_UNLOCK, &cmd,
-			     sizeof(cmd), NULL, 0);
+	return switchtec_mfg_cmd(dev, MRPC_DBG_UNLOCK, &cmd,
+				 sizeof(cmd), NULL, 0);
 }
 
 /**
@@ -461,8 +474,8 @@ int switchtec_dbg_unlock(struct switchtec_dev *dev, uint32_t serial,
 	cmd.unlock_ver = htole32(ver_sec_unlock);
 	memcpy(cmd.signature, signature->signature, SWITCHTEC_SIG_LEN);
 
-	return switchtec_cmd(dev, MRPC_DBG_UNLOCK, &cmd,
-			     sizeof(cmd), NULL, 0);
+	return switchtec_mfg_cmd(dev, MRPC_DBG_UNLOCK, &cmd,
+				 sizeof(cmd), NULL, 0);
 }
 
 /**
@@ -498,8 +511,8 @@ int switchtec_dbg_unlock_version_update(struct switchtec_dev *dev,
 	cmd.unlock_ver = htole32(ver_sec_unlock);
 	memcpy(cmd.signature, signature->signature, SWITCHTEC_SIG_LEN);
 
-	return switchtec_cmd(dev, MRPC_DBG_UNLOCK, &cmd, sizeof(cmd),
-			     NULL, 0);
+	return switchtec_mfg_cmd(dev, MRPC_DBG_UNLOCK, &cmd, sizeof(cmd),
+				 NULL, 0);
 }
 
 /**
@@ -592,8 +605,8 @@ static int kmsk_set_send_pubkey(struct switchtec_dev *dev,
 	       SWITCHTEC_PUB_KEY_LEN);
 	cmd.pub_key_exponent = htole32(public_key->pubkey_exp);
 
-	return switchtec_cmd(dev, MRPC_KMSK_ENTRY_SET, &cmd,
-			     sizeof(cmd), NULL, 0);
+	return switchtec_mfg_cmd(dev, MRPC_KMSK_ENTRY_SET, &cmd,
+				 sizeof(cmd), NULL, 0);
 }
 
 static int kmsk_set_send_signature(struct switchtec_dev *dev,
@@ -609,8 +622,8 @@ static int kmsk_set_send_signature(struct switchtec_dev *dev,
 	memcpy(cmd.signature, signature->signature,
 	       SWITCHTEC_SIG_LEN);
 
-	return switchtec_cmd(dev, MRPC_KMSK_ENTRY_SET, &cmd,
-		    	     sizeof(cmd), NULL, 0);
+	return switchtec_mfg_cmd(dev, MRPC_KMSK_ENTRY_SET, &cmd,
+				 sizeof(cmd), NULL, 0);
 }
 
 static int kmsk_set_send_kmsk(struct switchtec_dev *dev,
@@ -627,8 +640,8 @@ static int kmsk_set_send_kmsk(struct switchtec_dev *dev,
 	cmd.num_entries = 1;
 	memcpy(cmd.kmsk, kmsk->kmsk, SWITCHTEC_KMSK_LEN);
 
-	return switchtec_cmd(dev, MRPC_KMSK_ENTRY_SET, &cmd, sizeof(cmd),
-			     NULL, 0);
+	return switchtec_mfg_cmd(dev, MRPC_KMSK_ENTRY_SET, &cmd, sizeof(cmd),
+				 NULL, 0);
 }
 
 /**
diff --git a/lib/platform/linux-uart.c b/lib/platform/linux-uart.c
index 81eb2999deda..695db2c69277 100755
--- a/lib/platform/linux-uart.c
+++ b/lib/platform/linux-uart.c
@@ -431,6 +431,7 @@ static ssize_t uart_write_from_gas(struct switchtec_dev *dev, int fd,
 }
 
 static const struct switchtec_ops uart_ops = {
+	.flags = SWITCHTEC_OPS_FLAG_NO_MFG,
 	.close = uart_close,
 	.gas_map = uart_gas_map,
 
@@ -536,4 +537,3 @@ err_free:
 }
 
 #endif
-
diff --git a/lib/switchtec.c b/lib/switchtec.c
index 395c4aa27924..be56c0048d74 100755
--- a/lib/switchtec.c
+++ b/lib/switchtec.c
@@ -537,7 +537,7 @@ const char *switchtec_strerror(void)
 	case ERR_EVENT_INVALID: 	msg = "Invalid Event"; break;
 	case ERR_RST_RULE_FAILED: 	msg = "Reset rule search failed"; break;
 	case ERR_ACCESS_REFUSED: 	msg = "Access Refused"; break;
-
+	case ERR_UART_NOT_SUPPORTED:    msg = "UART interface not supported for this command"; break;
 	default: break;
 	}
 
diff --git a/lib/switchtec_priv.h b/lib/switchtec_priv.h
index 391463d65802..87baa299ea0f 100755
--- a/lib/switchtec_priv.h
+++ b/lib/switchtec_priv.h
@@ -64,7 +64,12 @@ enum switchtec_fw_image_part_id_gen4 {
 	SWITCHTEC_FW_PART_ID_G4_SEEPROM = 0xFE,
 };
 
+enum switchtec_ops_flags {
+	SWITCHTEC_OPS_FLAG_NO_MFG = (1 << 0),
+};
+
 struct switchtec_ops {
+	int flags;
 	void (*close)(struct switchtec_dev *dev);
 	int (*get_device_id)(struct switchtec_dev *dev);
 	int (*get_fw_version)(struct switchtec_dev *dev, char *buf,
```

xyan264: Updated, please review.

xyan264:
Added back CLI-level checking for UART in MFG commands. This provides a more user-friendly error message.

cli/main.c
```diff
@@ -90,6 +90,17 @@ int switchtec_handler(const char *optarg, void *value_addr,
 	return 0;
 }
 
+int mfg_handler(const char *optarg, void *value_addr,
+		const struct argconfig_options *opt)
+{
+	if(strstr(optarg, "tty")) {
```
lsgunth: No, this is junk, the point was you weren't supposed to need to do this because as soon as you do a bad command it returns a sensible error.... Why are you adding this?

xyan264: The previous fix works for MAIN boot phase. 

However, in BL1/2, program fails at 'switchtec_open()', when it tries to read device ID by reading GAS register. BL1/2 does not support UART access of GAS so the function fails with 'Error communicating with device. Please check your setup.'

By adding this check, we block mfg access from CLI level, before switchtec_open().

From an API perspective, in BL1/2, since switchtec_open() of UART will fail, user will not have access to other API functions.

lsgunth: Well then fix the actual problem, don't add hacks like this...

Something along these lines.... But I now suspect that the switchtec_mfg_cmd() is wrong and should be dropped in favor of something like this:

```diff
diff --git a/lib/platform/gasops.c b/lib/platform/gasops.c
index 629412face20..9c8e7fcf4710 100644
--- a/lib/platform/gasops.c
+++ b/lib/platform/gasops.c
@@ -104,7 +104,19 @@ int gasop_cmd(struct switchtec_dev *dev, uint32_t cmd,
 
 int gasop_get_device_id(struct switchtec_dev *dev)
 {
-       return gas_reg_read32(dev, sys_info.device_id);
+       int ret;
+
+       ret = gas_reg_read32(dev, sys_info.device_id);
+
+       if (ret == -1) {
+               if (dev->ops->flags & SWITCHTEC_OPS_FLAG_NO_BL)
+                       errno = ERR_IFACE_NOT_SUPPORTED_IN_BL;
+               else
+                       errno = EIO;
+       }
+
+       return ret;
 }
 
 int gasop_get_fw_version(struct switchtec_dev *dev, char *buf,
diff --git a/lib/switchtec.c b/lib/switchtec.c
index 395c4aa27924..2bb86c05a884 100755
--- a/lib/switchtec.c
+++ b/lib/switchtec.c
@@ -130,6 +130,8 @@ static int set_gen_variant(struct switchtec_dev * dev)
        dev->var = SWITCHTEC_VAR_UNKNOWN;
 
        dev->device_id = dev->ops->get_device_id(dev);
+       if (dev->device_id == -1)
+               return -1;
 
        while (id->device_id) {
                if (id->device_id == dev->device_id) {
```

xyan264: The previous switchtec_mfg_cmd() fix works for MAIN boot phase where the device ID can be read through UART, so it is a valid fix. 

In BL1/2, switchtec_open fails at uart_gas_read() where it does "raise(SIGBUS);", so the above patch will not work. 

Removing the "raise(SIGBUS)" will have a wider impact on other functions... So there is no quick solution. I'll look for another solution.

lsgunth: Add a check_gas_access() op (or something like that) that's called in switchtec_open() and is skipped if not set. The uart platform would provide an implementation that tries a gas read and fails with an appropriate errno if it doesn't work.

This will require a bit of cleanup in the uart platform....

xyan264: Also at this layer, we won't know if a UART gas read fail is because actual communication issue, or because we are in BL1/2 (boot phase is determined AFTER gas read), so we cannot give user a meaningful error code.

xyan264:
I have reverted CLI-level checking so we are back at previous fix that works in MAIN boot phase.

Considering that there is a release build scheduled next week, the BL1/2 boot phase issue is filed as an issue #181 , which will be fixed after next week's release build.

PR #170, opened 2020-01-22
Add prints to some mfg commands so user can verify settings.

xyan264:
Add the following print outs so user can verify the settings:
- In 'config-set', print out the settings in the given file
- In 'state-set', print out the current security settings
- In 'kmsk-entry-add', print out the KMSK entry in the given file


xyan264:
@lsgunth could you review this and approve if it looks OK?

PR #176, opened 2020-02-05
Use 'i2c_gas_write_no_retry' for specific commands

xyan264:
Due to the unreliable nature of I2C communication, function i2c_gas_write() is implemented with automatic retry (100 times). 

This poses a potential issue when a command is critical and expected to be sent only once (e.g., commands that adds a KMSK entry to chip OTP memory). Retrying could cause the command be sent multiple times (and multiple KMSK entry being added, if unlucky).

Some background information: an MRPC command consists of command ID and command data. To send an MRPC command, we send command data first, then send the command ID. Sending command ID triggers execution of the command on the firmware side.

At the moment we can think of 2 ways to fix this:
- We notice that in gasop_cmd() function, the command data is sent using __memcpy_to_gas(), and the command ID is sent using __gas_write32(). So we take out the retry in i2c_gas_write32() function. This is the code committed in this PR. This fix is ad-hoc, and will have impact on other functions that call __gas_write32().
- In i2c_gas_write() function, we check the address of the write, if it's writing to the command ID address, then we do not use retry. This fix is ad-hoc too, but it does not impact other functions that write data to GAS.

Other suggestions are welcome.

### DIFF
--- a/lib/platform/gasops.c
+++ b/lib/platform/gasops.c
@@ -47,6 +47,25 @@
 #define gas_reg_write64(dev, val, reg) __gas_write64(dev, val, \
 						     &dev->gas_map->reg)
 
+static const char gasop_noretry_cmds[] = {
+	[MRPC_SECURITY_CONFIG_SET] = 1,
+	[MRPC_KMSK_ENTRY_SET] = 1,
+	[MRPC_SECURE_STATE_SET] = 1,
+	[MRPC_BOOTUP_RESUME] = 1,
+	[MRPC_DBG_UNLOCK] = 1,
+};
+static const int gasop_noretry_cmds_count = sizeof(gasop_noretry_cmds) /
+					    sizeof(char);
+
+static inline bool gasop_is_no_retry_cmd(uint32_t cmd)
+{
+	cmd &= SWITCHTEC_CMD_MASK;
+
+	if (cmd >= gasop_noretry_cmds_count)
+		return 0;
+	return gasop_noretry_cmds[cmd];
+}
+
 int gasop_access_check(struct switchtec_dev *dev)
 {
 	uint32_t device_id;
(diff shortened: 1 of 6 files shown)


xyan264:
Extensive testings reveal that solution #1 is unstable. 

I have updated the code to solution #2, with specific filtering on gas write address (0x800 = MRPC command ID), gas write data (MFG commands, except FW-TRANSFER command 0x107).

This one passed all testing. Please review again. Apologies for the inconvenience.

lib/platform/linux-i2c.c
```diff
@@ -69,6 +69,11 @@ struct switchtec_i2c {
 	((struct switchtec_i2c *) \
 	 ((char *)d - offsetof(struct switchtec_i2c, dev)))
 
+#define IS_MRPC_CMD_ADDR(x) (((size_t)x & 0xfff) == 0x800)
+#define IS_MRPC_NO_RETRY_CMD(x) (((x & 0xfff) > 0x100) && \
+				 ((x & 0xfff) < 0x120) && \
+				 ((x & 0xfff) != 0x107))
```
lsgunth: That is way too gross.

Create an `i2c_gas_write32_noretry()` and a `__gas_write32_noretry()`  which uses it for I2C (and is a regular gas_write32 for other platforms)...  then use it, when appropriate, in gasop_cmd().

xyan264: Updated. Please review.

lib/platform/gasops.c
```diff
@@ -47,6 +47,10 @@
 #define gas_reg_write64(dev, val, reg) __gas_write64(dev, val, \
 						     &dev->gas_map->reg)
 
+#define IS_MRPC_CMD_NO_RETRY(x) (((x & 0xfff) > 0x100) && \
+				 ((x & 0xfff) < 0x120) && \
+				 ((x & 0xfff) != 0x107))
```
lsgunth: This is still gross. If we are going to operate on ranges of commands we should at least use proper defines that are called out in enum mrpc_cmd....  Not sure why the we need the "& 0xfff" here either as switchtec_cmd() is only called (and only should be called) with values from enum mrpc_cmd.

Though I'd actually probably prefer something more like this:

```
static const char gasop_noretry_cmds[] = {
    [MRPC_GET_DEV_INFO] = 1,
    [MRPC_SECURITY_CONFIG_GET] = 1,
    ....etc
};

static inline bool gasop_is_no_retry_cmd(uint32_t cmd)
{
        if (cmd >= array_size(gasop_noretry_cmds))
                return 0;
         return gasop_noretry_cmds[cmd];
}
```

xyan264: Good idea. Updated.

The 0xfff has been changed to SWITCHTEC_CMD_MASK. It is needed because the PAX ID will be OR-ed into command ID (bit 18 and up).

lib/platform/gasops.c
```diff
@@ -72,7 +89,24 @@ int gasop_cmd(struct switchtec_dev *dev, uint32_t cmd,
...
+        * This poses a potential issue when a command is critical
+        * and is expected to be sent only once (e.g., command that
+        * adds a KMSK entry to chip OTP memory). Retrying could
+        * cause the command be sent multiple times (and multiple
+        * KMSK entry being added, if unlucky).
+        *
+        * Here we filter out the specific commands and use 'no retry'
+        * version of gas_write32 for these commands.
+        */
+	if (gasop_is_no_retry_cmd(cmd & SWITCHTEC_CMD_MASK))
```
lsgunth: I'd put the mask in the helper function

xyan264: Updated and merged.

PR #182, opened 2020-02-19
Change submenu availability

xyan264:
Make the following changes to submenu availability:
- 'mfg' is only available in Linux build
- 'fabric' is not available by default. Only available in Linux build, when configured with '--with-fabric' option.

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -90,6 +90,25 @@ int switchtec_handler(const char *optarg, void *value_addr,
 	return 0;
 }
 
+/*
+ * The 'mfg' submenu commands are only available in Linux builds.
+ *
+ * Due to the difference in driver architecture, supporting Windows
+ * build is non-trivial. After evaluating the development effort
+ * and the resources available, we have decided to remove Windows
+ * build support from release roadmap.
+ * 
+ */
+int mfg_handler(const char *optarg, void *value_addr,
+		const struct argconfig_options *opt)
+{
+#ifndef __linux__
+	printf("WARNING: MFG COMMANDS ARE NOT SUPPORTED ON YOUR CURRENT OPERATING SYSTEM!\n"
+	       "Use this command at your own risk!!!\n\n\n");
+#endif
+	return switchtec_handler(optarg, value_addr, opt);
+}
+
 int pax_handler(const char *optarg, void *value_addr,
 		const struct argconfig_options *opt)
 {
--- a/cli/common.h
+++ b/cli/common.h
@@ -28,6 +28,8 @@
 int ask_if_sure(int always_yes);
 int switchtec_handler(const char *optarg, void *value_addr,
 		      const struct argconfig_options *opt);
+int mfg_handler(const char *optarg, void *value_addr,
+		const struct argconfig_options *opt);
 int pax_handler(const char *optarg, void *value_addr,
 		const struct argconfig_options *opt);
 enum switchtec_fw_type check_and_print_fw_image(int img_fd,
@@ -52,11 +54,11 @@ enum switchtec_fw_type check_and_print_fw_image(int img_fd,
 
 #define UART_HELP_TEXT " * a UART path (/dev/ttyUSB0)\n"
 
-#define DEVICE_OPTION_BASIC(extra_text) \
+#define DEVICE_OPTION_BASIC(extra_text, handler) \
 	{ \
 			"device", .cfg_type=CFG_CUSTOM, .value_addr=&cfg.dev, \
 			.argument_type=required_positional, \
-			.custom_handler=switchtec_handler, \
+			.custom_handler=handler, \
 			.complete="/dev/switchtec*", \
 			.env="SWITCHTEC_DEV", \
 			.help="Switchtec device to operate on. Can be any of:\n" \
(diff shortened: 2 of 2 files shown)


cli/mfg.c
```diff
@@ -22,6 +22,8 @@
  *
  */
 
+#ifdef __linux__
+
```
lsgunth: Why?  Is this really necessary? I don't see any good reason to disable this except "recommendations".

At worst, I'd suggest printing a warning in Windows and say it's not supported. Not building it entirely is just a pain given that there's no good reason why it wouldn't work.

If you really must do this,put a gigantic comment explaining why and preferably a much more detailed commit message why you're doing this.

cli/fabric.c
```diff
@@ -37,6 +38,8 @@
 #include <errno.h>
 #include <ctype.h>
 
+#if ENABLE_FABRIC_MENU
```
lsgunth: Nak.

This is a horrible idea. Compile time options should only be added when there is an optional *dependency*. 

Don't add them for silly policy reasons. Downstream users (think distro's who want to distrobute *binaries*) will hate you, or just enable them all by default. Don't make people re-compile for these reasons. 

A built binary should have all possible features that you can provide given the compile environment.

xyan264: The 'fabric' menu needs to be removed from the upcoming PSX release because fabric commands are not supported by PSX devices. 

The PAX release (which supports fabric commands) is still a few months out, and in that release, we will remove the #ifdef so user can have the full feature (regardless of whether they're using PSX or PAX). 

At this moment, we don't want to distract PSX user with 'fabric' submenu commands and create more explaining work for the support team. 

@lsgunth let me know what you think.

cli/mfg.c
xyan264: 'mfg' commands are used for configuring security settings and recovering from boot failures, and are most often used with an I2C device. While Linux provides a unified I2C abstraction layer, making it easy to support different I2C adapters, there is no equivalent I2C abstraction layer on Windows, so it becomes extremely difficult to support different I2C drivers. Due to this reason, the 'mfg' commands are removed from non-Linux build.

lsgunth: Creating a hard rule based on a "most often" statement seems silly to me.

If the commands can be done over PCI then allow them to be done over PCI. 

If they can only be done over I2C then encode that rule which has nothing to do with Linux/Windows. 

But don't create arbitrary rules based on weird expectations.

cli/fabric.c
lsgunth: If you want to do a release without parts of the code then create a PSX branch without those patches and release from there.

cli/mfg.c
xyan264: I agree with your points. From a technical perspective, we should fix the problem at the root, instead of creating a build rule that masks it.

However, the decision to remove Windows build is more business-related. The Windows customer base and the current development resource we have do not warrant supporting Windows build, plus we are short in testing/verification resources as well. As the 'mfg' commands have potential to brick a device if used incorrectly, it is not prudent to release the Windows tool without sufficient testing. It is based on these factors that the Windows build was dropped from release roadmap.

Having said that, I would be happy to find a middle ground where the technical and business considerations can meet. Feel free to propose other solutions if any.

lsgunth: Well, like I said, add a *warning* saying that these commands are not supported by Microsemi in Windows. Because that's what is actually happening. You could make it prompt with a Y/N continue or something like that.

cli/fabric.c
xyan264: Creating a new branch and release without the ~10 fabric-related commits is doable, but it incurs more effort in development and testing than previously planned, since the commits could affect other files in the projects, not just the fabric.c/h. 

While the "#if ENABLE_FABRIC_MENU" does look slightly out of place, considering the current RTP release deadline, I think it provides a good trade off between code quality and test effort, and we maintain our tradition of releasing from 'master' branch as well.

lsgunth: It makes more sense to do the release branch, isn't actually any more effort (really it's the same testing seeing you're either removing the code through an #if or via a commit). If you want to do the branch with a new commit that just removes all the code between the #ifs, that's fine too.

Adding the compile time define just leaves it all in for people to use anyway and isn't what you want because it will have to be removed later anyway (ie. more work).

cli/mfg.c
xyan264: I have updated the code to check for OS in `mfg_handler()`, print an error then exit. 

I think it is better to exit the program than continuing, since the command have not been tested on Windows.

cli/fabric.c
xyan264: We will create a new branch for PSX/PFX release. This commit has been dropped in this PR.

cli/mfg.c
xyan264: @lsgunth please take a look and let me know if it looks OK.

cli/main.c
```diff
@@ -90,6 +90,25 @@ int switchtec_handler(const char *optarg, void *value_addr,
...
+ * and the resources available, we have decided to remove Windows
+ * build support from release roadmap.
+ * 
+ */
+int mfg_handler(const char *optarg, void *value_addr,
+		const struct argconfig_options *opt)
+{
+#ifndef __linux__
+	fprintf(stderr, "Error: 'mfg' submenu commands are only supported in Linux build!\n");
+	return -1;
```
lsgunth: I'd much rather see this as warning than a hard error. Seeing there's no technical reason they would not work.

xyan264: OK, updated to print a warning message and continue. @lsgunth please take a look.

PR #190, opened 2020-03-17
Fix MRPC errors found on secured devices

xyan264:
Two issues were found on secured devices:
- 'switchtec_get_fw_version' fails when MRPC_GAS_READ command is whitelisted
- The error code displayed does not reflect the true error condition

These two issues are fixed in the 2 commits in this PR. Details are explained in the commit messages.


xyan264:
Insufficient testing, closed for now.

xyan264:
More testing shows this fix is sufficient. @lsgunth @kelvin-cao please review and comment.

PR #86, opened 2019-03-25
Update CONTRIBUTING.md file

wesleywesley:
Now that microsemi/switchtec-user repository equipped with travis-ci,
we can ask contributor create pull request against this repo.

Update the "pull request" section, to ask contributors create PR
based on the devel branch.

Signed-off-by: Wesley Sheng <wesley.sheng@microchip.com>

PR #191, opened 2020-03-18
Add reading error code when IB driver returns EIO.

xyan264:
When IB driver returns EIO after executing an MRPC command, an error code is also returned in the return buffer. This error code is read and returned to upper layer.

This PR should be reviewed along with switchtec-kernel PR #86

### DIFF
--- a/lib/platform/linux.c
+++ b/lib/platform/linux.c
@@ -329,9 +329,13 @@ static int read_resp(struct switchtec_linux *ldev, void *resp,
 
 	ret = read(ldev->fd, buf, bufsize);
 
-	if (ret < 0)
+	if (ret < 0) {
+		if (errno == EIO) {
+			memcpy(&ret, buf, sizeof(ret));
+			errno = ret;
+		}
 		return ret;
-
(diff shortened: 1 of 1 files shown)


kelvin-cao:
With below kernel patch, we don't need to update the user code, as current user code has already support TWI when status is error.  

https://github.com/Microsemi/switchtec-kernel/pull/87

xyan264:
@kelvin-cao This fix is for in-band driver error code handling (file: lib/platform/linux.c). 

TWI handling (file: lib/platform/linux-i2c.c) is already supported, even without the Kernel driver fix Microsemi/switchtec-kernel#87

kelvin-cao:
I don't think we should add a new error EIO in the kenrel to force user code (not only switchtec-user) to update accordingly. Instead, we should take care of the internal SWITCHTEC_MRPC_STATUS_ERROR status and just return user the real error via the return code.

xyan264:
This change is no longer needed after switchtec-kernel PR#87 is merged.

PR #194, opened 2020-03-25
Resolve big endian issues

cjengel:
This resolves a few of the endian issues I ran into when running this tool on our PowerPC system with an i2c attached switch.  I did not run through all possible features of the switchtec tool.

### DIFF
--- a/lib/platform/linux-i2c.c
+++ b/lib/platform/linux-i2c.c
@@ -29,6 +29,7 @@
 #include "switchtec/switchtec.h"
 #include "gasops.h"
 
+#include <endian.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/ioctl.h>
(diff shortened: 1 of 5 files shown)


lib/platform/linux-i2c.c
```diff
@@ -557,19 +558,35 @@ static ssize_t i2c_write_from_gas(struct switchtec_dev *dev, int fd,
...
+	uint8_t ret;
+	i2c_memcpy_from_gas(dev, &ret, addr, sizeof(ret));
+	return ret;
+}
+static uint16_t i2c_gas_read16(struct switchtec_dev *dev,
+                               uint16_t __gas *addr)
+{
+	uint16_t ret;
+	i2c_memcpy_from_gas(dev, &ret, addr, sizeof(ret));
+	return le16toh(ret);
```
lsgunth: Can we not do the same trick as above to avoid open coding this macro?

lsgunth:
We have to also resolve the build failures in windows.

cjengel:
Fixed another issue and squashed the commits into a different pull request

PR #197, opened 2020-04-14
Fix 'fw-info' version display issue.

xyan264:
Function 'dev->ops->get_fw_version' gets current firmware version of the local PAX device only. If user is running 'fw-info' on a remote PAX device, the version displayed is incorrect. We should use the version info from 'switchtec_fw_part_summary()' instead to determine current running firmware version.

Function 'print_fw_part_info' is re-structured to maintain program output consistency.

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -1270,15 +1270,10 @@ static void print_fw_part_line(const char *tag,
 	       inf->valid ? "" : "  (Invalid)");
 }
 
-static int print_fw_part_info(struct switchtec_dev *dev)
+static int print_fw_part_info(struct switchtec_fw_part_summary *sum)
 {
-	struct switchtec_fw_part_summary *sum;
 	struct switchtec_fw_image_info *inf;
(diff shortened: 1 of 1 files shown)


kelvin-cao:
I think we should update the switchtec_get_fw_version() to make it leveraging switchtec_fw_part_summary() to get the fw version.

lsgunth:
Yes, switchtec_get_fw_version() might be unusable... 

The only other caller of switchtec_get_fw_version() is print_dev_info()... If it is also broken in the same way than it seems like the existing switchtec_get_fw_version() is not useful.

xyan264:
Closed in favor of alternative approach.

kelvin-cao:
Yes, in theory user can get the version info by switchtec_fw_part_summary(). But to keep the compatibility, let's still keep the switchtec_get_fw_version() and just fix it by making it leveraging switchtec_fw_part_summary().

PR #198, opened 2020-04-20
Add app log parsing support to log_dump command

brian-mckenzie:
Added a new parameter "log_def_file" to the log_dump command. This allows the user to pass in a log definition data file, which is used for parsing app logs into human-readable output. This parsed output is written to the output file instead of the raw binary log.


cli/main.c
```diff
@@ -1,6 +1,6 @@
 /*
  * Microsemi Switchtec(tm) PCIe Management Command Line Interface
- * Copyright (c) 2017, Microsemi Corporation
+ * Copyright (c) 2020, Microsemi Corporation
```
kelvin-cao, line 3: I'm not sure if we need to update this. It seems the year here should be the "first year of the publication". Comment?

cli/main.c
```diff
@@ -905,31 +905,44 @@ static int log_dump(int argc, char **argv)
...
 	};
 	const struct argconfig_options opts[] = {
 		DEVICE_OPTION,
-		{"filename", .cfg_type=CFG_FD_WR, .value_addr=&cfg.out_fd,
+		{"output_file", .cfg_type=CFG_FD_WR, .value_addr=&cfg.out_fd,
 		  .argument_type=optional_positional,
 		  .force_default="switchtec.log",
-		  .help="image file to display information for"},
+		  .help="log output file"},
+        {"log_def_file", .cfg_type=CFG_FILE_R, .value_addr=&cfg.log_def_file,
```
kelvin-cao, line 920: It seems your editor generates leading white space... We use tab for indent.

cli/main.c
lsgunth: Yes, it's really meant to be first year of publication. Though the year isn't that important, we have git logs to show when things were first published.

More importantly, we don't need a bunch of churn in every commit changing every year.

cli/main.c
```diff
@@ -905,31 +905,44 @@ static int log_dump(int argc, char **argv)
...
+    ret = switchtec_log_to_file(cfg.dev, cfg.type, cfg.out_fd, cfg.log_def_file);
 	if (ret < 0) {
 		switchtec_perror("log");
-		return ret;
+	} else {
+		fprintf(stderr, "\nLog saved to %s.\n", cfg.out_filename);
 	}
 
-	fprintf(stderr, "\nLog saved to %s.\n", cfg.out_filename);
+	if (cfg.out_fd > 0) {
```
lsgunth, line 937: Nit: unnecessary braces on single line if

lib/switchtec.c
```diff
@@ -652,33 +673,215 @@ int switchtec_hard_reset(struct switchtec_dev *dev)
...
+				return -1;
+			}
+			log_defs[mod_id].num_entries = num_entries;
+
+			strncpy(log_defs[mod_id].mod_name, mod_name, FW_MOD_NAME_LEN_MAX);
+
+			/* extract each log entry for the module */
+			for (entry_idx = 0; entry_idx < num_entries; entry_idx++) {
+				if (fgets(line, sizeof(line), log_def_file) == NULL) {
+					fprintf(stderr, "Failed to get log entry %d for module %s\n", entry_idx, log_defs[mod_id].mod_name);
```
lsgunth, line 712: NIT: break this long line

cli/main.c
lsgunth, line 920: I think log_def_file should be an option argument with a flag, not a positional argument.

lsgunth, line 920: I think log_def_file should be an option argument with a flag, not a positional argument.

lib/switchtec.c
lsgunth, line 712: NIT: break this long line

lib/switchtec.c
```diff
@@ -652,33 +673,215 @@ int switchtec_hard_reset(struct switchtec_dev *dev)
...
+    			log_data[i].data[4], log_data[i].data[5], log_data[i].data[6], log_data[i].data[7]);
+
+		entry_idx++;
+    }
+
+	fflush(log_file);
+    return 0;
+}
+
+static int log_a_to_file(struct switchtec_dev *dev, int sub_cmd_id, int fd, FILE *log_def_file)
```
xyan264, line 830: Break this line.

cli/main.c
brian-mckenzie: I did this since for FW it's always been a requirement to update the copyright date whenever we touch a file. If that's not a requirement for this project then I'm good with that -- I agree that it's nice to avoid the churn.

cli/main.c
brian-mckenzie: It generally sticks to what the file is using, but I guess it missed that one. I'll fix it and make sure the rest is fine.

brian-mckenzie: Agreed, that would be nicer. I'll change it.

brian-mckenzie:
> There're some coding style issues. Please refer to CONTRIBUTING.md for the code style info.
> 
> It seems we can only do the parsing when we dump it, then we cannot offline parse a dumped log. Maybe seperate steps for dump and parse make more sense. What do you think? @brian-mckenzie @hachevin

I wasn't sure if there was a coding style document for this -- thanks for pointing it out.

I agree it might be nice to parse an offline log. If we're doing that then we may want a new command (not log_dump), as log_dump requires the device as an argument. Maybe we should add "log_parse". It could just have two required arguments of log_def_file and output_file. And then we'll remove the parsing functionality from log_dump entirely. What do you guys think?

lib/switchtec.c
brian-mckenzie: Will do.

lib/switchtec.c
brian-mckenzie: Will do.

lib/switchtec.c
```diff
@@ -652,33 +673,215 @@ int switchtec_hard_reset(struct switchtec_dev *dev)
...
+
+		/* validate the entry number */
+		if ((entry_num >= log_defs[mod_id].num_entries) ||
+		    (strlen(log_defs[mod_id].entries[entry_num]) == 0)) {
+			fprintf(log_file, "(Invalid log entry number: %d)\n", entry_num);
+			continue;
+		}
+
+		/* stringify and validate the log severity  */
+        switch (log_sev) {
```
xyan264, line 792: Maybe it's easier if we use something like this?

```
const char *sev_strs= {"DISABLED", "HIGHEST", ..."LOWEST"};
...
if(log_sev > 5) {
    fprintf(log_file, "(Invalid log severity: %d)\n", log_sev);
    continue;
}
log_sev_str = sev_strs[log_sev];
```

cli/main.c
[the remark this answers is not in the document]
brian-mckenzie: I'll update these instances to remove the braces.

xyan264:
Overall it looks fine to me. 

Just want to highlight some points in coding style (that's different from FW):
- Use tabs (tab size = 8) for indentation (FW uses 4 spaces)
- Break long lines to within <80 chars long (FW does not have this requirement)

brian-mckenzie:
> Github seems to be broken right now so some of my comments may be duplicated or missing.... And I'll probably have more specific stuff on the next pass..
> 
> More generally, I'm very happy to see the log information getting parsed into a human readable format.
> 
> However, I'm concerned about this log_def_file... Where are users supposed to get it from? And why should they need to put in any extra effort to read their logs?
> 
> I'd suggest that we include a def file (or more, if for example different versions have different log files) in this repository, ensure it's installed on the user's system and use it by default for dumping logs. (An option can be added to dump the raw binary if someone actually wants that).
> 
> With that in mind it may make sense to have a few more advance features with the log def files, so users that need to customize it can still include the default log definitions without having to duplicate everything.

The log_def_file is generated on every FW build, and is part of our FW releases. We could potentially include a log def file in our matching switchtec-user releases, but FW log definitions are constantly changing so if switchtec-util users just use the provided def file this could be invalid when using newer (or older) FW. They'd need to remember to use the default file only when the switchtec-user release matches the FW release, but I could see them forgetting to keep this in mind. It would also mean we'd need to make new switchtec-user releases to match each FW release even when the only thing that changed in switchtec-user was the log def file. @hachevin and @kelvin-cao, what do you think?

brian-mckenzie:
> Overall it looks fine to me.
> 
> Just want to highlight some points in coding style (that's different from FW):
> 
> * Use tabs (tab size = 8) for indentation (FW uses 4 spaces)
> * Break long lines to within <80 chars long (FW does not have this requirement)

I'll run through the changes and make sure these are enforced. Thanks.

lib/switchtec.c
brian-mckenzie: Sure, I'll do that.

hachevin:
> There're some coding style issues. Please refer to CONTRIBUTING.md for the code style info.
> 
> It seems we can only do the parsing when we dump it, then we cannot offline parse a dumped log. Maybe seperate steps for dump and parse make more sense. What do you think? @brian-mckenzie @hachevin

Yes, it might be useful to pass in a previously captured log file. The primary method for offline decoding would be ChipLink, but as a future improvement we could reuse the parsing code in switchtec-user for that purpose.

hachevin:
> > There're some coding style issues. Please refer to CONTRIBUTING.md for the code style info.
> > It seems we can only do the parsing when we dump it, then we cannot offline parse a dumped log. Maybe seperate steps for dump and parse make more sense. What do you think? @brian-mckenzie @hachevin
> 
> I wasn't sure if there was a coding style document for this -- thanks for pointing it out.
> 
> I agree it might be nice to parse an offline log. If we're doing that then we may want a new command (not log_dump), as log_dump requires the device as an argument. Maybe we should add "log_parse". It could just have two required arguments of log_def_file and output_file. And then we'll remove the parsing functionality from log_dump entirely. What do you guys think?

Whatever changes we make, we'll want to ensure we don't break the ease of use of parsing the log as part of the dump process. So, if we remove the option from log_dump, we'd want to make sure we can pipe its output into log_parse.

brian-mckenzie:
> > > There're some coding style issues. Please refer to CONTRIBUTING.md for the code style info.
> > > It seems we can only do the parsing when we dump it, then we cannot offline parse a dumped log. Maybe seperate steps for dump and parse make more sense. What do you think? @brian-mckenzie @hachevin
> > 
> > 
> > I wasn't sure if there was a coding style document for this -- thanks for pointing it out.
> > I agree it might be nice to parse an offline log. If we're doing that then we may want a new command (not log_dump), as log_dump requires the device as an argument. Maybe we should add "log_parse". It could just have two required arguments of log_def_file and output_file. And then we'll remove the parsing functionality from log_dump entirely. What do you guys think?
> 
> Whatever changes we make, we'll want to ensure we don't break the ease of use of parsing the log as part of the dump process. So, if we remove the option from log_dump, we'd want to make sure we can pipe its output into log_parse.

I don't think switchtec-user has general support for piped input at the moment. If we want this then we should probably to build it into argconfig.c. This might take some time to do properly and if we were doing it then it should probably be in a separate commit.

If we want to be able to parse as part of the log_dump process and also parse a stand-alone input file, it might be best to just keep the new log_dump argument so that it's parsed from there (though make it a flagged parameter as Logan suggested), and also add a new command for stand-alone log file parsing (e.g. log_parse). Thoughts?

kelvin-cao:
> 
> 
> > > > There're some coding style issues. Please refer to CONTRIBUTING.md for the code style info.
> > > > It seems we can only do the parsing when we dump it, then we cannot offline parse a dumped log. Maybe seperate steps for dump and parse make more sense. What do you think? @brian-mckenzie @hachevin
> > > 
> > > 
> > > I wasn't sure if there was a coding style document for this -- thanks for pointing it out.
> > > I agree it might be nice to parse an offline log. If we're doing that then we may want a new command (not log_dump), as log_dump requires the device as an argument. Maybe we should add "log_parse". It could just have two required arguments of log_def_file and output_file. And then we'll remove the parsing functionality from log_dump entirely. What do you guys think?
> > 
> > 
> > Whatever changes we make, we'll want to ensure we don't break the ease of use of parsing the log as part of the dump process. So, if we remove the option from log_dump, we'd want to make sure we can pipe its output into log_parse.
> 
> I don't think switchtec-user has general support for piped input at the moment. If we want this then we should probably to build it into argconfig.c. This might take some time to do properly and if we were doing it then it should probably be in a separate commit.
> 
> If we want to be able to parse as part of the log_dump process and also parse a stand-alone input file, it might be best to just keep the new log_dump argument so that it's parsed from there (though make it a flagged parameter as Logan suggested), and also add a new command for stand-alone log file parsing (e.g. log_parse). Thoughts?

Yes, I agree.

kelvin-cao:
> 
> 
> > Github seems to be broken right now so some of my comments may be duplicated or missing.... And I'll probably have more specific stuff on the next pass..
> > More generally, I'm very happy to see the log information getting parsed into a human readable format.
> > However, I'm concerned about this log_def_file... Where are users supposed to get it from? And why should they need to put in any extra effort to read their logs?
> > I'd suggest that we include a def file (or more, if for example different versions have different log files) in this repository, ensure it's installed on the user's system and use it by default for dumping logs. (An option can be added to dump the raw binary if someone actually wants that).
> > With that in mind it may make sense to have a few more advance features with the log def files, so users that need to customize it can still include the default log definitions without having to duplicate everything.
> 
> The log_def_file is generated on every FW build, and is part of our FW releases. We could potentially include a log def file in our matching switchtec-user releases, but FW log definitions are constantly changing so if switchtec-util users just use the provided def file this could be invalid when using newer (or older) FW. They'd need to remember to use the default file only when the switchtec-user release matches the FW release, but I could see them forgetting to keep this in mind. It would also mean we'd need to make new switchtec-user releases to match each FW release even when the only thing that changed in switchtec-user was the log def file. @hachevin and @kelvin-cao, what do you think?

Yeah, these concerns are valid. I'd think it makes more sense the log_def_file goes with firmware release only, and switchtec-user only provide the method for user to do the dump and parsing. (I used to suggest we add a parser together with the log_def_file in the firmware release package, so that user will definitly use the right log_def_file for the firmware, and whenever they find the log_def_file they get the parser too. However, some customers would like to use switchtec-user for it, so we add it here. )

hachevin:
> > Github seems to be broken right now so some of my comments may be duplicated or missing.... And I'll probably have more specific stuff on the next pass..
> > More generally, I'm very happy to see the log information getting parsed into a human readable format.
> > However, I'm concerned about this log_def_file... Where are users supposed to get it from? And why should they need to put in any extra effort to read their logs?
> > I'd suggest that we include a def file (or more, if for example different versions have different log files) in this repository, ensure it's installed on the user's system and use it by default for dumping logs. (An option can be added to dump the raw binary if someone actually wants that).
> > With that in mind it may make sense to have a few more advance features with the log def files, so users that need to customize it can still include the default log definitions without having to duplicate everything.
> 
> The log_def_file is generated on every FW build, and is part of our FW releases. We could potentially include a log def file in our matching switchtec-user releases, but FW log definitions are constantly changing so if switchtec-util users just use the provided def file this could be invalid when using newer (or older) FW. They'd need to remember to use the default file only when the switchtec-user release matches the FW release, but I could see them forgetting to keep this in mind. It would also mean we'd need to make new switchtec-user releases to match each FW release even when the only thing that changed in switchtec-user was the log def file. @hachevin and @kelvin-cao, what do you think?



> > > Github seems to be broken right now so some of my comments may be duplicated or missing.... And I'll probably have more specific stuff on the next pass..
> > > More generally, I'm very happy to see the log information getting parsed into a human readable format.
> > > However, I'm concerned about this log_def_file... Where are users supposed to get it from? And why should they need to put in any extra effort to read their logs?
> > > I'd suggest that we include a def file (or more, if for example different versions have different log files) in this repository, ensure it's installed on the user's system and use it by default for dumping logs. (An option can be added to dump the raw binary if someone actually wants that).
> > > With that in mind it may make sense to have a few more advance features with the log def files, so users that need to customize it can still include the default log definitions without having to duplicate everything.
> > 
> > 
> > The log_def_file is generated on every FW build, and is part of our FW releases. We could potentially include a log def file in our matching switchtec-user releases, but FW log definitions are constantly changing so if switchtec-util users just use the provided def file this could be invalid when using newer (or older) FW. They'd need to remember to use the default file only when the switchtec-user release matches the FW release, but I could see them forgetting to keep this in mind. It would also mean we'd need to make new switchtec-user releases to match each FW release even when the only thing that changed in switchtec-user was the log def file. @hachevin and @kelvin-cao, what do you think?
> 
> Yeah, these concerns are valid. I'd think it makes more sense the log_def_file goes with firmware release only, and switchtec-user only provide the method for user to do the dump and parsing. (I used to suggest we add a parser together with the log_def_file in the firmware release package, so that user will definitly use the right log_def_file for the firmware, and whenever they find the log_def_file they get the parser too. However, some customers would like to use switchtec-user for it, so we add it here. )

The log_def file is part of NDA-protected design collateral, so we wouldn't be including it as part of this package. Our customers also have the ability to customize its contents and the logs generated by the device, so we won't have the ability to include an appropriate log_def file for each platform using a Switchtec device. 

Ultimately, I don't think it's a big issue.  Users of 'log_dump' will fall into two categories:
- System developers or administrators trying to diagnose or debug sightings on a platform. These individuals will have intimate knowledge of the platform and will have access to its design collateral, so they'll be able to get their hands on the correct log_def file.
- People just messing around with system tools when they shouldn't be. The contents of the logs will mean nothing to them, even when parsed, so they don't really need a copy of the log_def file.

lsgunth:
> The log_def_file is generated on every FW build, and is part of our FW releases. We could potentially include a log def file in our matching switchtec-user releases, but FW log definitions are constantly changing so if switchtec-util users just use the provided def file this could be invalid when using newer (or older) FW. They'd need to remember to use the default file only when the switchtec-user release matches the FW release, but I could see them forgetting to keep this in mind. It would also mean we'd need to make new switchtec-user releases to match each FW release even when the only thing that changed in switchtec-user was the log def file. @hachevin and @kelvin-cao, what do you think?

By constantly changing, I assume the newer firmware is only adding new types and old types still stay valid. I also expect that this doesn't change quite as often as you say. So just try to keep the file in the repo up to date and if the user has an out of date file they just might see some unknown log entries or something and they have the option of getting a more update version through other means.

hachevin:
> By constantly changing, I assume the newer firmware is only adding new types and old types still stay valid. I also expect that this doesn't change quite as often as you say. So just try to keep the file in the repo up to date and if the user has an out of date file they just might see some unknown log entries or something and they have the option of getting a more update version through other means.

You're right that the file won't change often and our log entries won't be removed or changed (we don't have control over whether customers delete old types). That said, we won't be adding a log_def file to the repo. We still consider this NDA-protected collateral that wouldn't be released openly in the repo.

lsgunth:
Well that sounds particularly hostile to your users. An NDA to protect information on how to decode logs? That's mad. The information is literally only useful to a users of your hardware; competitors would never care or want to copy your log formats.

hachevin:
> The information is literally only useful to a users of your hardware; competitors would never care or want to copy your log formats.

Sounds like you haven’t dealt with our competitors :) I guarantee you that within weeks of posting a log_def file to the repo we’ll receive concerned emails from ODM customers claiming they were told by our competitor’s sales team that our part has major stability issues related [insert log decode data here]. Technically speaking, it’s only useful to system designers building with our switch, not users of the switch. For example, a data center admin would use Switchtec-user to update FW on a platform, would have no idea the meaning or significance of log messages about “GPIO X asserted”. Our customers designing these platforms wouldn’t be releasing their log entry definition files, either. 

The file doesn’t define to format of the log entries - that can be inferred from the code. The file lists in plaintext all the events that could take place. That includes descriptions of the order of operations of proprietary features.

cli/main.c
```diff
@@ -905,31 +905,103 @@ static int log_dump(int argc, char **argv)
...
+		 .argument_type = required_positional,
+		 "log definition file"},
+		{"parsed_output", .cfg_type = CFG_FILE_W,
+		 .value_addr = &cfg.parsed_log_file,
+		 .argument_type = optional_positional,
+		 .force_default = "app_log.txt",
+		 .help = "parsed output file"},
+		{NULL}};
+
+	argconfig_parse(argc, argv, CMD_DESC_LOG_PARSE, opts, &cfg, sizeof(cfg));
```
xyan264, line 985: Wrap this line.

brian-mckenzie: Will do.

lib/switchtec.c
```diff
@@ -652,33 +674,223 @@ int switchtec_hard_reset(struct switchtec_dev *dev)
...
+		/*
+		 * Parse the module name, module ID, and number of module log
+		 * entries
+		 */
+		if (sscanf(line, "%31s %i %i",
+			   mod_name, &mod_id, &num_entries) != 3)
+			continue;
+
+		/* validate the module ID */
+		if (mod_id > FW_MOD_ID_MAX) {
```
lsgunth, line 701: I'd probably re-code this to allocate the output dynamically instead of hard coding random limits which may cause issues down the road and wastes a lot of memory. There are a number of approaches to this, depending on what the actual data looks like but I'm thinking something along the lines of using realloc... roughly (note I didn't even compile test this):

```
struct module_log_defs {
	char *mod_name;		 
	char **entries; 
	int num_entries;			
};

static void free_log_defs(struct module_log_defs *defs, int count)
{
	int i, j; 

	if (!defs)
		return;

	for (i = 0; i < count; i++) {
		free(defs[i].mod_name);
		if (!defs[i].entries)
			continue;
		for (j = 0; j < defs[i].num_entries; j++) {
			free(defs[i].entries[j]);
		}
		free(defs[i].entries);
	}
	free(defs);
}

static struct module_log_defs *realloc_log_defs(struct module_log_defs *old, 
						int num_entries, int *allocated)
{
	struct module_log_defs *new;
	int i;
        
	new = reallocarray(old, num_entries, sizeof(*new));
	if (!new) {
		free_log_defs(old, *allocated);
		return NULL;
	}
        
	for (i = *allocated; i < num_entries; i++) {
		new[i].mod_name = NULL;
		new[i].entries = NULL;
	}
           
	*allocated = num_entries;

	return new;
}

static struct module_log_defs *read_log_defs(FILE *log_def_file, log_defs)
{ 
	struct module_log_defs *defs = NULL;
	int mod_id, num_entries;  	
	int mods_allocated = 0;
	char *line, *mod_name;
	int ret, i;

	while (fscanf(log_def_file, "%ms", &line)) {
		ret = sscanf(line, "%ms %i %i", &mod_name, &mod_id, 
			     &num_entries);
		free(line);
		if (ret != 3) 
			continue;

		if (mod_id >= mods_allocated)
			defs = realloc_log_defs(defs, mod_id * 2, 
						&mods_allocated);
		if (!defs) {
			free(mod_name);
			return NULL;
		}

		defs[mod_id].mod_name = mod_name;
		defs[mod_id].num_entries = num_entries;
		defs[mod_id].entries = calloc(num_entries, 
					      sizeof(*defs->entries));
		if (!defs[mod_id].entries) {
			free_log_defs(defs, mods_allocated);
			return NULL;
		}

		for (i = 0; i < num_entrise; i++) {
			ret = fscanf(log_def_file, "%ms", 
				     &defs[mod_id].entries[i]);
			if (!ret) {
				errno = SWITCHTEC_ERR_LOGDEFS_PARSE_ERR;
				free_log_defs(defs, mods_allocated);
				return NULL;                                
			}
		}
			
	}

	return defs;
}
```

lib/switchtec.c
```diff
@@ -777,6 +991,62 @@ int switchtec_log_to_file(struct switchtec_dev *dev,
...
+	int num_entries;
+	int entry_idx = 0;
+
+	log_data = calloc(PARSE_LOG_INPUT_BUF_SIZE, sizeof(struct log_a_data));
+
+	log_defs = calloc((FW_MOD_ID_MAX + 1), sizeof(struct module_log_defs));
+
+	if (log_defs == NULL)
+		return -1;
+
```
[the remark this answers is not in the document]
brian-mckenzie: Sure, I'll read and parse one record at a time.

lib/switchtec.c
```diff
@@ -652,33 +674,223 @@ int switchtec_hard_reset(struct switchtec_dev *dev)
...
+		/* validate the module ID */
+		if (mod_id > FW_MOD_ID_MAX) {
+			fprintf(stderr, "Invalid module ID: %d\n", mod_id);
+			return -1;
+		}
+
+		/* validate the number of log entries */
+		if (num_entries >= LOG_ENTRIES_MAX) {
+			fprintf(stderr, "Invalid number of log entries: %d\n",
```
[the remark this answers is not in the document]
brian-mckenzie: Will do.

lib/switchtec.c
brian-mckenzie: I agree that allocating things dynamically would be nice, particularly for memory wastage. One thing I don't think we can do though is use the 'm' scanf modifier to allocate strings, as it's not portable. It doesn't even work in MSYS2.

What we could do instead is to read each line into a large fixed-size buffer. Lines would be read one at a time so we'd only need one such buffer. Then when storing the log text we can calloc each entry based on the string size.

For the module name, these are usually around 3-4 characters and typically no more than 10. A fixed 16-byte buffer should be fine, and worst case if there happened to be a huge module name we can just truncate it. Keeping it at a known fixed size also simplifies printing the heading.

I do think the number of modules and the number of entries per module should be dynamically allocated, as you're suggesting, as this could save a lot of memory. A minor thing I noticed is that your sample code is allocating room for (mod_id * 2) modules when it runs out of space. This might be excessive as the module IDs get into the hundreds. How about something like (mod_id + 20), so we're guaranteed not to over-allocate by 20 modules?

lsgunth: Oh, yes, if msys2 doesn't support it then you can't use it. You'll have to have a fixed line buffer and then strdup(). Just be careful with handling a too long module-name...  This change makes the code a bit more complicated but it shouldn't be terrible.

Using *2 for reallocated arrays is pretty common. It's better to have a logarithmic scale than a linear one for reallocation. The size of the struct is also now pretty small so it's not like we are wasting much memory by having it be double the size used. The point is to not reallocate too many times because it requires a lot of memory copying. It may make sense to tweak this to a typical log file. So, if it's typically in the 100s of elements maybe have the initial allocation start at ~200ish. Test with smaller values first though.

brian-mckenzie: Sounds good -- I'll do the fixed-length buffer and strdup() approach.

I'll use *2 for the reallocation then. We currently have about 135 modules and that probably won't grow very much, so I'll start it at 200 (but test using small values).

lib/switchtec.c
```diff
@@ -56,6 +57,35 @@
...
  */
 
+/* maximum firmware module string length */
+#define FW_MOD_NAME_LEN_MAX	32
+
+/* maximum number of log entries for each module */
+#define LOG_ENTRIES_MAX		512
+
+/* maximum log entry string length */
+#define LOG_ENTRY_LEN_MAX	512
```
lsgunth, line 67: We really don't need all these defines at the top of the file. Most of them are only used in one place...

LOG_ENTRIES_MAX isn't even used.

lib/switchtec.c
```diff
@@ -56,6 +57,35 @@
...
+
+/* initial number of log definition modules to allocate */
+#define INIT_ALLOC_LOG_MODULES	200
+
+/**
+ * @brief Module-specific log definitions
+ */
+struct module_log_defs {
+	char mod_name[FW_MOD_NAME_LEN_MAX];	//!< module name
```
lsgunth, line 76: I'd still use strdup for mod name (more on that below)...

lib/switchtec.c
```diff
@@ -652,33 +698,306 @@ int switchtec_hard_reset(struct switchtec_dev *dev)
...
+	ret = realloc_log_defs(defs, INIT_ALLOC_LOG_MODULES);
+	if (ret < 0)
+		return ret;
+
+	while (fgets(line, sizeof(line), log_def_file)) {
+		/*
+		 * Parse the module name, module ID, and number of module log
+		 * entries
+		 */
+		if (sscanf(line, "%31s %i %i",
```
lsgunth, line 775: How about instead something like:

```
        tok = strtok(line, " \t");
        if (!tok)
            continue;
    
        tok = strtok(NULL, " \t");
        if (!tok)
            continue;
        mod_id = strtol(tok, NULL, 10);

        /* realloc, etc */

        tok = strtok(NULL, " \t");
        if (!tok)
            continue;

        mod_def->num_entries = strtol(tok, NULL, 10);
        mod_def->mod_name = strdup(line);
```

Removes the module name length mess.

lib/switchtec.c
```diff
@@ -652,33 +698,306 @@ int switchtec_hard_reset(struct switchtec_dev *dev)
...
+		 * if the module is duplicated in the log definition file.
+		 */
+		if (strlen(mod_defs->mod_name) != 0) {
+			for (i = 0; i < num_entries; i++) {
+				if (!fgets(line, sizeof(line),
+					  log_def_file))
+					break;
+			}
+			continue;
+		}
```
lsgunth, line 799: So if an ID exists twice, the subsequent one is ignored? What if they are different? I'd suggest it's better to just error out instead to help ensure you don't have confusing log-def files with multiple instances of the same module.

brian-mckenzie:
> Getting closer.
> 
> Can we also split this into multiple patches? One for introducing the log def parsing, another for introducing switchtec_parse_log() and a third for adding the log_parse command to the CLI?

Sure, I can split it. There's both the new log_parse CLI command and the new log_def option for the existing log_dump CLI command. Should I group the two CLI updates in one patch or separate them?

lib/switchtec.c
brian-mckenzie: Yep I forgot to remove LOG_ENTRIES_MAX when changing things to dynamically allocate.

I'm used to our FW convention of grouping all #define's at the top of the file. I'll move them above the functions that use them.

lib/switchtec.c
brian-mckenzie: Sure, I'll use strdup. I didn't bother since our module names will never be more than around 12 characters, and malloc block sizes won't be less than 32 bytes anyway.

lib/switchtec.c
brian-mckenzie: I did the sscanf since it's simpler and I know our module names won't be anywhere near 31 characters, but I'll use strtok instead just to be safe.

lib/switchtec.c
brian-mckenzie: Yes, the subsequent module and its log entries will be ignored. If a module is duplicated then all the log entries will be identical as well. This was the case for our log definition files as there was a FW bug that caused some of the modules to be duplicated. The bug was fixed a couple days ago, but we want to allow parsing to still work when using the older definition files.

lib/switchtec.c
lsgunth: My point was you probably didn't need a define at all for most of these things...

Eg the line length can probably just be

`char line[512];`

and use `sizeof(line)` in the function, if it's necessary. No point in adding any kind of global define for it.

lib/switchtec.c
lsgunth: The point is to get rid of the ugly global define that doesn't need to be there and the ugly %31s (where the 31 should actually be the global define).

lsgunth:
If in doubt, separate them into more commits. Unless there's a hard to manage dependency, more commits are better.

lib/switchtec.c
brian-mckenzie: Ok, I'll do that.

brian-mckenzie:
> If in doubt, separate them into more commits. Unless there's a hard to manage dependency, more commits are better.

This basically means a commit for all of the parsing stuff (whatever's in switchtec.c) and then a couple minor commits for the two CLI updates. The parsing stuff in switchtec.c can't really be broken up without the compiler complaining about unused functions. Is that separation fine with you?

lsgunth:
Yes. Each commit must still be correct.

lib/switchtec.c
```diff
@@ -652,33 +684,325 @@ int switchtec_hard_reset(struct switchtec_dev *dev)
...
+
+	return 0;
+
+err_free_log_defs:
+	free_log_defs(defs);
+	return -1;
+}
+
+/* maximum firmware module character length to display */
+#define FW_MOD_NAME_LEN_MAX	13
```
lsgunth, line 837: This appears to be unused now...

brian-mckenzie: Yes it is -- I'll remove it.

brian-mckenzie:
> Looks good to me. Thanks for your work on the improvements.

Great, thanks for reviewing.

lib/switchtec.c
```diff
@@ -777,6 +1102,48 @@ int switchtec_log_to_file(struct switchtec_dev *dev,
...
+		.num_alloc = 0};
+	int entry_idx = 0;
+
+	/* read the log definition file into defs */
+	ret = read_log_defs(log_def_file, &defs);
+	if (ret < 0)
+		return ret;
+
+	/* parse each app log entry */
+	while (fread(&log_data, sizeof(struct log_a_data), 1, bin_log_file) == 1) {
```
xyan264, line 1128: wrap this line.

PR #200, opened 2020-04-24
Fix 'with-ssl' naming issue in configure script

xyan264:
Also move 'pem.h' header file checking into 'with-openssl' condition.

### DIFF
--- a/configure
+++ b/configure
@@ -3528,7 +3528,9 @@ fi
 
 done
 
-for ac_header in openssl/pem.h
+
+if test "x$with_openssl" != xno; then :
+  for ac_header in openssl/pem.h
 do :
   ac_fn_c_check_header_mongrel "$LINENO" "openssl/pem.h" "ac_cv_header_openssl_pem_h" "$ac_includes_default"
 if test "x$ac_cv_header_openssl_pem_h" = xyes; then :
@@ -3540,9 +3542,7 @@ fi
 
 done
 
-
-if test "x$with_ssl" != xno; then :
-  { $as_echo "$as_me:${as_lineno-$LINENO}: checking for RSA_free in -lcrypto" >&5
+        { $as_echo "$as_me:${as_lineno-$LINENO}: checking for RSA_free in -lcrypto" >&5
 $as_echo_n "checking for RSA_free in -lcrypto... " >&6; }
 if ${ac_cv_lib_crypto_RSA_free+:} false; then :
   $as_echo_n "(cached) " >&6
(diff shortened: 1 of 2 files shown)


xyan264:
Commit message is updated to describe the issue: '--with-openssl=no' did not work. The issue was caused by different names -- argument defined as 'with_openssl' while testing for 'with_ssl'. 

I also took the opportunity to move the openssl header file checking into the condition clause, and combined the two testing into one single clause. The code looks cleaner this way. 

I put these in one commit since they both work on the same `with_openssl` test clause, and the number of lines are manageable. This way we can update the `configure` file once. Otherwise we will have two commits, each modifying the `configure` script. 

I can split the commit if you prefer.

configure
```diff
@@ -3528,7 +3528,9 @@ fi
 
 done
 
-for ac_header in openssl/pem.h
+
+if test "x$with_openssl" != xno; then :
+  for ac_header in openssl/pem.h
```
kelvin-cao, line 3533: Intent need to be adjusted.

configure
```diff
@@ -3540,9 +3542,7 @@ fi
 
 done
 
-
-if test "x$with_ssl" != xno; then :
-  { $as_echo "$as_me:${as_lineno-$LINENO}: checking for RSA_free in -lcrypto" >&5
+        { $as_echo "$as_me:${as_lineno-$LINENO}: checking for RSA_free in -lcrypto" >&5
```
kelvin-cao, line 3545: Is this leading white spaces intended?

configure
```diff
@@ -3587,10 +3587,7 @@ _ACEOF
 
 fi
 
-fi
-
-if test "x$with_ssl" != xno; then :
-  ac_fn_c_check_decl "$LINENO" "RSA_get0_key" "ac_cv_have_decl_RSA_get0_key" "#include <openssl/pem.h>
+        ac_fn_c_check_decl "$LINENO" "RSA_get0_key" "ac_cv_have_decl_RSA_get0_key" "#include <openssl/pem.h>
```
kelvin-cao, line 3590: Same as above.

xyan264:
I think we only need to focus on `configure.ac` file since `configure` script is auto-generated from it.

The indentation in `configure` is not the best, but maybe we can live with it?

kelvin-cao:
> 
> 
> I think we only need to focus on `configure.ac` file since `configure` script is auto-generated from it.
> 
> The indentation in `configure` is not the best, but maybe we can live with it?

Ooh, I just realized it, sure.

PR #201, opened 2020-04-28
Fix "make install" on MSYS2

brian-mckenzie:
"make install" fails in an MSYS2 environment, complaining that it can't create the symlink "/usr/local/lib/libswitchtec.dll.a". This is because the source of the symlink doesn't exist, and since MSYS2 just creates copies instead of symlinks, the copy fails.

Additionally, the install/uninstall Makefile targets run ldconfig commands. $LDCONFIG isn't defined in MSYS2 as ldconfig isn't supported, so these commands fail.

The symlink creation is now only done in a non-MinGW environment, as are the ldconfig commands.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -57,6 +57,10 @@ STLIBNAME ?= libswitchtec.a
 MACHINE=$(shell $(CC) -dumpmachine)
 
 ifeq ($(findstring mingw,$(MACHINE)),mingw)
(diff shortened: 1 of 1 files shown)


brian-mckenzie:
> Looks good to me. Please make sure this change works on Linux too.

Yep, I tested it out on Linux as well.

PR #203, opened 2020-04-30
Add EP CSR read command

xyan264:
`ep-csr-read` command allows user to read the CSR space of an endpoint device. 

It works similar to GAS read, but requires an extra parameter of the endpoint ID (pdfid).


xyan264:
Closed due to Travis CI check failure. Will reopen later.

inc/switchtec/switchtec.h
```diff
@@ -63,6 +63,12 @@ struct switchtec_dev;
 #define SWITCHTEC_PAX_ID_MASK 0x1f
 #define SWITCHTEC_PAX_ID_LOCAL SWITCHTEC_PAX_ID_MASK
 
+#ifdef __CHECKER__
+#define __csr __attribute__((noderef, address_space(1)))
```
lsgunth, line 67: You have to choose a unique number for the address space. Otherwise sparse will allow gas pointers and csr pointers to be used together.

Though, I'm really not sure why you are using pointers or mmaps for this at all.

The gas was done that way because on non-i2c/uart architectures the BAR is actually mapped into memory and could actually be accessed directly in appropriate places.

Here you might as well just use an integer instead of a pointer, unless/until there's actually going to be a way to map it into memory.

xyan264: Good suggestion. I'm thinking about removing the mmap as well. Will update the code later.

PR #205, opened 2020-05-01
Add EP CSR read command

xyan264:
`ep-csr-read` command allows user to read the CSR space of an endpoint device.

It works similar to GAS read, but requires an extra parameter of the endpoint ID (pdfid).

-- Updated to without mmap version.


xyan264:
@lsgunth please review and comment. Thanks!

inc/switchtec/fabric.h
```diff
@@ -634,6 +634,17 @@ int switchtec_ep_tunnel_enable(struct switchtec_dev *dev, uint16_t pdfid);
...
+
+/********** EP RESOURCE MANAGEMENT *********/
+#define SWITCHTEC_EP_CSR_MAX_READ_LEN  4
+
+int switchtec_ep_csr_read8(struct switchtec_dev *dev, uint16_t pdfid,
+			   uint16_t addr, uint8_t *val);
+int switchtec_ep_csr_read16(struct switchtec_dev *dev, uint16_t pdfid,
+			    uint16_t addr, uint16_t *val);
+int switchtec_ep_csr_read32(struct switchtec_dev *dev, uint16_t pdfid,
+			    uint16_t addr, uint32_t *val);
```
lsgunth, line 646: I probably wouldn't use specific type sizes in function arguments... but seeing a precedent has already been set above, I guess it can be left.

PR #207, opened 2020-05-05
Add EP BAR read command

xyan264:
This command reads memory regions of an endpoint device.

### DIFF
--- a/cli/fabric.c
+++ b/cli/fabric.c
@@ -1703,6 +1703,118 @@ static int ep_csr_write(int argc, char **argv)
 	return ret;
 }
 
+#define CMD_DESC_EP_BAR_READ "read BAR of an EP"
+static int ep_bar_read(int argc, char **argv)
+{
+	unsigned long long val;
+	unsigned long addr;
+	unsigned bytes;
+	int i;
+	int ret = 0;
+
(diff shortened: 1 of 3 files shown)


cli/fabric.c
```diff
@@ -1703,6 +1703,117 @@ static int ep_csr_write(int argc, char **argv)
...
+		return -1;
+	}
+
+	addr = cfg.addr;
+	bytes = cfg.bytes;
+	for (i = 0; i < cfg.count; i++) {
+		addr = addr & ~(bytes - 1);
+		val = 0;
+		switch (bytes) {
+			case 1: ret = switchtec_ep_bar_read8(cfg.dev,
```
lsgunth, line 1784: Nit:

case statements should be aligned with the switch. And the command should be on the following line, unless it all fits on one line.

In ugly cases like this it's better to violate the align with open parenthesis rule and just indent by an extra tab or two:

```
		switch (bytes) {
		case 1: 
			ret = switchtec_ep_bar_read8(cfg.dev, cfg.pdfid, 
					cfg.bar, addr, (uint16_t)&val);
...
```

xyan264: Updated.

PR #209, opened 2020-05-07
Add 'device-manage' command

xyan264:
This command sends management command to an endpoint device.


cli/fabric.c
```diff
@@ -1435,6 +1435,105 @@ static int gfms_events(int argc, char **argv)
...
+	};
+
+	const struct argconfig_options opts[] = {
+		DEVICE_OPTION,
+		{"pdfid", 'f', "NUM", CFG_INT, &cfg.req.hdr.pdfid,
+		 required_argument, "Endpoint function's FID",
+		 .require_in_usage = 1},
+		{"cmd_data", 'c', "String", CFG_STRING, &cmd_string,
+		 required_argument, .require_in_usage = 1,
+		 .help= "Command raw data in dword, format example: \"0x040b0006 0x00000001\""},
```
lsgunth, line 1493: This is exceptionally gross asking users to pass random data on the command line. switchtec-user was written to get away from that nonsense we saw in its predecessor.

How about writing a proper interface?

xyan264: The data exchanged depends completely on the device, so there is no good way to further structure the data. 

How about something like:
```switchtec fabric device-manage --dw0 0x0000001 --dw1 0x000002 --dw2 0x00003 ...```
This allows user to specify by DWORD instead of a string.

lsgunth: Well what actually is the data.... There's no documentation, how are users supposed to figure this out?

xyan264: This command provides a way for user to talk to EP device _directly_. The data sent using this command will be passed to the device _as-is_.

Each device has it's own definition of management data format, which is completely device-specific. User needs to refer to the EP device manual for the data format and definitions. 

This command looks like a development-phase feature that engineers use, but the APPS team wants to keep it. Otherwise we would have dropped this command in final release.

We do need to print a warning message to make user aware that the command could cause possible issues with EP, if it is not used properly.

lsgunth: Well, I don't like it at all. I'd much prefer actual commands for various EP devices be coded up instead of making users do it manually with hex numbers. That's way too ugly, error prone and time wasting; and will just make users write even uglier scripts to deal with the mess. Better to do it right than provide a horrid interface for people to do it wrong. 

Remember, also, this is open source so if you provide examples for common EPs you can let users add others that do not exist by following your example.

xyan264:
@hachevin would you like to add your comments?

xyan264:
Based on the discussion above, I think we can discuss the details about specific EP command in Issue #223 , and merge the existing `device_manage` command. Let me know if you think otherwise. 

I will add more details in Issue #223 .

lsgunth:
I'd personally rather see proper use-case before actually merging anything. It's hard to review code without seeing how it needs to be used.

The suggestion that this is used for NVMe admin commands makes the command line interface even more egregious. The suggestion that users would need to program and then parse those manually using hex values on the command line sounds like a nightmare.

xyan264:
@lsgunth quoting from your previous comments, "I'd much prefer actual commands for various EP devices be coded up instead of making users do it manually with hex numbers", could you give some examples of the commands you'd like to see implemented in this PR?

lsgunth:
I can't give you that. Part of the reason doing things this way is bad is because I have no idea what the commands should be. We just have this amorphous interface that gives you no help in how to use it or what it is for. It asks you to put in some opaque hex numbers and  maybe it will do what you want if you happened to get the numbers right. It's awful.

xyan264:
@hachevin quoting your previous comments, "As a bare minimum, we should code an interface for NVMe 1.3 Admin commands (the primary use case for this command). More specifically, the command required to enable VFs."

By 'command to enable VFs', did you mean commands like the ones from `nvme-cli` that does EP configuration such as creating and attaching namespaces?

xyan264:
@lsgunth Vince suggested we keep this raw data command, which "provides an important starting point for customers developing new platforms without them having to code up support for the device(s) they're trying to manage."

While I'm also collecting ideas on the more user-friendly commands, I think it's safe to move that discussion to #223 , and merge this raw data command.

Schedule-wise, we are going to release a version of `nvme-cli` based on the API provided in this PR next week, that's another reason I'd like to merge this PR and move further discussions to #223.

Let me know your thoughts.

lsgunth:
I disagree. That's just asking for trouble. Without know how something is going to be used you can't know if it's designed correctly and once you introduce something like this people might rely on it or build nightmare structures over top of it; then you have to support that garbage forever.

"without them having to code up support for the device(s) they're trying to manage."

Is completely false. The user will still have to code up some horrible script to create these hex commands and parse the result. Better to give them a more sensible option than to encourage them to use a bad interface and pretend that it solves their problem.

xyan264:
@lsgunth I agree this is quite clumsy to use, it did gave me a lot of grief when I tested command.

However, the decision to keep this (though hard-to-use) command is a business one, beyond technical, so let's stay with Vince's suggestion. I will continue the discussion on the more user-friendly commands in #223 .

Thanks!

PR #210, opened 2020-05-08
Enable 64-bit BAR address read/write

xyan264:
Existing `CFG_LONG` type cannot handle 64-bit BAR address. 

Add `CFG_LONG_LONG` type to enable parsing 64-bit address used in command argument.

### DIFF
--- a/cli/argconfig.c
+++ b/cli/argconfig.c
@@ -470,6 +470,23 @@ static int cfg_long_suffix_handler(const char *optarg, void *value_addr,
 	return 0;
 }
 
+static int cfg_long_long_handler(const char *optarg, void *value_addr,
+				 const struct argconfig_options *opt)
+{
+	char *endptr;
+
+	*((unsigned long long *)value_addr) = strtoull(optarg, &endptr, 0);
+	if (errno || optarg == endptr) {
+		fprintf(stderr,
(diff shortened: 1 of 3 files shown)


kelvin-cao:
It seems CFG_SIZE can be used for this purpose, but the current cfg_size_handler() need some tweak to use strtoull() to replace strtol().

xyan264:
@kelvin-cao Here's what I found on google:
```size_t type is a base unsigned integer type .... On a 32-bit system size_t will take 32 bits, on a 64-bit one 64 bits.```

So it might not be suitable for this purpose.

lsgunth:
Why? Are you expanding the BAR to be greater than 4GB? That should be done with caution and only for very good reason because the larger the BAR, the more trouble bioses have with it.

kelvin-cao:
> 
> 
> @kelvin-cao Here's what I found on google:
> `size_t type is a base unsigned integer type .... On a 32-bit system size_t will take 32 bits, on a 64-bit one 64 bits.`
> 
> So it might not be suitable for this purpose.

I talked with F/W, it's possible for a 32-bit host functions as fabric manager to manipulate a 64-bit EP behind the fabric. So, we cannot use size_t for this operation.

PR #211, opened 2020-05-11
Add 'mfg debug-token' command

xyan264:
This command generates a token file which can be readily used by security management system to generate the signature file required to unlock chip debug feature.

Binary format is required so that the customer's Security Management system can directly generate signature file from the token file, simplifying the debug procedure.


xyan264:
Renamed to `debug-unlock-token`. Merged.

PR #212, opened 2020-05-12
Resolve issues when running switchtec on a big endian powerpc embedded device

cjengel:


kelvin-cao:
Hi Chris,

Could you change the target branch to Microsemi:devel, which is for accepting PR?

Thanks,
Kelvin

PR #215, opened 2020-05-20
Fix "fw-info" command for Gen3 devices on Windows

brian-mckenzie:
Running the "fw-info" command in Windows for a Gen3 device gives the error "A device attached to the system is not functioning." This is because the MRPC_GET_DEV_INFO command is being issued but the command isn't supported by Gen3 firmware. switchtec_get_device_info() attempts to account for this by checking for ERR_MPRC_UNSUPPORTED, but the MRPC return value isn't available when the MRPC status indicates failure.

This is fixed by checking for a Gen3 device in switchtec_get_device_info() rather than relying on the MRPC command being unsupported.

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -1393,11 +1393,7 @@ static int fw_info(int argc, char **argv)
 
 	argconfig_parse(argc, argv, CMD_DESC_FW_INFO, opts, &cfg, sizeof(cfg));
 
-	ret = switchtec_get_device_info(cfg.dev, &phase_id, NULL, NULL);
-	if (ret) {
-		switchtec_perror("print fw info");
-		return ret;
-	}
+	phase_id = switchtec_boot_phase(cfg.dev);
 	if (phase_id == SWITCHTEC_BOOT_PHASE_BL1) {
 		fprintf(stderr,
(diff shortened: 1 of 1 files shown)


lsgunth:
Well sounds like if the MRPC_GET_DEV_INFO indicates a failure you have to assume the hardware is Gen3. Seems unlikely that Gen4 hardware should fail this call for any reason.

brian-mckenzie:
> Well sounds like if the MRPC_GET_DEV_INFO indicates a failure you have to assume the hardware is Gen3. Seems unlikely that Gen4 hardware should fail this call for any reason.

The MRPC system is fairly complex and requires both the FW and driver to be working correctly. There are other reasons it could fail, so I don't think we want to rely on its failure as an indication of Gen3.

I had a conversation with Paul earlier and I think we have a solution (basically just removing the switchtec_get_device_info() call from set_gen_variant() and having set_gen_variant() make the MRPC_GET_DEV_INFO call directly). Paul is just thinking it over.

brian-mckenzie:
Paul came up with this much simpler change -- we don't need to call switchtec_get_device_info() from fw_info() at all to get the boot phase, which avoids the whole issue.

PR #217, opened 2020-05-26
Fix "fabric.c" Windows build warnings.

xyan264:


cli/fabric.c
```diff
@@ -36,6 +36,26 @@
...
+#define PFMTllx	"llx"
+#define PFMTllX	"llX"
+#define PFMTlld	"lld"
+#define PFMThhx	"hhx"
+#define PFMThhu	"hhu"
+#define PFMThx	"hx"
+#define PFMThhd	"hhd"
+#define PFMTlx	"lx"
+#else
+#define PFMTllx	PRIx64
```
brian-mckenzie, line 51: Aren't these PRI* macros supported on Linux as well? Can't you just use them directly in the printf calls, and you wouldn't need these PFMT defines?

cli/fabric.c
```diff
@@ -729,12 +749,12 @@ static int hvd_detail_body_print(struct switchtec_dev *dev,
 	char bdf_str1[32];
 	char bdf_str2[32];
 	char vep_type_str[32];
-	uint64_t enable_bitmap;
-	uint64_t bitmap;
+	unsigned long long enable_bitmap;
```
brian-mckenzie, line 752: Why were these changed from uint64_t? Isn't uint64_t preferable since it's a guaranteed 64 bits?

xyan264: `uint64_t` is fine for data type, but when we try to print on 64-bit Linux with "%llx", it expects 'unsigned long long', but `uint64_t` is defined as 'unsigned long', so we will get a mismatch warning.

On 32-bit system, uint64_t is defined as 'unsigned long long' and we won't have issue. 

So the best fix across these different architectures is to define the data type as 'unsigned long long', and print with "%llx", so the consistency is maintained.

brian-mckenzie: If you use the PRI* macros on Linux as well, I think you could keep it at uint64_t then just use PRIx64. This should be compatible across platforms.

xyan264: Updated based on your suggestion. Thanks.

cli/fabric.c
xyan264: It's easier to create our own defines here. Otherwise we have to change the data definition as well. We use a mixture of `uintXX_t` and `unsigned XXX` definitions in our code. Changing these can be a pain.

kelvin-cao:
We already have some effort in the portabilities in inc/switchtec/portable.h, could you expand them there? @xyan264

cli/fabric.c
```diff
@@ -46,14 +66,14 @@ enum {
 static int raw_print_hex(unsigned long long value,
 			 int offset, int bytes)
 {
-	printf("%06X - 0x%0*llX\n", offset, bytes * 2, value);
+	printf("%06X - 0x%0*" PFMTllX "\n", offset, bytes * 2, value);
```
lsgunth, line 69: You should probably tell us what the build warnings are... but this all looks really wrong....

'llx' should always be correct for any long long value.... PRIx64 should be used for uint64_t and 'value' is a long long...

xyan264:
Code updated to use matching format in printf for different data types:
- `uint8/16/32/64_t`: use `PRIu8/16/32/64` format 
- `long`: use `FMT_lx`
- `long long`: use `FMT_llx`

cli/fabric.c
xyan264: The warnings was due to difference in Windows, e.g.,:
- Format `%llx` is not recognized on Windows. On Windows it should be `I64x`. These are now replaced with `FMT_llx` macro, which expands to the right format for the platform.
- Format `%hhu` is not recognized on Windows. Replaced with `PRIu8` which expands to the right format.
...

lsgunth: According to this they should be recognized... 

https://docs.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=vs-2019

And when I try a simple program with x86_64-w64-mingw32-gcc it recognizes the %ll and %hh just fine...

The actual warning messages you are seeing should really be in the commit message even if this patch was obviously correct.

Also, if there were build warnings in the code (even for windows), Travis should have caught them and prevented the code from being merged. So are you trying to do the windows build on some ancient compiler or something like that? This info needs to be in the commit message.

xyan264: Our code is compiled with `-Wall` so I got these warnings:
```
cli/fabric.c: In function 'raw_print_hex':
cli/fabric.c:49:23: warning: unknown conversion type character 'l' in format [-Wformat=]
  printf("%06X - 0x%0*llX\n", offset, bytes * 2, value);
                       ^
cli/fabric.c:49:9: warning: too many arguments for format [-Wformat-extra-args]
  printf("%06X - 0x%0*llX\n", offset, bytes * 2, value);
...
```


I compiled the following code with `-Wall` using x86_64-w64-mingw32-gcc (ver. 8.3.0)and got the same warnings.
```
#include "stdio.h"
int main()
{
        unsigned long long x = 0x1234567898765432;
        printf("%llx\n", x);
        return 0;
}
```

The code does print correctly, so it's just the warning from the compiler that we need to worry about.

For Travis, the warnings are there in the logs, however, they are not considered as build errors. So in the summary we see `build pass`.

lsgunth: Ah I see...

How about instead of this mess we set __USE_MINGW_ANSI_STDIO per this [document][1]? I'd just set it on the command line for windows (ie -D__USE_MINGW_ANSI_STDIO).

Also, we should probably ensure Travis is running with -Werror so we fix warnings before they enter the source.

[1]: https://sourceforge.net/p/mingw-w64/wiki2/gnu%20printf/

xyan264: The macro -D__USE_MINGW_ANSI_STDIO does work for `fabric.c`, but `portable.h` has the following definitions:
```
#ifdef __WINDOWS__
#define FMT_SIZE_T_x  "Ix"
#define FMT_llX       "I64X"
#define FMT_llx       "I64x"
#define FMT_lld       "I64d"
```
So the files that uses these defines (gas.c, etc.) failed to build. 

There are 2 paths we can take from here:
- remove definitions from `portable.h` and fix `gas.c`, use __USE_MINGW_ANSI_STDIO for all files
- keep `portable.h` definitions and use current build fix

I think it's better to keep `portable.h` definitions so the code has better portability to other platform (with different definitions for `llx`).

What do you think? @lsgunth

lsgunth: I think it's better to remove the portable.h junk so the code is much cleaner.

There are no platforms that have different definitions of llx. Only ancient windows platforms have that issue and we can substitute in a good version for all cases with the flag above.

xyan264: Updated based on your suggestions. Please take a look. @lsgunth

cli/fabric.c
```diff
@@ -676,16 +676,16 @@ static int hvd_body_print(struct switchtec_dev *dev,
 		}
 
 		if (!bound) {
-			printf("        Logical PID %hhu:\t\tUnbound\n", i);
+			printf("        Logical PID %d:\t\tUnbound\n", i);
```
lsgunth, line 679: These changes from %hhu to %d looks like an unrelated bug fix.... It should be done in a different patch (same PR is fine).

cli/fabric.c
```diff
@@ -729,8 +729,8 @@ static int hvd_detail_body_print(struct switchtec_dev *dev,
 	char bdf_str1[32];
 	char bdf_str2[32];
 	char vep_type_str[32];
-	uint64_t enable_bitmap;
-	uint64_t bitmap;
+	unsigned long long enable_bitmap;
+	unsigned long long bitmap;
```
lsgunth, line 733: I kind of feel uint64_t is still the better choice here... You should be able to use the PRIu64 macro to obtain the correct format specifier. But this is also an unrelated bugfix for a third patch.

xyan264:
Updated with separate commits. @lsgunth

PR #218, opened 2020-05-27
Add additional checking for PAX ID.

xyan264:
Previous checking is insufficient in handling negative numbers.

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -112,11 +112,15 @@ int mfg_handler(const char *optarg, void *value_addr,
 int pax_handler(const char *optarg, void *value_addr,
 		const struct argconfig_options *opt)
 {
-	int ret;
+	char *end;
+	long num;
 
-	ret = sscanf(optarg, "%i", &global_pax_id);
+	errno = 0;
+	num = strtol(optarg, &end, 0);
+	global_pax_id = num;
 
-	if (ret != 1 || (global_pax_id & ~SWITCHTEC_PAX_ID_MASK)) {
+	if ((end == optarg) || errno || num < 0 ||
+	    (global_pax_id & ~SWITCHTEC_PAX_ID_MASK)) {
 		fprintf(stderr, "Invalid PAX ID specified: %s\n", optarg);
 		return 1;
 	}


cli/main.c
```diff
@@ -112,11 +112,14 @@ int mfg_handler(const char *optarg, void *value_addr,
...
-	int ret;
+	char *end;
+	long num;
 
-	ret = sscanf(optarg, "%i", &global_pax_id);
+	num = strtol(optarg, &end, 0);
+	global_pax_id = num;
 
-	if (ret != 1 || (global_pax_id & ~SWITCHTEC_PAX_ID_MASK)) {
+	if ((end == optarg) || errno || num < 0 ||
```
lsgunth, line 121: It's a bad idea to look at errno here unless strtol returns LONG_MIN or LONG_MAX... anything could have changed errno before it reached here. 

Though it doesn't seem like a huge improvement switching from sscanf to strtol: the error checking is more complicated and I'm not sure it improves anything besides adding the negative check. You could also consider using `sscanf(optarg, "%i%n"... `  to check if you've matched to the end.

xyan264: Updated with clearing `errno` before call to `strol`.

For this fix, besides the negative check, the `strtol` also does range check so the out-of-range PAX IDs like `-0xfffffffffffffffff` can be caught as well.

lsgunth: No clearing errno is equally wrong.

You have to make sure to only look at errno if the function that was just called returns an error (ie if LONG_MIN or LONG_MAX). Otherwise it is not correct.

Weird that sscanf doesn't do something sensible for very large values. Pretty corner case though, I probably wouldn't have woried about it.

xyan264: Clearing `errno` before calling `strtol` is a recommended practice, based on Linux manual page https://linux.die.net/man/3/strtol
```Notes

Since strtol() can legitimately return 0, LONG_MAX, or LONG_MIN (LLONG_MAX or LLONG_MIN for strtoll()) on both success and failure, the calling program should set errno to 0 before the call, and then determine if an error occurred by checking whether errno has a nonzero value after the call. ```

Yes this is a pretty corner case, but it was caught by our PV team, so it is address in this PR.

PR #219, opened 2020-06-02
Fix GAS dump when some regions are inaccessible

xyan264:
Due to security considerations, access to some GAS regions through MRPC command are blocked. This affects 'gas dump' command which reads through the whole region.

Previous `memcpy_from_gas` function raises signal when gas read fails, causing program to exit when some regions are inaccessible. We define new `no_fault` version of the function that returns an error code. CLI checks the error code and print `XX` for those inaccessible memory regions.

### DIFF
--- a/cli/gas.c
+++ b/cli/gas.c
@@ -325,18 +325,27 @@ static int read_gas(struct switchtec_dev *dev, void __gas *addr,
 {
 	int ret = 0;
 
+	uint8_t u8;
+	uint16_t u16;
+	uint32_t u32;
+	uint64_t u64;
+
 	switch (bytes) {
 	case 1:
-		*val = gas_read8(dev, addr);
+		ret = gas_read8(dev, addr, &u8);
+		*val = u8;
 		break;
 	case 2:
-		*val = gas_read16(dev, addr);
+		ret = gas_read16(dev, addr, &u16);
+		*val = u16;
 		break;
 	case 4:
-		*val = gas_read32(dev, addr);
+		ret = gas_read32(dev, addr, &u32);
+		*val = u32;
 		break;
 	case 8:
-		*val = gas_read64(dev, addr);
+		ret = gas_read64(dev, addr, &u64);
+		*val = u64;
 		break;
 	default:
 		errno = EINVAL;
--- a/cli/gas.c
+++ b/cli/gas.c
@@ -71,13 +71,18 @@ static void hexdump_data(struct switchtec_dev *dev, void __gas *map,
 	unsigned long addr = 0;
 	size_t bytes;
 	int last_match = 0;
+	int err;
 
 	while (map_size) {
 		if (is_alive && !is_alive())
 			return;
 
 		bytes = map_size > sizeof(line) ? sizeof(line) : map_size;
-		memcpy_from_gas(dev, line, map, bytes);
+		err = memcpy_from_gas(dev, line, map, bytes);
+		if (err && errno == EPERM) {
+			fprintf(stderr, "GAS dump: permission denied\n");
+			return;
+		}
 
 		if (bytes != sizeof(line) ||
 		    memcmp(last_line, line, sizeof(last_line))) {
@@ -405,9 +410,15 @@ static int print_str(struct switchtec_dev *dev, void __gas *addr,
 		     int offset, int bytes)
 {
 	char buf[bytes + 1];
+	int ret;
 
 	memset(buf, 0, bytes + 1);
-	memcpy_from_gas(dev, buf, addr + offset, bytes);
+
+	ret = memcpy_from_gas(dev, buf, addr + offset, bytes);
+	if (ret) {
+		switchtec_perror("gas read");
+		return -1;
+	}
 
 	printf("%06X - %s\n", offset, buf);
 	return 0;
--- a/cli/gas.c
+++ b/cli/gas.c
@@ -320,20 +320,48 @@ static int gas_dump(int argc, char **argv)
 	return pipe_to_hd_less(cfg.dev, map, cfg.count);
 }
 
+static int read_gas(struct switchtec_dev *dev, void __gas *addr,
+		    int bytes, unsigned long long *val)
+{
+	int ret = 0;
+
+	switch (bytes) {
+	case 1:
+		*val = gas_read8(dev, addr);
+		break;
+	case 2:
+		*val = gas_read16(dev, addr);
+		break;
+	case 4:
+		*val = gas_read32(dev, addr);
+		break;
+	case 8:
+		*val = gas_read64(dev, addr);
+		break;
+	default:
+		errno = EINVAL;
+		return -1;
+	}
+
+	return ret;
+}
+
 static int print_hex(struct switchtec_dev *dev, void __gas *addr,
 		     int offset, int bytes)
 {
 	unsigned long long x;
+	int ret;
 
 	offset = offset & ~(bytes - 1);
 
-	switch (bytes) {
-	case 1: x = gas_read8(dev, addr + offset);  break;
-	case 2: x = gas_read16(dev, addr + offset); break;
-	case 4: x = gas_read32(dev, addr + offset); break;
-	case 8: x = gas_read64(dev, addr + offset); break;
-	default:
+	if (bytes != 1 && bytes != 2 && bytes != 4 && bytes != 8) {
 		fprintf(stderr, "invalid access width\n");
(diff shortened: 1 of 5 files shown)


xyan264:
Switched to signal-based version. @lsgunth  please take a look.

cli/gas.c
```diff
@@ -63,6 +72,13 @@ static void print_line(unsigned long addr, uint8_t *bytes, size_t n)
 	printf("|\n");
 }
 
+static void gas_dump_sig_handler(int signum)
+{
+	if (signum == SIGBUS) {
```
lsgunth, line 77: This if shouldn't be necessary given this handler is only installed for SIGBUS.

xyan264:
Most of the memory regions should be accessible, so it is not typical for a GAS READ to fail. For the failed GAS READ, we can use this rule: if a chunk is inaccessible, we raise signal for the whole GAS READ call (as in the updated code)
```
		ret = switchtec_cmd(dev, MRPC_GAS_READ, &cmd,
				    sizeof(cmd), dest, cmd.len);
		if (ret) {
			memset(dest, 0xff, n);
			raise(SIGBUS);
			return;
		}
```
This can simplify error handling at this level, and it is up to the upper layer to decide the appropriate read size per GAS READ.

lsgunth:
Ok, yes, but one common reason for that MRPC to fail is the user doesn't have appropriate permissions (we added a check for that in switchtec-kernel). In that case no  read is likely to succeed and a gas dump command should just fail with an appropriate message instead of printing a whole bunch of Xs...

xyan264:
GAS memory space is divided into different regions, some of these regions have their own permission for MRPC access. The permission table is stored on the device side, so we cannot do a permission check first, as done in kernel driver. 

From what I understand (and saw during testing), most of the GAS regions are accessible, only a small portion of the GAS space has permissions to protect them from GAS READ MRPC. So user will see meaningful data, with a few XX here and there, when using `gas dump` command.

lsgunth:
That's not my point. My point is if the mrpc command returns -EPERM or some such (because the kernel rejected it) we shouldn't just print out all the XXs.... instead print an informative error message.

xyan264:
Let me understand... you prefer output like this:

```
0000c400  e3 a8 01 00 48 89 84 24  08 01 00 00 e8 ff 44 00  |....H..$......D.|
0000c410  00 48 8b 7c 24 40 48 8d  74 24 20 e8 70 15 01 00  |.H.|$@H.t$ .p...|
0000c420  85 c0 75 7c 8b 44 24 20  8b 7c 24 48 48 8d 74 24  |..u|.D$ .|$HH.t$|

0000c430 - 0000c530  REGION INACCESSIBLE

0000c540  16 02 00 00 48 8d 3d 38  a8 01 00 e8 b0 5f ff ff  |....H.=8....._..|

...```

**Comment by lsgunth** [comment]
No that's not at all what I'm saying. I'm saying if the kernel rejects it with -EPERM  then NONE of the memory will be accessible and we shouldn't just dump it all as inaccessible; we should print a single sensible error message.

**Comment by xyan264** [comment]
I see. The -EPERM is a special case where no GAS region is accessible. 

I have added handling of this case by raising a different signal, which is handled at CLI level by printing an error message and exit. 

Please review and see if it's good enough.

File: lib/gas_mrpc.c [hunk 2]
```diff
@@ -97,14 +100,21 @@ void gas_mrpc_memcpy_from_gas(struct switchtec_dev *dev, void *dest,
 
 	while (n) {
 		cmd.len = n;
-		if (cmd.len > MRPC_MAX_DATA_LEN)
-			cmd.len = MRPC_MAX_DATA_LEN;
+		if (cmd.len > MRPC_GAS_MEMCPY_MAX_SIZE)
+			cmd.len = MRPC_GAS_MEMCPY_MAX_SIZE;
```
lsgunth, line 104: This should really be a second patch... If you write "also" or "and" or whatever, it probably means it's a completely different topic.

cli/gas.c
```diff
@@ -63,6 +72,17 @@ static void print_line(unsigned long addr, uint8_t *bytes, size_t n)
 	printf("|\n");
 }
 
+static void gas_dump_sig_handler(int signum)
+{
+	if (signum == SIGTERM) {
```
lsgunth, line 77: No, this is not a good idea. If someone kills the process at just the wrong time they will get this odd error message.

lsgunth:
Actually, sorry, I think I might have been wrong in my initial review. The Linux and Windows platform code will not normally generate SIGBUS's we just added that because most gas_reads()/gas_writes() have no way to handle an error return.

So maybe go back to the original approach of adding proper error returns to the memcpy_from_gas() and similar functions. We don't need a _nofault version, we just need to ensure all existing callers check for an error. In which case it might be worth adding an `__attribute__((warn_unused_result))`

inc/switchtec/gas_mrpc.h
```diff
@@ -43,8 +43,8 @@ struct gas_mrpc_read {
 
 void gas_mrpc_memcpy_to_gas(struct switchtec_dev *dev, void __gas *dest,
 			    const void *src, size_t n);
-void gas_mrpc_memcpy_from_gas(struct switchtec_dev *dev, void *dest,
-			      const void __gas *src, size_t n);
+int gas_mrpc_memcpy_from_gas(struct switchtec_dev *dev, void *dest,
```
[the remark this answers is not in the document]
xyan264: gas_mrpc_memcpy_from_gas() is the underlying function for gas(_mrpc)_read8/16/32/64 so it belongs to this commit.

cli/gas.c
```diff
@@ -345,19 +384,21 @@ static int print_dec(struct switchtec_dev *dev, void __gas *addr,
...
-	case 8: x = gas_read64(dev, addr + offset); break;
-	default:
+	if (bytes != 1 && bytes != 2 && bytes != 4 && bytes != 8) {
 		fprintf(stderr, "invalid access width\n");
 		return -1;
 	}
 
+	ret = read_gas(dev, addr + offset, bytes, &x);
+	if (ret) {
+		fprintf(stderr, "Error reading from GAS\n");
```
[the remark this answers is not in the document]
xyan264: Changed to switchtec_perror() so it prints an error code for tracing.

xyan264:
Updated with cleaner commits. @lsgunth please take a look.

To simply error handling, it returns error if any region of the memory requested is not readable. It is up to the caller to find the right size for each GAS read.

inc/switchtec/gas_mrpc.h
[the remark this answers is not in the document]
lsgunth: No... just do it first.... and let the callers ignore the return code for a patch or two...

I did like your cleanup which added the gas_read_access() helper in the gas cli code... this patch would also be easier to review if you still did that, but ahead of this patch.

xyan264:
@lsgunth thanks for the suggestions. Code is now updated. Please take a look.

inc/switchtec/gas.h
```diff
@@ -67,8 +67,8 @@
 void memcpy_to_gas(struct switchtec_dev *dev, void __gas *dest,
 		   const void *src, size_t n);
 
-void memcpy_from_gas(struct switchtec_dev *dev, void *dest,
-		     const void __gas *src, size_t n);
+int memcpy_from_gas(struct switchtec_dev *dev, void *dest,
```
lsgunth, line 70: Should we not have a warn_unused_result for this? Seems like it would be a good idea seeing people might skip it given that it is a memcpy like interface?

cli/gas.c
```diff
@@ -71,17 +77,18 @@ static void hexdump_data(struct switchtec_dev *dev, void __gas *map,
...
 	int last_match = 0;
+	int err;
 
 	while (map_size) {
 		if (is_alive && !is_alive())
 			return;
 
 		bytes = map_size > sizeof(line) ? sizeof(line) : map_size;
-		memcpy_from_gas(dev, line, map, bytes);
+		err = memcpy_from_gas(dev, line, map, bytes);
```
lsgunth, line 87: This is where we should have `if (errno == EPERM)` print an error and exit.

And any other error that indicates no GAS can be read.

xyan264: Updated below.

inc/switchtec/gas.h
xyan264: Updated above.

xyan264:
Updated. @lsgunth please take a look.

PR #225, opened 2020-06-16
Add mailbox log parsing support

brian-mckenzie:
Add mailbox log parsing support to the Switchtec library file. Mailbox logs are different from app logs as follows:
-the log definition file doesn't divide the definitions into modules. All definitions are in a single list.
-log severities are always the same
-bit 27 of the 3rd DWord of a log entry indicates the log source (BL1/BL2)

The "log-parse" command now takes a parameter to specify the log type (APP or MAILBOX) so that mailbox logs can be parsed.

### DIFF
--- a/cli/argconfig.c
+++ b/cli/argconfig.c
@@ -470,6 +470,23 @@ static int cfg_long_suffix_handler(const char *optarg, void *value_addr,
 	return 0;
 }
 
+static int cfg_long_long_handler(const char *optarg, void *value_addr,
+				 const struct argconfig_options *opt)
+{
+	char *endptr;
+
+	*((unsigned long long *)value_addr) = strtoull(optarg, &endptr, 0);
+	if (errno || optarg == endptr) {
+		fprintf(stderr,
+			"Expected long long integer argument for '--%s/-%c' "
+			"but got '%s'!\n",
+			opt->option, opt->short_option, optarg);
+		return 1;
+	}
+
+	return 0;
+}
+
 static int cfg_double_handler(const char *optarg, void *value_addr,
 			      const struct argconfig_options *opt)
 {
@@ -758,6 +775,7 @@ static type_handler cfg_type_handlers[_CFG_MAX_TYPES] = {
 	[CFG_SIZE_SUFFIX] = cfg_size_suffix_handler,
 	[CFG_LONG] = cfg_long_handler,
 	[CFG_LONG_SUFFIX] = cfg_long_suffix_handler,
+	[CFG_LONG_LONG] = cfg_long_long_handler,
 	[CFG_DOUBLE] = cfg_double_handler,
 	[CFG_BOOL] = cfg_bool_handler,
 	[CFG_BYTE] = cfg_byte_handler,
--- a/cli/main.c
+++ b/cli/main.c
@@ -944,13 +944,20 @@ static int log_dump(int argc, char **argv)
 	return ret;
 }
 
-#define CMD_DESC_LOG_PARSE "parse a binary app log to a text file"
+#define CMD_DESC_LOG_PARSE "parse a binary app log or mailbox log to a text file"
 
 static int log_parse(int argc, char **argv)
 {
 	int ret;
 
+	const struct argconfig_choice log_types[] = {
+		{"APP", SWITCHTEC_LOG_PARSE_TYPE_APP, "app log"},
+		{"MAILBOX", SWITCHTEC_LOG_PARSE_TYPE_MAILBOX, "mailbox log"},
+		{}
+	};
+
 	static struct {
+		enum switchtec_log_parse_type log_type;
 		FILE *bin_log_file;
 		const char *bin_log_filename;
 		FILE *log_def_file;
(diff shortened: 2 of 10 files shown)


lib/switchtec.c
```diff
@@ -774,7 +768,7 @@ static int read_log_defs(FILE *log_def_file, struct log_defs *defs)
 		tok = strtok(NULL, " \t");
 		if (!tok)
 			continue;
-        	mod_id = strtol(tok, NULL, 0);
+		mod_id = strtol(tok, NULL, 0);
```
lsgunth, line 771: Always do whitespace cleanup in another patch.

lib/switchtec.c
```diff
@@ -789,40 +783,121 @@ static int read_log_defs(FILE *log_def_file, struct log_defs *defs)
...
+ * @return 0 on success, negative value on failure
+ */
+static int read_log_defs(FILE *log_def_file,
+			 enum switchtec_log_parse_type log_type,
+			 struct log_defs *defs)
+{
+	int ret;
+
+	/* allocate some log definition entries */
+	ret = realloc_log_defs(defs, 200);
```
lsgunth, line 888: This is confusing me.... read_mailbox_log_defs() never seems to call realloc_log_defs() so presumably it's only actually using one module... But we always allocate 200?

I don't really see any value in this static function, it's probably better to just put the allocation into each of read_app_log_defs() and read_mailbox_log_defs() respectively... they clearly need different things.

lib/switchtec.c
```diff
@@ -880,26 +964,45 @@ static int write_parsed_log(struct log_a_data log_data[],
...
-
-		/* validate the module ID */
-		if ((mod_id > defs->num_alloc) ||
-		    (strlen(defs->module_defs[mod_id].mod_name) == 0)) {
-			if (fprintf(log_file, "(Invalid module ID: 0x%x)\n",
-				    mod_id) < 0)
-				goto ret_print_error;
-			continue;
+		if (log_type == SWITCHTEC_LOG_PARSE_TYPE_APP) {
+			/*
```
lsgunth, line 968: So why is the functionality changing in this branch? That's not expected. If you want to change the functionality of APP logs, it should be done first in another patch with a commit message that tells me why we are making that change.

lib/switchtec.c
```diff
@@ -880,26 +964,45 @@ static int write_parsed_log(struct log_a_data log_data[],
...
+			 * app log: module ID and log severity are in the 3rd
+			 * DWord
+			 */
+			mod_id = (log_data[i].data[2] >> 16) & 0xFFF;
+			log_sev = (log_data[i].data[2] >> 28) & 0xF;		
+
+			if ((mod_id > defs->num_alloc) ||
+			    (defs->module_defs[mod_id].mod_name == NULL) ||
```
lsgunth, line 976: So why is the functionality changing here? That's not expected. If you want to change the functionality of APP logs, it should be done first in another patch with a commit message that tells me why we are making that change.

lib/switchtec.c
```diff
@@ -960,7 +1068,7 @@ static int log_a_to_file(struct switchtec_dev *dev, int sub_cmd_id,
 
 	if (log_def_file != NULL) {
 		/* read the log definition file into defs */
-		ret = read_log_defs(log_def_file, &defs);
+		ret = read_log_defs(log_def_file, true, &defs);
```
lsgunth, line 1071: Not withstanding that I think we should drop this function; but using "true" here is very confusing.

lib/switchtec.c
```diff
@@ -986,7 +1094,8 @@ static int log_a_to_file(struct switchtec_dev *dev, int sub_cmd_id,
 
 			/* parse the log data and write it to a file */
 			ret = write_parsed_log(res.data, res.hdr.count,
-					       entry_idx, &defs, log_file);
+					       entry_idx, &defs, true,
```
lsgunth, line 1097: Don't use true/false when the argument takes an enum. Use actual enum values.

cli/main.c
```diff
@@ -944,29 +944,39 @@ static int log_dump(int argc, char **argv)
...
+		{"MAILBOX", SWITCHTEC_LOG_PARSE_TYPE_MAILBOX, "mailbox log"},
+		{}
+	};
+
 	static struct {
-		struct switchtec_dev *dev;
-		int out_fd;
-		const char *out_filename;
-		unsigned type;
+		enum switchtec_log_parse_type log_type;
```
lsgunth, line 960: Cleanup should happen in it's own commit. It's confusing, but I assume the fields being removed were never used in the first place?

brian-mckenzie: Right, they were never used. I'll remove them in a separate commit.

lib/switchtec.c
brian-mckenzie: Sure.

lib/switchtec.c
brian-mckenzie: Sure. I'll remove this function and move the allocation into those ones.

lib/switchtec.c
brian-mckenzie: The functionality isn't changing for app logs. The parsing is just reorganized since some of it doesn't apply to mailbox logs.

lib/switchtec.c
brian-mckenzie: Same as above.

lib/switchtec.c
brian-mckenzie: Right -- originally the parameter was a bool, but I updated it to the enum and forgot to change this.

lib/switchtec.c
brian-mckenzie: Same as the above -- I forgot to update it to the enum.

cli/argconfig.c
```diff
@@ -470,6 +470,23 @@ static int cfg_long_suffix_handler(const char *optarg, void *value_addr,
 	return 0;
 }
 
+static int cfg_long_long_handler(const char *optarg, void *value_addr,
```
brian-mckenzie, line 473: Not sure why these unrelated commits showed up -- I'll resolve this.

brian-mckenzie: I'm cancelling this PR. Please use this one instead: https://github.com/Microsemi/switchtec-user/pull/226

PR #226, opened 2020-06-16
Add mailbox log parsing support

brian-mckenzie:
Convert spaces to tabs for a few lines in switchtec.c

---

Add mailbox log parsing support to the Switchtec library file. Mailbox logs are different from app logs as follows:
-the log definition file doesn't divide the definitions into modules. All definitions are in a single list.
-log severities are always the same
-bit 27 of the 3rd DWord of a log entry indicates the log source (BL1/BL2)

---

The "cfg" struct in main.c::log_parse() has some unused elements that should be removed.

---

The "log-parse" command now takes a parameter to specify the log type (APP or MAILBOX) so that mailbox logs can be parsed.


lib/switchtec.c
```diff
@@ -880,26 +950,45 @@ static int write_parsed_log(struct log_a_data log_data[],
...
+		if (log_type == SWITCHTEC_LOG_PARSE_TYPE_APP) {
+			/*
+			 * app log: module ID and log severity are in the 3rd
+			 * DWord
+			 */
+			mod_id = (log_data[i].data[2] >> 16) & 0xFFF;
+			log_sev = (log_data[i].data[2] >> 28) & 0xF;		
+
+			if ((mod_id > defs->num_alloc) ||
+			    (defs->module_defs[mod_id].mod_name == NULL) ||
```
lsgunth, line 962: I still don't understand why this line is being snuck into this patch...

brian-mckenzie: Ah, right. I was trying out passing mailbox files but specifying the type as app log, which was leading to a segfault. I added this line as a test, which fixed it, but it's a bit of a hack and should probably be removed.

I'm not sure if we want to handle to case of the user specifying the wrong log type (app / mailbox) for the input files. If we do it would probably require some more advanced parsing of the log definition file.

lsgunth: Ah, well you can add that line but do it in a different patch with justification.

But clearly there is a bigger problem, if code can get this far when you pass it an invalid data file and that needs to be addressed properly.

brian-mckenzie: Sure, I'll move this addition to a new patch.

I agree that it would be better to validate the data earlier. That's why I originally preferred parsing the definition file using something like regular expressions or sscanf, as we can at least check for the proper data types (e.g. the module description is a string, a 4-digit hex number, and a decimal number). We can deal with this in a later PR.

lsgunth: I don't see why sscanf or re would be any different. You can get the same error checking, it just wasn't done.

Seems like we are missing some error checking on the strtol and it might make sense to error out if strtok doesn't find the requisite number of fields instead of ignoring the line.

The mailbox file looks a bit worse as it seems to just store every line regardless of what's in the file. Can we maybe at least at a header to that file so this code can at least verify something?

brian-mckenzie: I'll update the strtol error checking, which should catch the unexpected values. Lines are ignored if the strtok fails as some lines are blank. I can instead explicitly skip blank lines and error out if strtok fails.

The mailbox definition file is just a raw list of error strings, which makes it difficult to validate. We could potentially add a header, but I don't think we'd want to rely on that in case users pass in earlier versions that don't have the header.

lsgunth: Maybe add a force option to accept old mailbox files that don't have a header?

brian-mckenzie: I've updated the strol() error checking. The mod_name NULL check is no longer needed.

We can add headers to the definition files, but this will need to be approved and implemented on the firmware side first. If we're going ahead with that then I can implement the header checking in a separate PR.

lib/switchtec.c
```diff
@@ -736,6 +738,24 @@ static int realloc_log_defs(struct log_defs *defs, int num_modules)
...
+ * @brief Parse an integer from a string
+ * @param[in] str  - string to parse
+ * @param[out] val - integer
+ * @return true on success, false on failure
+ */
+static bool parse_int(char *str, int *val)
+{
+	char *endptr;
+
+	*val = strtol(str, &endptr, 0);
```
lsgunth, line 751: In other recent work we were setting errno=0 and then checking that errno is set after calling strtol.... We should probably do the same here.n

brian-mckenzie: Sure, I'll add that.

PR #227, opened 2020-06-22
Fix Travis build warnings

xyan264:
Fix Travis build warnings and enable `-Werror` in Travis yml file.

### DIFF
--- a/lib/switchtec.c
+++ b/lib/switchtec.c
@@ -949,7 +949,7 @@ static int write_parsed_log(struct log_a_data log_data[],
 	unsigned int nanos, micros, millis, secs, mins, hours, days;
 	unsigned int entry_num;
 	unsigned int mod_id;
-	unsigned int log_sev;
+	unsigned int log_sev = 0;
 	const char *log_sev_strs[] = {"DISABLED", "HIGHEST", "HIGH", "MEDIUM",
 				      "LOW", "LOWEST"};
 	bool is_bl1;
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,14 +7,18 @@ sudo: required
 matrix:
   include:
     - os: linux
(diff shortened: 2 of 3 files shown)


lib/switchtec.c
```diff
@@ -949,7 +949,7 @@ static int write_parsed_log(struct log_a_data log_data[],
 	unsigned int nanos, micros, millis, secs, mins, hours, days;
 	unsigned int entry_num;
 	unsigned int mod_id;
-	unsigned int log_sev;
+	unsigned int log_sev = 0;
```
kelvin-cao, line 952: I think this warning is valid, in the if (log_type == SWITCHTEC_LOG_PARSE_TYPE_APP) branch. This change makes 0 as a default value for log_sev, does it make sense?

xyan264: @brian-mckenzie could you comment here?

lsgunth: Looks like the warning is bogus to me.... log_sev is always set if log_type == SWITCHTEC_LOG_PARSE_TYPE_APP and only read if log_type == SWITCHTEC_LOG_PARSE_TYPE_APP.

But please ensure that you put the warning/error you are fixing in the commit message when you fix this type of bug.

brian-mckenzie: I don't really think the warning is valid. log_sev is only set in a (log_type == SWITCHTEC_LOG_PARSE_TYPE_APP) branch, but it's also only used based on that same condition. So if it's used, it's guaranteed to be set.
My version of GCC (7.5.0) doesn't report it as a warning, but older GCC versions do. I think it's fine to initialize it to 0 though to satisfy all versions.

xyan264:
Commit logs updated to include warning message.

PR #231, opened 2020-07-17
Fix PCIe address parsing in switchtec_open()

xyan264:
Insufficient checking in `switchtec_open()` causes PCIe address being parsed as Ethernet address.

Previously `0000:17:00.1` was parsed as Ethernet address `0000:17`, ignoring `00.1`

### DIFF
--- a/lib/switchtec.c
+++ b/lib/switchtec.c
@@ -212,8 +212,8 @@ struct switchtec_dev *switchtec_open(const char *device)
 	char *endptr;
 	struct switchtec_dev *ret;
 
-	if (sscanf(device, "%2049[^:]:%i", path, &inst) == 2) {
-		ret = switchtec_open_eth(path, inst);
+	if (sscanf(device, "%i@%i", &bus, &dev) == 2) {
+		ret = switchtec_open_i2c_by_adapter(bus, dev);
 		goto found;
 	}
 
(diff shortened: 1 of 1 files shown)


lib/switchtec.c
```diff
@@ -212,7 +212,8 @@ struct switchtec_dev *switchtec_open(const char *device)
 	char *endptr;
 	struct switchtec_dev *ret;
 
-	if (sscanf(device, "%2049[^:]:%i", path, &inst) == 2) {
+	if (sscanf(device, "%2049[^:]:%i%n", path, &inst, &idx) == 2 &&
+	    device[idx]=='\0') {
```
kelvin-cao, line 216: It's too tricky I guess. Can we use a format like what we do for pci address at line 242?

xyan264: Then we need to change the `switchtec_open_eth(path, inst)` function format. Do you want to do this?

lsgunth: Oh, yeah, this doesn't look spectacular... Why are we trying to match one of the most general methods first? It looks like it also breaks open_i2c with a colon below as well.

We should either try this last and allow any hostname  even without the instance number (which looks oddly conflated with a port number) or we should make the ethernet option require a better indication that the user really means to connect via ethernet, like using a URL. This seems like the best choice because the user should really be able to optionally specify the port. 

"switchtec://hostname[:port]/[inst]" would make a bit more sense for ethernet.

xyan264:
Closed due to internals issue. Will continue the discussion in another PR.

PR #232, opened 2020-07-23
Fix switchtec_open device name parsing issues

xyan264:
Change parsing pattern order to fix PCIe and I2C address parsing issues.

All cases have been verified to work now.

------------------------

I have moved the Ethernet address parsing to the end. Also found I2C parsing didn't work, fixed here as well. I think it's OK to combine the 2 fixes in on PR since they all fix the same 'device name parsing' issue. Let me know if otherwise. 

The team has decided not to adopt the more formal `switchtec://hostname:port` proposal, due to compatibility considerations.

### DIFF
--- a/lib/switchtec.c
+++ b/lib/switchtec.c
@@ -212,11 +212,6 @@ struct switchtec_dev *switchtec_open(const char *device)
 	char *endptr;
 	struct switchtec_dev *ret;
 
-	if (sscanf(device, "%2049[^:]:%i", path, &inst) == 2) {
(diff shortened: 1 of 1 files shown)


xyan264:
OK, updated to fix PCIe address parsing in this PR only. 

I will create another PR to fix I2C device name parsing issue.

lsgunth:
I said two commits. You don't need a separate PR.

xyan264:
I see. Code is now updated with 2 separate commits.

PR #233, opened 2020-07-29
Removed a call to the deprecated MRPC Subcommand DIE_TEMP_SET_MEAS

yungwesl:
Switchtec devices running FW 03.50.003E and newer have deprecated sub-command 1 in the MRPC GET_DIE_TEMP command.  Sub-command 1 allowed users to set the measurement type which is no longer an option and the switch always reads the temperature in a fixed way.  

I have made a small edit to the temp_linux.py call.


examples/temp_linux.py
```diff
@@ -83,7 +83,7 @@ def echo_cmd(dev):
                            format(sub_cmd_in, sub_cmd_out))
 
 def die_temp(dev):
-    dev.cmd(MRPC_DIETEMP, struct.pack("<L", MRPC_DIETEMP_SET_MEAS))
+    #dev.cmd(MRPC_DIETEMP, struct.pack("<L", MRPC_DIETEMP_SET_MEAS))
```
lsgunth, line 86: This is example code. It really should be clean and clear and having commented out bits is very confusing. Just delete the code if it shouldn't be done that way anymore.

yungwesl: Agreed.  I will commit the new change.

xyan264:
I see 2 commits here. Maybe we should combine them into 1 commits?

lsgunth:
Oh, yes I missed that. They should definitely be squashed.

yungwesl:
Sorry, I pushed a commit of some un-related code today and did not realize it would taint the pull request.  I will close this request and make a larger pull request at a later date as this one small change is pretty benign.

yungwesl:
Closing pull request in preparation for larger request when I add aa new tool.  I will also include the fix to the example at this time

PR #239, opened 2020-11-02
Add switchtec_adm_passthru function

xyan264:
This function supports NVMe admin passthru command with 4K byte response data from SRIOV devices. 

To support 4K response data, the MRPC command uses 3 subcommands:
- `start`: send the admin passthru command and read response size
- `data`: read response data from a given offset
- `end`: notify FW that data read has completed


lib/fabric.c
```diff
@@ -1374,3 +1374,119 @@ int switchtec_ep_bar_write64(struct switchtec_dev *dev, uint16_t pdfid,
...
+	return switchtec_cmd(dev, MRPC_ADM_PASSTHRU,
+			     &cmd, sizeof(cmd), NULL, 0);
+}
+
+int switchtec_adm_passthru(struct switchtec_dev *dev,
+			   struct switchtec_adm_passthru_req *req,
+			   struct switchtec_adm_passthru_rsp *rsp)
+{
+	int ret;
+
```
kelvin-cao, line 1478: Test req and rsp for NULL.

inc/switchtec/fabric.h
```diff
@@ -700,6 +700,28 @@ int switchtec_ep_bar_write32(struct switchtec_dev *dev, uint16_t pdfid,
 			     uint8_t bar_index, uint32_t val, uint64_t addr);
 int switchtec_ep_bar_write64(struct switchtec_dev *dev, uint16_t pdfid,
 			     uint8_t bar_index, uint64_t val, uint64_t addr);
+
+/********** ADMIN PASSTHRU COMMAND *********/
+#define SWITCHTEC_ADM_PASSTHRU_MAX_RESP (4096 + 128)
+
+struct switchtec_adm_passthru_req {
```
kelvin-cao, line 707: Would admin/ADMIN a better abbr.?

lib/fabric.c
```diff
@@ -1374,3 +1374,119 @@ int switchtec_ep_bar_write64(struct switchtec_dev *dev, uint16_t pdfid,
 	val = htole64(val);
 	return ep_bar_write(dev, pdfid, bar, addr, &val, 8);
 }
+
+static int admin_passthru_start(struct switchtec_dev *dev,
+				struct switchtec_adm_passthru_req *req,
+				struct switchtec_adm_passthru_rsp *rsp)
```
kelvin-cao, line 1380: The only output is the rsp_len in *rsp, we can just use it as the output parameter instead of the structure.

kelvin-cao: I just realize you are using the rsp to pass the length to _data(). For an EP related API, I'd suggest we use (struct switchtec_dev *dev, uint16_t pdfid) as the first two parameters to designate the EP. I know switchtec_device_manage() is using the req/rsp style parameter list, but it seems (struct switchtec_dev *dev, uint16_t pdfid) makes more sense and also align with most of the other APIs. We can retire the old switchtec_device_manage() in the future. So maybe we can also have other parameters directly in the parameter list, instead of embedding them in a structure.

lib/fabric.c
```diff
@@ -1374,3 +1374,133 @@ int switchtec_ep_bar_write64(struct switchtec_dev *dev, uint16_t pdfid,
...
+	struct {
+		uint8_t subcmd;
+		uint8_t rsvd[3];
+		uint16_t rsp_len;
+		uint16_t rsvd1;
+	} reply = {};
+
+	cmd.subcmd = MRPC_ADMIN_PASSTHRU_START;
+	cmd.pdfid = htole16(pdfid);
+	cmd.expected_rsp_len = htole16(*rsp_len);
```
lsgunth, line 1401: Why not just initialize the structs in the initializer?

lib/fabric.c
```diff
@@ -1374,3 +1374,133 @@ int switchtec_ep_bar_write64(struct switchtec_dev *dev, uint16_t pdfid,
...
+		uint8_t subcmd;
+		uint8_t rsvd[3];
+		uint16_t offset;
+		uint16_t len;
+		uint8_t data[MRPC_MAX_DATA_LEN - 8];
+	} reply = {};
+
+	while (offset < rsp_len) {
+		cmd.subcmd = MRPC_ADMIN_PASSTHRU_DATA;
+		cmd.pdfid = htole16(pdfid);
```
lsgunth, line 1442: Move the constant stuff outside the loop into the initializer of the struct.

lib/fabric.c
```diff
@@ -1374,3 +1374,133 @@ int switchtec_ep_bar_write64(struct switchtec_dev *dev, uint16_t pdfid,
...
+{
+	struct {
+		uint8_t subcmd;
+		uint8_t rsvd[3];
+		uint16_t pdfid;
+		uint16_t rsvd1;
+	} cmd = {};
+
+	cmd.subcmd = MRPC_ADMIN_PASSTHRU_END;
+	cmd.pdfid = htole16(pdfid);
```
lsgunth, line 1469: Ditto

lib/fabric.c
```diff
@@ -1374,3 +1374,133 @@ int switchtec_ep_bar_write64(struct switchtec_dev *dev, uint16_t pdfid,
...
+ * @param[in] pdfid	The PDFID for the device
+ * @param[in] data_len	Length of data for this command
+ * @param[in] data	Data for this command
+ * @param[in/out] rsp_len Expected reply length/actual response length
+ * @param[out] rsp	Reply from device
+ * @return 0 on success, error code on failure
+ */
+int switchtec_admin_passthru(struct switchtec_dev *dev, uint16_t pdfid,
+			     uint16_t data_len, void *data,
+			     uint16_t *rsp_len, void *rsp)
```
lsgunth, line 1487: Suggest using size_t for API functions.... specific sized types for these is not a great idea.

lsgunth (commented):
It's very vague on what this does and what this is used for. And the arguments seem awfully generic which, to me, hints at bad design but it's hard to evaluate with the given information.

xyan264:
@lsgunth Code updated based on the comments above. Also added the following message to the commit log to explain what the function does, copied below:
```This function supports sending an Admin Command, as defined in Chapter 5 of NVMe specification v1.4, to the device's Admin Submission Queue, and reads back reply from the device.

As the reply from the device is 4096 bytes long, exceeding the maximum data length of an MRPC packet, the command is further divided into 3 subcommands: sending the Admin Command, reading the reply data (in chunks), and completing the command.```

lsgunth:
There's an NVMe host in switchtec now?

Might be worth calling it switchtec_nvme_admin_passthru() then.... "adm" alone is too generic.

lsgunth (approved):
Looks fine to me, thanks.

xyan264:
@lsgunth Agree with the naming, `switchtec_nvme_admin_passthru` is better. 

The switch just acts as a channel through which the host can submit Admin Commands to the NVMe v1.4 - compliant NVMe devices, and reads back reply from the device. From the host perspective the switch is a transparent channel.

This function is provided specifically for this purpose.

kelvin-cao (approved):
One more change: MRPC_ADMIN_* macros to MRPC_NVME_ADMIN_*. Otherwise looks good to me. Thanks.

PR #242, opened 2020-11-24
Add new command fw_meta_info for getting metadata info from Flash partition

rakeshvitta:
This change is to add a new command "fw_meta_info" to get the Metadata information from the flash partitions. 

Changes include:
1. New function fw_meta_info() to handle the command. Metadata info is derived by calling the switchtec_fw_part_info() function that already exists in fw.c which gets the required info from the switch.
2. Helper function part_to_string() to convert partition ID to partition name string.
3. Add the definition of switchtec_fw_part_info() to switchtec.h file instead of it being a static function in fw.c.
4. Moved the structure switchtec_fw_metadata_gen4 from fw.c to switchtec.h

Thanks and Regards,
Rakesh Vitta

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -1491,7 +1491,7 @@ static int fw_meta_info(int argc, char **argv)
 
 	if (inf.valid) {
 		printf("Image Version: %s \n", inf.version);
-		printf("Image Length: 0x%lx \n", inf.image_len);
+		printf("Image Length: 0x%x \n", (int)inf.image_len);
 		printf("Image CRC: 0x%lx \n", inf.image_crc);
 		printf("Secure Version: 0x%lx \n", inf.secure_version);
 		printf("Public Key Exponent : 0x");
--- a/cli/main.c
+++ b/cli/main.c
@@ -1434,24 +1434,25 @@ static int fw_info(int argc, char **argv)
 	return 0;
 }
 
-static char * part_to_string(int partition_id) {
-       switch(partition_id){
+static char *part_to_string(int partition_id)
+{
+	switch (partition_id) {
 
-               case 0: return "Partition Map 0";
-               case 1: return "Partition Map 1";
-               case 2: return "Key Manifest 0";
-               case 3: return "Key Manifest 1";
-               case 4: return "BL2 0";
-               case 5: return "BL2 1";
-               case 6: return "Cfg 0";
-               case 7: return "Cfg 1";
-               case 8: return "Main Fw 0";
-               case 9: return "Main FW 1";
-               default: return "Invalid Partition ID";
-       }
+	case 0: return "Partition Map 0";
+	case 1: return "Partition Map 1";
+	case 2: return "Key Manifest 0";
+	case 3: return "Key Manifest 1";
+	case 4: return "BL2 0";
+	case 5: return "BL2 1";
+	case 6: return "Cfg 0";
+	case 7: return "Cfg 1";
+	case 8: return "Main Fw 0";
+	case 9: return "Main FW 1";
+	default: return "Invalid Partition ID";
+	}
 }
 
-#define CMD_DESC_FW_META_INFO "return metadata information from the selected flash partition"
+#define CMD_DESC_FW_META_INFO "Return metadata information from the selected flash partition"
 
 static int fw_meta_info(int argc, char **argv)
 {
--- a/cli/main.c
+++ b/cli/main.c
@@ -1434,6 +1434,88 @@ static int fw_info(int argc, char **argv)
 	return 0;
 }
 
+static char * part_to_string(int partition_id) {
+       switch(partition_id){
+
+               case 0: return "Partition Map 0";
+               case 1: return "Partition Map 1";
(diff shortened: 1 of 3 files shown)


cli/main.c
```diff
@@ -1434,6 +1434,88 @@ static int fw_info(int argc, char **argv)
 	return 0;
 }
 
+static char * part_to_string(int partition_id) {
+       switch(partition_id){
+
```
lsgunth, line 1439: We use the Linux code style in this project. Please review the documentation for that before submitting:

https://www.kernel.org/doc/html/latest/process/coding-style.html

The first two lines have a ton of style issues:

- Pointer should not have a space after the asterisk
- The open brace for a function should be on it's own line
- There should be a space after "switch"
- There should be a space before the open brace of switch
- Tabs should be used instead of 8 spaces.

inc/switchtec/switchtec.h
```diff
@@ -241,6 +241,38 @@ struct switchtec_fw_image_info {
 	unsigned long secure_version;
 };
 
+struct switchtec_fw_metadata_gen4 {
```
lsgunth, line 244: Changes to the library should be in another patch.

This is a low level hardware interface that should not be exposed outside of the library.

If we need to print more information we should extend the existing interfaces, not bypass them.

cli/main.c
```diff
@@ -1434,6 +1434,88 @@ static int fw_info(int argc, char **argv)
...
+               int partition_id;
+       } cfg = {};
+
+       struct switchtec_fw_image_info inf;
+       struct switchtec_fw_metadata_gen4 *metadata;
+
+       const struct argconfig_options opts[] = {
+               DEVICE_OPTION,
+               {"flash_part", 'p', "", CFG_NONNEGATIVE, &cfg.partition_id, required_argument,
+                       "Flash Partition ID"},
```
lsgunth, line 1470: How does the user know which number for which partition ID? Why should the CLI be creating a new number for this? Other CLI commands use short strings to reference the partition type ("BOOT", "MAP", etc) not an opaque number.

lsgunth (changes requested):
It's not clear what the benefit here is. Please describe in the commit message what this new command is providing over the existing firmware info interfaces.

rakeshvitta:
> It's not clear what the benefit here is. Please describe in the commit message what this new command is providing over the existing firmware info interfaces.

This command prints out the metadata information for the flash partition specified. The metadata includes information like public key modulus, exponent and secure version which are not shown in the fw-info command.

cli/main.c
rakeshvitta: Fixed the style issues and checked the patch using checkpatch.pl script. 
Commit: 10889e6acb50ac44dce8a53e9cd622d993713aba

inc/switchtec/switchtec.h
rakeshvitta: Done. Implemented the library changes as a different commit. I have put the metadata definition back to the library and added new variables to the existing structure to expose the metadata fields from the flash partition. 
I had to expose the switchtec_part_info() as this provides a interface to get the flash info for a particular partition instead of all the flash partitions that the current exposed function provides.

cli/main.c
rakeshvitta: This partition ID is the partition number in the flash and device spec has the details. We cannot use the short strings as they only specify what type of partition it is and does not specify if we are looking at the active or redundant type of the partition.

kelvin-cao: The partition ID can be inconsistent over generations in the firmware. Using short strings provides better compatibility.

kelvin-cao:
> 
> 
> > It's not clear what the benefit here is. Please describe in the commit message what this new command is providing over the existing firmware info interfaces.
> 
> This command prints out the metadata information for the flash partition specified. The metadata includes information like public key modulus, exponent and secure version which are not shown in the fw-info command.

Current fw-info provides some info from meta data, like CRC. We don't need a brand new command just provide the info that the meta data MRPC command returns. Instead, maybe we can introduce a '-v | --verbose' mode to the fw-info to include these extra data. (We don't want to change the default fw-info behavior as it will hurt compatibility)

kelvin-cao:
> 
> 
> > It's not clear what the benefit here is. Please describe in the commit message what this new command is providing over the existing firmware info interfaces.
> 
> This command prints out the metadata information for the flash partition specified. The metadata includes information like public key modulus, exponent and secure version which are not shown in the fw-info command.

Current fw-info provides some info from meta data, like CRC. We don't need a brand new command just provide the info that the meta data MRPC command returns. Instead, maybe we can introduce a '-v | --verbose' mode to the fw-info to include these extra data. (We don't want to change the default fw-info behavior as it will hurt compatibility)

kelvin-cao:
To fix issues in a patch in the PR, you may want to re-base to the patch and fix the issues then hard push them, instead of adding a new patch. And there should be no CI check failures.

rakeshvitta:
Will close this pull request and add the code under fw-info with a --v option instead of adding a brand new command to print the meta-data info of a partition. Since the code would be very different from what I added here will close this pull request and create a new one.

PR #243, opened 2020-11-26
Enable sending 4K command data

xyan264:
In the previous implementation, data over 1016 bytes are truncated, due to MRPC data size limitation. 

In the new subcommand `MRPC_NVME_ADMIN_PASSTHRU_START` definition, a `more_data` field is added to indicate if there is more data to be sent for this transaction. This enables use to send command data of more than 1016 bytes in length.

A loop is added to iteratively send all data using this subcommand.


lsgunth (commented):
Looks good to me.

kelvin-cao (approved):
Looks good. Thanks.

PR #244, opened 2020-12-14
Bypass secure version checking for unsiged image

xyan264:
Secure version checking compares image secure version to that of the device, and issue a warning if image secure version is higher.

For unsigned image, the image secure version has no effect, so for such images, we should bypass secure version checking.

We determine if an image is unsigned by checking the public_key_exponent field.

### DIFF
--- a/lib/fw.c
+++ b/lib/fw.c
@@ -603,6 +603,7 @@ static int switchtec_fw_file_info_gen3(int fd,
 	info->type = switchtec_fw_id_to_type(info);
 
 	info->secure_version = 0;
+	info->signed_image = 0;
 
 	return 0;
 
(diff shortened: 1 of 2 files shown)


lib/fw.c
```diff
@@ -667,6 +669,9 @@ static int switchtec_fw_file_info_gen4(int fd,
 	info->type = switchtec_fw_id_to_type(info);
 
 	info->secure_version = le32toh(hdr.secure_version);
+	ret = memcmp(hdr.public_key_exponent, key_exp_zero, 4);
+	info->signed_image = ret? 1 : 0;
```
lsgunth, line 673: Style nit:

```
info->signed image = !!memcmp(...);
```

lsgunth (approved):
Looks good, except for one nit.

xyan264:
Updated and merged.

PR #245, opened 2021-01-18
Adding GEN5 support in switchtec-user

xyan264:
This is the first of a series of PRs created to support the new PCIe GEN5 devices.

The plan is to review and submit changes to 'gen5-devel' branch, then after a final validation, merge all changes to 'devel' branch in one batch.

This PR adds PCIe GEN5 device/variant types and IDs.

Other changes will be submitted in later PRs.


xyan264:
@lsgunth @kelvin-cao Any comments?

lsgunth (approved):
Looks good to me. No real functional changes just adding IDs.

lib/switchtec.c
```diff
@@ -135,6 +135,36 @@ static const struct switchtec_device_id switchtec_device_id_tbl[] = {
...
+	{0x5152, SWITCHTEC_GEN5, SWITCHTEC_PSX},   //PSX 52XG5
+	{0x5136, SWITCHTEC_GEN5, SWITCHTEC_PSX},   //PSX 36XG5
+	{0x5128, SWITCHTEC_GEN5, SWITCHTEC_PSX},   //PSX 28XG5
+	{0x5200, SWITCHTEC_GEN5, SWITCHTEC_PAX},   //PAX 100XG5
+	{0x5284, SWITCHTEC_GEN5, SWITCHTEC_PAX},   //PAX 84XG5
+	{0x5268, SWITCHTEC_GEN5, SWITCHTEC_PAX},   //PAX 68XG5
+	{0x5252, SWITCHTEC_GEN5, SWITCHTEC_PAX},   //PAX 52XG5
+	{0x5236, SWITCHTEC_GEN5, SWITCHTEC_PAX},   //PAX 36XG5
+	{0x5228, SWITCHTEC_GEN5, SWITCHTEC_PAX},   //PAX 28XG5
+	{0x5300, SWITCHTEC_GEN5, SWITCHTEC_PAXA},  //PAX-A 100XG5
```
kelvin-cao, line 156: What does -A here mean? Is it for rev A? If yes I don't think we need it as the rev B chip will use the same device ID.

kelvin-cao: If it's for another variant, I'm good with these patches.

xyan264: Confirmed with PM, PAX-A/PFX-A are indeed different variants for specific markets. Will keep the IDs as-is.

PR #246, opened 2021-01-21
Add support for parsing GEN5 FW images

xyan264:
Background: 
- GEN5 device uses the same firmware image format as GEN4, so in theory we can reuse all GEN4 firmware image parsing functions. 
- The firmware image version number contains 8 bits that tells us which device generation this image is targeted for

In this PR, we
- reuse GEN4 FW image parsing function on GEN5 images
- add checks to ensure `fw-update` only update FW image with matching generation type

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -1493,6 +1493,14 @@ static int fw_update(int argc, char **argv)
 		return -1;
 	}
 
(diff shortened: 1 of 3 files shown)


lsgunth (approved):
Looks very clean to me.

Thanks.

kelvin-cao (approved):
Looks good. Thanks.

PR #247, opened 2021-01-22
Fix 'temp' command for GEN5 device

xyan264:
GEN5 uses the same command as GEN4 to get chip temperature readings. 

Since the device GEN field is always known during initialization (otherwise program errors out), it is safe to delete the GEN3/GEN4 check.


lsgunth (approved):
Makes sense to me.

kelvin-cao (approved):
Thanks.

PR #248, opened 2021-01-25
Add support for fw-info/read command

xyan264:
Background: GEN5 uses the same MRPC commands for `fw-info` and `fw-read` command as GEN4 devices.

The PR adds support for GEN5 for `fw-info` and `fw-read` command.

### DIFF
--- a/lib/fw.c
+++ b/lib/fw.c
@@ -1495,7 +1495,8 @@ int switchtec_fw_img_write_hdr(int fd, struct switchtec_fw_image_info *info)
 {
 	switch (info->gen) {
 	case SWITCHTEC_GEN3: return switchtec_fw_img_write_hdr_gen3(fd, info);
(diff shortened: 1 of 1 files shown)


lib/fw.c
```diff
@@ -1075,7 +1075,7 @@ static int switchtec_fw_part_info(struct switchtec_dev *dev, int nr_info,
 	if (info == NULL || nr_info == 0)
 		return -EINVAL;
 
-	if (dev->gen == SWITCHTEC_GEN4) {
+	if (dev->gen != SWITCHTEC_GEN3) {
```
kelvin-cao, line 1078: Maybe we can use the style '> SWITCHTEC_GEN3', which implies that the change happened from GEN4. And when we have something new in GEN5 we can use the same style. Otherwise looks good to me. Thanks.

xyan264: Updated.

kelvin-cao (approved):
Thanks.

lsgunth (approved):
Looks good to me.

PR #249, opened 2021-01-26
Fix security settings for GEN5 devices

xyan264:
Background: GEN5 device uses a different MRPC command to read chip security settings.  The security settings of GEN5 are different from GEN4 devices in the following:
- I2C address changes from 7-bit to 8-bit
- I2C command map starting bits changes from bit 29 to bit 30
- I2C command map is now 14-bit long, instead of 12-bit in GEN4
- Adds "SPI clock rate high" field (a reserved field, always 0 on GEN4) to indicate that SPI is running in 'turbo' mode

In this PR, we submit the following changes:
- Add function to get I2C address and command map operands (bit shift, bit mask) based on device generation
- Add `gen` field to security setting structure so I2C address can display accordingly
- Add `spi_core_clk_high` to security setting structure so SPI clock rate can display accordingly

### DIFF
--- a/cli/mfg.c
+++ b/cli/mfg.c
@@ -64,11 +64,6 @@ static const struct argconfig_choice secure_state_choices[] = {
 	{}
 };
 
-static char *spi_rate_str[] = {
-	"100", "67", "50", "40", "33.33", "28.57",
-	"25", "22.22", "20", "18.18"
-};
-
 static const char* phase_id_to_string(enum switchtec_boot_phase phase_id)
 {
 	switch(phase_id) {
@@ -176,8 +171,7 @@ static void print_security_config(struct switchtec_security_cfg_state *state,
 	printf("\tJTAG/EJTAG Unlock AFTER BL1: \t%d\n",
 		state->jtag_post_bl1_unlock_allowed);
 
-	printf("\tSPI Clock Rate: \t\t%s MHz\n",
-		spi_rate_str[state->spi_clk_rate-1]);
+	printf("\tSPI Clock Rate: \t\t%.2f MHz\n", state->spi_clk_rate);
 
 	printf("\tI2C Recovery TMO: \t\t%d Second(s)\n",
 		state->i2c_recovery_tmo);
@@ -246,8 +240,7 @@ static void print_security_cfg_set(struct switchtec_security_cfg_set *set)
 	printf("\tJTAG/EJTAG Unlock AFTER BL1: \t%d\n",
 		set->jtag_post_bl1_unlock_allowed);
 
-	printf("\tSPI Clock Rate: \t\t%s MHz\n",
-		spi_rate_str[set->spi_clk_rate-1]);
+	printf("\tSPI Clock Rate: \t\t%.2f MHz\n", set->spi_clk_rate);
 
 	printf("\tI2C Recovery TMO: \t\t%d Second(s)\n",
 		set->i2c_recovery_tmo);
@@ -830,7 +823,8 @@ static int config_set(int argc, char **argv)
 		return -2;
 	}
 
-	ret = switchtec_read_sec_cfg_file(cfg.setting_fimg, &settings);
+	ret = switchtec_read_sec_cfg_file(cfg.dev, cfg.setting_fimg,
+					  &settings);
 	fclose(cfg.setting_fimg);
 	if (ret) {
 		fprintf(stderr, "Invalid secure setting file: %s!\n",
--- a/inc/switchtec/mfg.h
+++ b/inc/switchtec/mfg.h
@@ -32,6 +32,8 @@
 #define SWITCHTEC_KMSK_LEN	64
 #define SWITCHTEC_KMSK_NUM	4
 
+#define SWITCHTEC_SECURITY_SPI_RATE_MAX_NUM	16
+
 struct switchtec_sn_ver_info {
 	uint32_t chip_serial;
 	uint32_t ver_km;
@@ -155,12 +157,19 @@ struct switchtec_signature{
 	uint8_t signature[SWITCHTEC_SIG_LEN];
 };
 
+struct switchtec_security_spi_avail_rate {
+	int num_rates;
+	float rates[SWITCHTEC_SECURITY_SPI_RATE_MAX_NUM];
+};
+
 int switchtec_sn_ver_get(struct switchtec_dev *dev,
 			 struct switchtec_sn_ver_info *info);
 int switchtec_security_config_get(struct switchtec_dev *dev,
 			          struct switchtec_security_cfg_state *state);
 int switchtec_security_config_get_ext(struct switchtec_dev *dev,
 		struct switchtec_security_cfg_state_ext *ext);
+int switchtec_security_spi_avail_rate_get(struct switchtec_dev *dev,
+		struct switchtec_security_spi_avail_rate *rates);
 int switchtec_security_config_set(struct switchtec_dev *dev,
 				  struct switchtec_security_cfg_set *setting);
 int switchtec_mailbox_to_file(struct switchtec_dev *dev, int fd);
--- a/inc/switchtec/mfg.h
+++ b/inc/switchtec/mfg.h
@@ -52,19 +52,6 @@ enum switchtec_secure_state {
 	SWITCHTEC_SECURE_STATE_UNKNOWN = 0xff,
 };
 
-enum switchtec_spi_clk_rate {
-	SWITCHTEC_SPI_RATE_100M = 1,
-	SWITCHTEC_SPI_RATE_67M,
-	SWITCHTEC_SPI_RATE_50M,
-	SWITCHTEC_SPI_RATE_40M,
-	SWITCHTEC_SPI_RATE_33_33M,
-	SWITCHTEC_SPI_RATE_28_57M,
-	SWITCHTEC_SPI_RATE_25M,
-	SWITCHTEC_SPI_RATE_22_22M,
-	SWITCHTEC_SPI_RATE_20M,
-	SWITCHTEC_SPI_RATE_18_18M
-};
-
 struct switchtec_security_cfg_state {
 	uint8_t basic_setting_valid;
 	uint8_t public_key_exp_valid;
@@ -80,7 +67,7 @@ struct switchtec_security_cfg_state {
 	uint8_t jtag_bl1_unlock_allowed;
 	uint8_t jtag_post_bl1_unlock_allowed;
 
-	enum switchtec_spi_clk_rate spi_clk_rate;
+	float spi_clk_rate;
 	uint32_t i2c_recovery_tmo;
 	uint32_t i2c_port;
 	uint32_t i2c_addr;
@@ -128,7 +115,7 @@ struct switchtec_security_cfg_set {
 	uint8_t jtag_bl1_unlock_allowed;
 	uint8_t jtag_post_bl1_unlock_allowed;
 
-	uint32_t spi_clk_rate;
+	float spi_clk_rate;
 	uint32_t i2c_recovery_tmo;
 	uint32_t i2c_port;
 	uint32_t i2c_addr;
@@ -199,8 +186,9 @@ int switchtec_dbg_unlock_version_update(struct switchtec_dev *dev,
 					uint32_t ver_sec_unlock,
 					struct switchtec_pubkey *public_key,
 			 		struct switchtec_signature *signature);
-int switchtec_read_sec_cfg_file(FILE *setting_file,
-			        struct switchtec_security_cfg_set *set);
+int switchtec_read_sec_cfg_file(struct switchtec_dev *dev,
+				FILE *setting_file,
+				struct switchtec_security_cfg_set *set);
 int switchtec_read_pubk_file(FILE *pubk_file, struct switchtec_pubkey *pubk);
 int switchtec_read_kmsk_file(FILE *kmsk_file, struct switchtec_kmsk *kmsk);
 int switchtec_read_signature_file(FILE *sig_file,
--- a/inc/switchtec/mrpc.h
+++ b/inc/switchtec/mrpc.h
@@ -91,6 +91,7 @@ enum mrpc_cmd {
 	MRPC_DBG_UNLOCK = 266,
 	MRPC_BOOTUP_RESUME = 267,
 	MRPC_SECURITY_CONFIG_GET_GEN5 = 268,
+	MRPC_SECURITY_CONFIG_SET_GEN5 = 269,
 };
 
 enum mrpc_bg_status {
--- a/inc/switchtec/mrpc.h
+++ b/inc/switchtec/mrpc.h
@@ -90,6 +90,7 @@ enum mrpc_cmd {
 	MRPC_SN_VER_GET = 265,
 	MRPC_DBG_UNLOCK = 266,
 	MRPC_BOOTUP_RESUME = 267,
+	MRPC_SECURITY_CONFIG_GET_GEN5 = 268,
 };
 
 enum mrpc_bg_status {
--- a/lib/mfg.c
+++ b/lib/mfg.c
@@ -367,6 +367,7 @@ int switchtec_security_config_set(struct switchtec_dev *dev,
 	uint32_t map_shift;
 	uint32_t map_mask;
 	int spi_clk;
+	uint8_t cmd_buf[20] = {};
 
 	ret = get_configs(dev, &reply);
 	if (ret)
@@ -411,8 +412,15 @@ int switchtec_security_config_set(struct switchtec_dev *dev,
 
 	sd.pub_key_exponent = htole32(setting->public_key_exponent);
 
-	ret = switchtec_mfg_cmd(dev, MRPC_SECURITY_CONFIG_SET, &sd, sizeof(sd),
-				NULL, 0);
+	if (switchtec_gen(dev) == SWITCHTEC_GEN4) {
+		ret = switchtec_mfg_cmd(dev, MRPC_SECURITY_CONFIG_SET,
+					&sd, sizeof(sd), NULL, 0);
+	} else {
+		cmd_buf[0] = 1;
+		memcpy(cmd_buf + 4, &sd, sizeof(sd));
+		ret = switchtec_mfg_cmd(dev, MRPC_SECURITY_CONFIG_SET_GEN5,
+					cmd_buf, sizeof(cmd_buf), NULL, 0);
+	}
 	return ret;
 }
 
--- a/lib/mfg.c
+++ b/lib/mfg.c
@@ -169,6 +169,27 @@ static int get_configs(struct switchtec_dev *dev, struct get_cfgs_reply *cfgs)
 	return ret;
 }
 
+int switchtec_security_spi_avail_rate_get(struct switchtec_dev *dev,
+		struct switchtec_security_spi_avail_rate *rates)
+{
+	int ret;
+	struct get_cfgs_reply reply;
+
+	ret = get_configs(dev, &reply);
+	if (ret)
+		return ret;
+
+	rates->num_rates = 10;
+	if (reply.spi_core_clk_high)
+		memcpy(rates->rates, spi_clk_hi_rate_float,
+		       sizeof(spi_clk_hi_rate_float));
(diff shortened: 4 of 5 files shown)


lib/mfg.c
```diff
@@ -82,8 +82,11 @@
 #define SWITCHTEC_I2C_PORT_BITMASK		0x0f
 #define SWITCHTEC_I2C_ADDR_BITSHIFT		22
 #define SWITCHTEC_I2C_ADDR_BITMASK		0x7f
+#define SWITCHTEC_I2C_ADDR_BITMASK_GEN5	0xff
 #define SWITCHTEC_CMD_MAP_BITSHIFT		29
+#define SWITCHTEC_CMD_MAP_BITSHIFT_GEN5	30
 #define SWITCHTEC_CMD_MAP_BITMASK		0xfff
+#define SWITCHTEC_CMD_MAP_BITMASK_GEN5		0x3fff
```
xyan264, line 89: Some editors show this line as aligned, others don't. So it should be OK to ignore this 'mis-align issue'.

inc/switchtec/mrpc.h
```diff
@@ -90,6 +90,7 @@ enum mrpc_cmd {
 	MRPC_SN_VER_GET = 265,
 	MRPC_DBG_UNLOCK = 266,
 	MRPC_BOOTUP_RESUME = 267,
+	MRPC_SECURITY_INFO_GET = 268,
```
kelvin-cao, line 93: If it's a Gen5 version of MRPC_SECURITY_CONFIG_GET, maybe we can use suffix _GEN5 to name it. (and also align with other naming in this patch)

cli/mfg.c
```diff
@@ -182,7 +182,14 @@ static void print_security_config(struct switchtec_security_cfg_state *state,
 	printf("\tI2C Recovery TMO: \t\t%d Second(s)\n",
 		state->i2c_recovery_tmo);
 	printf("\tI2C Port: \t\t\t%d\n", state->i2c_port);
-	printf("\tI2C Address (7-bits): \t\t0x%02x\n", state->i2c_addr);
+	if (state->gen == SWITCHTEC_GEN5) {
```
kelvin-cao, line 195: No braces for single line statement in if-else branch.

cli/mfg.c
```diff
@@ -182,7 +182,14 @@ static void print_security_config(struct switchtec_security_cfg_state *state,
...
+		printf("\tI2C Address (7-bits): \t\t0x%02x\n",
+		       state->i2c_addr>>1);
```
kelvin-cao, line 197: Why not use 7-bit value when we assign it to state->i2c_addr?

inc/switchtec/mfg.h
```diff
@@ -66,6 +66,7 @@ enum switchtec_spi_clk_rate {
 };
 
 struct switchtec_security_cfg_state {
+	enum switchtec_gen gen;
```
kelvin-cao, line 69: If we always have 7-bit address in state->i2c_addr, no need to add gen here.

inc/switchtec/mfg.h
```diff
@@ -89,6 +89,7 @@ struct switchtec_security_cfg_state {
 	uint32_t public_key_exponent;
 	uint32_t public_key_num;
 	uint32_t public_key_ver;
+	uint32_t spi_core_clk_high;
```
kelvin-cao, line 92: Why not expand `switchtec_spi_clk_rate` in the structure? I don't see why this new field is necessary.

lib/mfg.c
```diff
@@ -163,8 +183,16 @@ static int secure_config_get(struct switchtec_dev *dev,
 				*otp_valid = true;
 		}
 	} else {
-		ret = switchtec_mfg_cmd(dev, MRPC_SECURITY_CONFIG_GET, NULL, 0,
-					&reply, sizeof(reply));
+		if (switchtec_gen(dev) == SWITCHTEC_GEN4) {
```
lsgunth, line 186: It would be better if the I2c changes and MRPC changes were in different patches...

cli/mfg.c
lsgunth: Yeah, this seems highly questionable.... We should store the address the same for each generation -- not have a change in how it's printed...

inc/switchtec/mfg.h
lsgunth: Yes, I agree.

lib/mfg.c
```diff
@@ -111,13 +114,30 @@ static void RSA_get0_key(const RSA *r, const BIGNUM **n,
 }
 #endif
 
+static void get_i2c_operands(enum switchtec_gen gen, uint32_t *addr_mask,
+			     uint32_t *map_shift, uint32_t *map_mask)
```
lsgunth, line 118: This is a pretty ugly function and it's only used in one place.

Why not just inline the if and avoid the indirection.

inc/switchtec/mfg.h
xyan264: `enum switchtec_spi_clk_rate spi_clk_rate` can take value from [1-10]. By default, these values map to [100 - 18.18MB] clock rate. 

On GEN5, there is a bootstrap pin (or consider it as a _turbo_ key), when set, the same value [1-10] maps to [120 - 21.8MB] clock rate. The bootstrap pin state is returned in `spi_core_clk_high` during MRPC. 

So from the chip's perspective, it's the same set of values [1-10] that are interpreted differently, according to the bootstrap pin, or `spi_core_clk_high` flag. 

I think it's safer if we align with this view, otherwise it adds overhead (translation in `cfg_read` function, translation-back in `config_set` function), and could cause more pain in the future.

The `config-set` function will be submitted later.

lsgunth: I'd make a change to store the SPI rate as a float inside switchtec_security_cfg_set, so these details can be managed in the library -- instead of having all these hardware specific details leaking into the CLI where they don't belong and just complicate the print functions.

The CLI shouldn't contain a mapping between a hardware specific number and the actual useful piece of information.

Upon revisiting, I find a lot of the mfg structs confusing. They can't be hardware specifc structs because there are enums in some of them, but then why are they using fixed size types (uint32, etc)?

lib/mfg.c
xyan264: Yes it's not nice, but we need this later for 'config-set' as well. So let's keep it if it's OK.

cli/mfg.c
```diff
@@ -66,7 +66,9 @@ static const struct argconfig_choice secure_state_choices[] = {
 
 static char *spi_rate_str[] = {
 	"100", "67", "50", "40", "33.33", "28.57",
-	"25", "22.22", "20", "18.18"
+	"25", "22.22", "20", "18.18",
+	"120", "80", "60", "48", "40", "34",
+	"30", "26.67", "24", "21.82"
```
lsgunth, line 71: This is still ugly, but it's the existing code that is ugly.

This array is highly questionably and shouldn't even be in the CLI. I still say change it to a float and pass the actual rate instead of making it even more ugly by having a special offset per generation.

lib/mfg.c
lsgunth: Hmm, ok, it's ugly, but I can't think of anything better.

But shouldn't the config-set be fixed up in the same patch?

xyan264: `config-set` has ~3 patches as well: different MRPC command, different I2C settings, different config file format. 
It might be too many commits for one PR. So this PR focuses on 'config-get' only, making it easier to review.

cli/mfg.c
xyan264: Float is much harder to work with... e.g., ``` if (rate < 21.8201 && rate > 21.8199)``` vs ``` if (rate == SWITCHTEC_SPI_RATE_21_82)```

Plus it gives user the impression that all 0.00 - inf.xxx rates are available when in fact only a few rate levels are supported.

lib/mfg.c
xyan264: I see what you mean. `config-set` is fixed as well, below.

cli/mfg.c
lsgunth: We don't have any code that's doing comparisons on the rate, it's really only there to be printed or for basic information.

But it's even harder to do comparisons with the current scheme when the user have to map it to an array of strings and then convert it to an integer.

I don't think the user would assume that any valid float could be used for the rate just because it takes a float in. That's not really sensible.

The config_set function already looks pretty broken but nobody uses it, as it requires the user to know the weird mapping which was just changed by this patch.

Use a float, have the config_set find the nearest value in the table to set it to and have config_get set the float from the table. Then you can have a different table for GEN4 and GEN5 in the library where it belongs.

lsgunth: If you really don't like floats then use an int specified in KHz or some such.... But you still need to do a find nearest operation in confg_set, so it doesn't really get you anything except making it harder to print in the CLI.

xyan264: @lsgunth Printing `float` is tricky ('22.22' printed as `21.19999`) and is hard to maintain compatibility with output of earlier versions. So the code has been updated to use `int` for clock rate in KHz. 

Considering that GEN4 has already been deployed, we do the following to maintain compatibility with earlier versions:
- at API level, we add a new set of `xxxx_gen5` functions and keep the exist functions as-is
- at CLI level, we are free to use the new API functions, but needs to keep output compatible with earlier versions

Please review and give your comments.

inc/switchtec/mfg.h
```diff
@@ -92,6 +92,33 @@ struct switchtec_security_cfg_state {
 	uint8_t public_key[SWITCHTEC_KMSK_NUM][SWITCHTEC_KMSK_LEN];
 };
 
+struct switchtec_security_cfg_state_gen5 {
```
lsgunth, line 95: This is a public struct and is almost identical to the one above it.

We really shouldn't be creating generation specific interfaces unless there is a really good reason to.

Seems like the library should be able to handle all the differences here instead of creating massive duplication all over the place....

lsgunth (commented):
I still do not like this. We should be fixing the interface to be better and general, not creating generation specific, duplicating code and adding extra pain for the callers.

And fix the printing first, for everyone, in some cleanup patches; instead of leaving the existing API mess the way it is. Just because it's a mess now isn't an excuse to create a bigger mess.

> Printing float is tricky ('22.22' printed as 21.19999) so it's hard to maintain compatibility with output of earlier versions. So the code has been updated to use int for clock rate in KHz.

Actually, that is trivial. "%.2f" or "%.4g" would prevent the rounding issue. 

I don't think we need character for character compatibility in the output, anybody who is parsing it would almost certainly understand changes in zeros padded.

xyan264:
@lsgunth 
Code has been updated with the following changes:
- `float` type is used for `spi_clk_rate`, and printed with `%.2f`
- Add 'new' get security settings functions to support both GEN4 and GEN5 devices, and these will be the default interface to use from now on
- Considering that some customers are already developing/deploying GEN4 using existing interface, we have renamed existing interface to `xxx_gen4`, which will eventually become deprecated and be removed from code base in future releases. This allows a smooth transition for existing GEN4 customers

Let us know what you think.

kelvin-cao:
> 
> 
> @lsgunth
> Code has been updated with the following changes:
> 
>     * `float` type is used for `spi_clk_rate`, and printed with `%.2f`
> 
>     * Add 'new' get security settings functions to support both GEN4 and GEN5 devices, and these will be the default interface to use from now on
> 
>     * Considering that some customers are already developing/deploying GEN4 using existing interface, we have renamed existing interface to `xxx_gen4`, which will eventually become deprecated and be removed from code base in future releases. This allows a smooth transition for existing GEN4 customers
> 
> 
> Let us know what you think.

With the function renamed with _gen4 suffix, it breaks any client that was built on top of the original function. And we should avoid having generation string in the naming of public functions like this as the boundary may be vague and what if the firmware back port features from gen5 to gen4?

kelvin-cao:
And we still have `uint32_t spi_clk_rate;` in `struct switchtec_security_cfg_set`. I don't see how we unify all the related APIs.

kelvin-cao:
I think we can go in one of the following paths.
1. Keep the old interfaces (with function names unchanged) to maintain compatibility with any client (I doubt there's any) built on top of these interfaces, and introduce a set of new APIs to support all gens.
2. Replace old interfaces with new ones which will support all gens.
3. A mix of 1 and 2. Keeping both for a while, and remove old ones in the future. 

What do you think?

lsgunth:
If you don't think there are any clients that use it, then just make the change and clean it up completely so that it is sane.

But it appears the only API change is the spi_clock_rate... If we don't care about ABI compatibility and only about API, we can just add a new field for the float and leave the old spi_clock_rate as is. Any existing code that expects the old struct will still compile -- but it won't work for GEN5 because it won't know any of the rate values. 

There's certainly no sense keeping the old interface but renaming it. Then users have to change their code so they might as well just change it to the new interface.

kelvin-cao:
Agree. Let's confirm if there are any clients of the mfg library first.

xyan264:
@lsgunth we have received go-ahead from our team to change the API. Code is now updated. Please review and comment.

xyan264:
@lsgunth Any comments?

lib/mfg.c
```diff
@@ -216,10 +225,18 @@ static int secure_config_get(struct switchtec_dev *dev,
 	state->jtag_bl1_unlock_allowed = !!(reply.cfg & 0x0100);
 	state->jtag_post_bl1_unlock_allowed = !!(reply.cfg & 0x0200);
 
-	state->spi_clk_rate =
-		(reply.cfg >> SWITCHTEC_CLK_RATE_BITSHIFT) & 0x0f;
-	if (state->spi_clk_rate == 0)
-		state->spi_clk_rate = SWITCHTEC_SPI_RATE_25M;
+	spi_clk = (reply.cfg >> SWITCHTEC_CLK_RATE_BITSHIFT) & 0x0f;
+	if (spi_clk == 0) {
+		if (switchtec_gen(dev) == SWITCHTEC_GEN5)
```
lsgunth, line 230: I probably would have seperated this into two patches: one making the float change, the second adding GEN5 support.

lsgunth (approved):
Looks good to me, besides the presentation of the one commit.

lib/mfg.c
```diff
@@ -300,6 +313,31 @@ int switchtec_mailbox_to_file(struct switchtec_dev *dev, int fd)
...
+	if (clk_float >= p[0])
+		return 0;
+	else if (clk_float <= p[9])
+		return 9;
+
+	for (i = 0; i < 9; i++) {
+		if (clk_float >= (p[i] + p[i+1])/2)
+			return i;
+		else if (clk_float >= p[i+1])
+			return i+1;
```
kelvin-cao, line 339: Why not use a string for the clock rate if we will have to lose accuracy when using float? A string is also good for printing.

lsgunth: I don't know why you are worried about accuracy. If you want more accuracy, use a double; but you really don't need it.  A string might be good for printing, but it's not good for comparing or anything else.

kelvin-cao: I'm thinking we should error out if the user input clock rate doesn't hit any one in the array, instead of picking a close one. When user want a particular rate but input an unintended wrong value, we should report the error rather than take it. Even using float, we can compare it with each one in the array within the accuracy provided by float, rather than choose a close one. I believe user would want to have errors reported when the input is not a valid value, since this is an OTP thing. So why not use more accurate type like string or enum? 

It's not hard to compare the input string with an array of valid values, just like what we're doing here for the float. If string is not ideal, how about using enum and a helper in the library to map it to a string for printing?

lsgunth: Generally the user doesn't really care about the exact rate. They only really need to specify a maximum.  In which case we should probably ensure that the we don't set it higher than the user requests.

kelvin-cao: That might be true for some other setting operation. However, I still believe we should set the exact rate user requested for this one time programming operation, as any mistake would cause unrecoverable damage to the chip (and the system). User should be responsible for the OTP setting values, and we should execute exactly what user asks us to do.

lsgunth: It's just an SPI port.... How would using the "wrong" rate cause unrecoverable damage? If the port is clocked at 10Mb/s vs 11Mb/s there should be absolutely no visible difference. 

Theoretically if the clock is pushed higher than what the devices support, that could cause bit errors or some such, but that's why I suggested always going lower; never higher.

xyan264: @lsgunth  I agree with Kelvin's observation. 

Normally for SPI clock rate settings, `float` type is more intuitive and a high rate is sufficient. 

However, with the One-Time-Programming (OTP) settings, we want to ensure no ambiguity, and that users understand exactly what they are doing, so `enum` is a perfect way to ensure that. 

How about we add a new function `switchtec_get_spi_clk_str` that converts `enum` to a printable string, as in patch 
684a459 ?

lsgunth: I really do not like the enum approach. It was very broken, at doesn't let the user actually specify the SPI rate (which is likely going to produce even worse errors).

If you want to ensure the float during programming is exactly what they specified then do that. Compare the float to the int that was looked up in the table and error out if it's not a valid rate.... However then you make it really hard to use as the user doesn't know all the valid rates.  Ensure you print an error with the nearest valid rate or some such.

Though, I really don't think it's as big an issue as you make it out to be.

lsgunth: Yes, the user interface was broken. The ability to add backward compatibility was broken. And it's a cumbersome interface.

The float interface was so much better for all these reasons. Keep it. If you want to enforce the user to pick a supported baudrate and error out when they don't pick a valid value, then do that (but make sure you communicate the valid values). But don't go back to the enum which was the wrong approach from the start.

The whole idea that the exact SPI rate will be a problem is so overblown. It's not like a UART where both sides have to have the same rate. SPI is synchronous so the master can use whatever baudrate it feels like as long as it's not higher than what the other side supports. The exact rate is not important.

xyan264: @lsgunth I agree float would be more intuitive. But I still don't get why `enum` is broken and blocks future backward compatibility. Could you elaborate? I believe an agreement on this would save tons of time in the future reviews.

lsgunth: Well, how does the user know what values they can enter?

The original implementation was broken for two reasons: there was no way to change the enum values because they were hard coded outside the library. Second there was no way for the user to know how to set the values because the interface took only a number with no way to know what each number meant. 

Using an enum means there will need to be a bunch of work to map the enumeration values to actual floating point values. Both for printing and for taking input from the user. We'd need an argconfig choices map that differs depending on the depending on the device but can't print help unless it connects to the device. Doing this in such a way that allows for future expansion (by only changing the library) pretty much impossible. Also imagine someone using the library directly without the CLI, they have to repeat this huge pain that then needs to be reworked if the library or enum values need to change or be expanded.

Better to just use a float with some clear methods to determine if a given number is valid. And perhaps a way to list valid floating point numbers. Most users should not care about the exact SPI rate, they just want it to work -- and they especially don't want what they are doing to break if they upgrade their switch.

xyan264: @lsgunth Thanks for the suggestions. 

I have reverted the code back to `float`, but added checking in `config_set` function to ensure (close to) exact match of the input clock rate, or return error code if not. 

Also added a function to allow user to get available rates, so they can chose the preferred rate from there. 

If nothing else, I will go ahead and merge this.

kelvin-cao (approved):

lib/mfg.c
```diff
@@ -136,6 +136,49 @@ static float spi_clk_hi_rate_float[] = {
...
+					MRPC_SECURITY_CONFIG_GET_GEN5,
+					&subcmd, sizeof(subcmd),
+					&reply, sizeof(reply));
+	} else {
+		ret = switchtec_mfg_cmd(dev, MRPC_SECURITY_CONFIG_GET,
+					NULL, 0, &reply,
+					sizeof(reply));
+	}
+	if (ret)
+		return ret;
```
lsgunth, line 169: Looks like a bunch of this should be factored out of secure_config_get() into a shared common function instead of copied and pasted. Should help to cleanup secure_config_get() as well given how complicated that function has gotten.

xyan264: Thanks for the suggestion. Code is now much cleaner with the common function.

PR #251, opened 2021-03-03
Check device generation in security setting file

xyan264:
Add checks to ensure that the  `generation` flag in the security setting file matches that of the current device.


lsgunth (approved):
Looks good to me.

kelvin-cao (approved):
Looks good. Better use another negative value like '-4' for the new error branch.

xyan264:
Updated and merged.

PR #252, opened 2021-03-10
Fix OTP status get/display for Gen5

xyan264:
This PR contains the following changes:
- merge `switchtec_security_cfg_otp_region` structure into `switchtec_security_cfg_state`, so a `switchtec_security_config_get` returns both security setting and OTP status
- delete redundant `switchtec_security_config_get_ext` function and associated data structures
- fix OTP status get/display for Gen5: Gen5 returns OTP settings in `MRPC_SECURITY_CONFIG_GET_GEN5` command reply, so we can parse OTP status from there

### DIFF
--- a/lib/mfg.c
+++ b/lib/mfg.c
@@ -190,10 +190,14 @@ int switchtec_security_spi_avail_rate_get(struct switchtec_dev *dev,
 	return 0;
 }
 
-static int secure_config_get(struct switchtec_dev *dev,
-			     struct switchtec_security_cfg_state *state,
-			     struct switchtec_security_cfg_otp_region *otp,
-			     bool *otp_valid)
+/**
+ * @brief Get secure boot configurations
+ * @param[in]  dev	Switchtec device handle
+ * @param[out] state	Current secure boot settings
+ * @return 0 on success, error code on failure
+ */
+int switchtec_security_config_get(struct switchtec_dev *dev,
+				  struct switchtec_security_cfg_state *state)
 {
 	int ret;
 	uint8_t subcmd = 0;
@@ -203,8 +207,7 @@ static int secure_config_get(struct switchtec_dev *dev,
 	int spi_clk;
 	struct get_cfgs_reply reply;
 
-	if (otp_valid)
-		*otp_valid = false;
+	state->otp_valid = false;
 
 	ret = switchtec_mfg_cmd(dev, MRPC_SECURITY_CONFIG_GET_EXT,
 				&subcmd, sizeof(subcmd),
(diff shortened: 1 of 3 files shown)


lsgunth (approved):
Code looks fine. But the commit messages could use some work.  Appearantly the second commit fixes something, but I have no idea what....

lib/mfg.c
```diff
@@ -216,23 +237,7 @@ int switchtec_security_config_get(struct switchtec_dev *dev,
 		return ret;
```
kelvin-cao, line 237: Why not move this call into get_configs() where we make other config get MRPC calls?

lib/mfg.c
```diff
@@ -250,6 +255,11 @@ int switchtec_security_config_get(struct switchtec_dev *dev,
 	state->public_key_ver_valid = !!(reply.valid & 0x08);
 	state->public_key_valid = !!(reply.valid & 0x10);
 
+	if (switchtec_gen(dev) == SWITCHTEC_GEN5) {
+		parse_otp_settings(&state->otp, reply.valid);
+		state->otp_valid = true;
```
kelvin-cao, line 260: Then we can merge the checks and have only one place do the parse.

lib/mfg.c
xyan264: At the moment, function `get_configs()` only returns raw MRPC reply data. It does not parse the data in any way.

Not all callers need parsing the raw data (e.g., some callers only care if the bit `spi_core_clk_high` is set). 

Also in GEN4, the `otp_valid` field is determined by if `MRPC_SECURITY_CONFIG_GET_EXT` succeeds, so it's hard to merge the MRPC call to `get_configs()`.

kelvin-cao: In the function switchtec_security_config_get(), we have a call to MRPC_SECURITY_CONFIG_GET_EXT to get the raw reply data, or a call to get_configs() which calls MRPC_SECURITY_CONFIG_GET_GEN5 or MRPC_SECURITY_CONFIG_GET to get the raw data. We can have the code of the three MRPC calls into the get_configs() and then avoid duplication of the below code section:
```             
parse_otp_settings(&state->otp, reply.valid);
state->otp_valid = true;
```
I can think of adding a pass out parameter to get_configs() to indicate if otp is valid (GEN5 or MRPC_SECURITY_CONFIG_GET_EXT works)

xyan264: Good idea. Code is now updated, please review.

kelvin-cao (approved):
Looks good. Thanks.

PR #255, opened 2021-03-17
Fw download error

kelvin-cao:
This PR introduces some fw-update related changes.

- Add some new error codes
- Only check error when background download status is 0xFF to follow the fw behavior


inc/switchtec/switchtec.h
```diff
@@ -719,7 +719,7 @@ enum switchtec_fw_dlstatus {
 	SWITCHTEC_DLSTAT_COMPLETES = 7,
 	SWITCHTEC_DLSTAT_SUCCESS_FIRM_ACT = 8,
 	SWITCHTEC_DLSTAT_SUCCESS_DATA_ACT = 9,
-	SWITCHTEC_DLSTAT_DOWNLOAD_TIMEOUT = 14,
+	SWITCHTEC_DLSTAT_DOWNLOAD_TIMEOUT = 0xe,
```
lsgunth, line 726: If you're going to use hex, you should change all the above to use hex as well, to make it consistent.

lsgunth (approved):
Minor nit, otherwise this looks good to me.

inc/switchtec/switchtec.h
kelvin-cao: Updated.

PR #256, opened 2021-03-22
Fix log dump for Gen5 device

xyan264:
Fix the following for `log-dump` on Gen5 device:
- add extra checking to avoid segmentation fault when parsing log file
- use different subcommand when dumping RAM and FLASH log on Gen5 device


lsgunth (approved):
Seems good. The second patch could have a better commit title "Fix issues" is not very informative.

kelvin-cao (approved):

xyan264:
Commit log updated. Thanks.

PR #257, opened 2021-03-26
Add Gen and HW revision info to mfg ping command output

xyan264:
Add HW revision and generation info to 'mfg ping' command output.

### DIFF
--- a/cli/mfg.c
+++ b/cli/mfg.c
@@ -83,7 +83,6 @@ static const char* phase_id_to_string(enum switchtec_boot_phase phase_id)
 static int ping(int argc, char **argv)
 {
 	int ret;
-	enum switchtec_boot_phase phase_id;
 	static struct {
 		struct switchtec_dev *dev;
 	} cfg = {};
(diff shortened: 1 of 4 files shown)


cli/mfg.c
```diff
@@ -94,14 +95,16 @@ static int ping(int argc, char **argv)
...
+	ret = switchtec_get_device_info(cfg.dev, &phase_id, NULL, &revision);
 	if (ret) {
 		switchtec_perror("mfg ping");
 		return ret;
 	}
 
 	printf("Mfg Ping: \t\tSUCCESS\n");
 	printf("Current Boot Phase: \t%s\n", phase_id_to_string(phase_id));
+	printf("Generation: \t\t%s\n", switchtec_gen_str(cfg.dev));
+	printf("HW Revision: \t\t%s\n", switchtec_rev_str(revision));
```
lsgunth, line 107: Ping seems like a poor choice of command for this.... Not even sure why the boot phase is here. Ping was meant to just be a basic MRPC test...

Why don't we add a 'switchtec info' command to dump version info and other things about the hardware?

xyan264: Agree it's a poor choice of command name... 

We already have a 'switchtec info' command that shows similar info. However, 'switchtec info' is for Main FW only. 

'switchtec mfg' submenu features commands that works for all boot phases (BL1/BL2/MAIN FW), used by customers during manufacturing process.

The 'switchtec mfg ping' command can be considered as an extension of 'switchtec info' command, for use in manufacturing process.

kelvin-cao: Yeah, it seems adding the gen/rev to 'switchtec info' is more proper. If we keep the command name 'Ping', then we keep it as simple as its name suggests, therefore the boot phase is not suitable here (we already have it in the 'mfg info').

Then we will have 'switchtec info' and 'mfg info' for the basic and extended (for mfg) info of the hardware.

xyan264: @kelvin-cao Code is updated with the following:
- add 'hw revision' to `info' command output
- delete 'boot phase' from `mfg ping` command output
Please review and comment.

kelvin-cao (approved):
Looks good.

Issue #259, opened 2021-04-08
gas write/read over serial/mmapped interface not working as expected with PAX 42100


maquefel:
Hello, i am experiencing strange behavior trying to read temperature over serial/mmapped interface:

Results in zero read with no visible error:
```
./switchtec gas write /dev/switchtec0 -a 0x0 -v 0x2 -y
./switchtec gas write /dev/switchtec0 -a 0x800 -v 0x04 -y
./switchtec gas read /dev/switchtec0 -a 0x400
000400 - 0x00000000
```
```
./switchtec gas write /dev/ttyUSB0 -a 0x0 -v 0x2 -y
./switchtec gas write /dev/ttyUSB0 -a 0x800 -v 0x04 -y
./switchtec gas read /dev/ttyUSB0 -a 0x400
000400 - 0x00000000
```
Same sequence over terminal by hand gives expected result:
```
0x00000000:004f>gaswr 0x0 0x2
gas_reg_write() success
0x00000000:0050>gaswr 0x800 0x04
gas_reg_write() success
0x00000000:0051>gasrd 0x400
gas_reg_read <0x400> [1]
0x000011bd
0x00000000:0052>
```

tested on v2.6-rc1

lsgunth:
You really shouldn't be trying to do direct gas reads and writes.  Your stepping on the kernel drivers toes. Why not just use `switchtec temp /dev/switchtec0`?

maquefel:
Hello @lsgunth, temperature was taken just as an easy reproducible example.

switchtec temp uses a different mechanism than gas read/write and i am aware of it.

So are you just saying that GAS operations should be avoided and not recommended for usage ?

lsgunth:
Yes. There are a number of ways to do custom MRPC operations safely and properly in the 'examples' folder. Using the GAS interface is not a good idea.

maquefel:
Hello @lsgunth this issue is indeed is solved by #260. Is not that i really want to use gas read/write operations, but ChipLink program that i use for PAX configuration uses them in case of PCIe configuration and UART is really slow.

PR #260, opened 2021-04-08
Add 'no MRPC' version of switchtec_open function

xyan264:
This new function opens a device without issuing MRPC command, so the contents in gas MRPC region is intact after the call.

### DIFF
--- a/inc/switchtec/switchtec.h
+++ b/inc/switchtec/switchtec.h
@@ -313,6 +313,7 @@ enum switchtec_event_id {
 /*********** Platform Functions ***********/
 
 struct switchtec_dev *switchtec_open(const char *device);
+struct switchtec_dev *switchtec_open_no_mrpc(const char *device);
 struct switchtec_dev *switchtec_open_by_path(const char *path);
 struct switchtec_dev *switchtec_open_by_index(int index);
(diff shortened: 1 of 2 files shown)


lsgunth:
Why?

xyan264:
We've seen requests to fix the 'gas read/write' command issues like PR #259 . 

The issue is caused by `switchtec_open` sending MRPC command during initialization, overwriting contents in MRPC GAS region (0x0-0x1000).

This new `switchtec_open_no_mrpc` function does not issue MRPC command, so it can be a good workaround for this issue, if needed.

lsgunth:
People should simple not do that. In fact, I think the proper thing to do is to disallow access to the MRPC region of the gas entirely.

There's no way to know if you're stepping on someone's toes if you try to do MRPC through the GAS and there are much better ways to do it. (See the examples folder). So just say no.

maquefel:
This pull request solves some issues with gas read/write - at least i could get temperature (i replaced every switchtec_open with switchtec_open_no_mrcp). But using modified tool with ChipLink hanged both host with PAX 42100 and slave connected to it. So @lsgunth is right it us a really dangerous thing. 

On the other hand is it supposed that configuration should be done only by serial or ethernet terminal (that still uses gas read/write but in other form) ? This is quite dissapointing.

PR #261, opened 2021-04-08
Implement switchtec_list_free

BertusGreeff:
switchtech_list() returns a pointer allocated by switchtec.dll calloc().  I'm using the MSVC toolchain to call switchtec.dll and there is no way to free the returned memory since calling free() from my code will attempt to free the block using a different heap.  This PR add a library function that will free the memory using the switchec.dll free().


lsgunth (approved):
Ah, ok. That makes sense to me. 

Thanks!

kelvin-cao (approved):
It seems to be a common issue, will consider this in the future API design. Thanks!

PR #263, opened 2021-04-21
Fixed invalid pointer in cli list when no devices found

mpiszczek:
This patch initialized the devices pointer in list() in cli/main.c NULL to prevent a errors when freeing the list when no devices were found.

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -137,7 +137,7 @@ int pax_handler(const char *optarg, void *value_addr,
 
 static int list(int argc, char **argv)
 {
-	struct switchtec_device_info *devices;
+	struct switchtec_device_info *devices = NULL;
 	int i, n;
 
 	static struct {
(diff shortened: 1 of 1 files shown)


xyan264:
Thanks for the input. 

I think it would be better to also change line 155 to:
``` if (n <= 0)```

**Comment by mpiszczek** [comment]
> Thanks for the input.
> 
> I think it would be better to also change line 155 to:
> ` if (n <= 0)`

**Comment by mpiszczek** [comment]
Done.

File: cli/main.c [hunk 1]
```diff
@@ -137,7 +137,7 @@ int pax_handler(const char *optarg, void *value_addr,
 
 static int list(int argc, char **argv)
 {
-	struct switchtec_device_info *devices;
+	struct switchtec_device_info *devices = NULL;
```
lsgunth, line 140: I don't understand how this changes anything... Doesn't switchtec_list() assign to devices even when there are no devices?

mpiszczek: In switchtec_list() in lib/platform.linux.c devices is not set until after scandir finds no devcies.

lsgunth: Ah, right. Thanks!

lsgunth (approved):

xyan264:
The target branch should be _devel_ instead of _master_. 

I have applied this patch to _devel_: 
     https://github.com/Microsemi/switchtec-user/commit/2a5e37819f372f7fc616bff8ddc58dd6f0cd844a
This patch will be included in the next master release.

Thanks for your contribution.

PR #264, opened 2021-04-26
Fix fw-read segmentation fault

xyan264:
Added the following checks:
- ensure the partition is not empty 
- only print the partition info if the partition is valid (avoid possible issues caused by random data on the flash)

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -1778,11 +1778,14 @@ static int fw_read(int argc, char **argv)
 		goto close_and_exit;
 	}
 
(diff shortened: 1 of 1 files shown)


lsgunth (approved):
Looks good to me, thanks.

kelvin-cao (approved):
I think some description about the issue in the commit message would be helpful. Otherwise people will not know what was observed and what is being fixed, like in this case a segmentation fault. 

Otherwise it looks good to me.

xyan264:
Commit message updated. Merged.

PR #265, opened 2021-05-04
SerDes Receiver Object Status and Equalization Status

lsgunth:
This PR implements diagnostic commands for Receiver Object Dump, Port Equalization Status and Extended Receiver Object Dump. Most of this is pretty tedious translations of commands to library interface to command line. 

I've made a quick [asciinema](https://asciinema.org/a/RS5kdLk20tZ10DIntZ5iWnupg) detailing the output of all the commands.

I've made some basic assumptions on output format but can easily make changes if there are different preferences.

I haven't been able to test any of the previous-link sub commands of Extended Receiver Object dump. The command is only documented for PAX switches but I was told they work on all variants. However, that seems like it might not be true. So do I need to flash PAX firmware to test these? If so is there anything I should be aware of when flashing that variant (I've never used it).


yungwesl:
Hey Logan,

This is really good work considering how little guidance we provided.  Some comments:
port-eq-coeff
- Rename this port-eq-txcoeff as it signifies the transmitter coefficients we are using (local being default)
- Then port-eq-txcoeff -f signifies the far end tx coefficients the link partner is using
- Then port-eq-txcoeff -prev signifies MRPC command ID 0x3A (the extended objects from the previous run) local tx coeff
- And port-eq-txcoeff -prev -f signifies MRPC command ID 0x3A (the extended objects from the previous run) remote tx coeff
- I checked the FW and PSX/PFX should support the extended objects as of build 3.60.0.49 so my only guess is that the -prev data is only there if you have linked up at least twice (otherwise there is no -prev data)

port-eq-table
- Rename this port-eq-txtable
- This MRPC is for the far end only so i think it is ok to leave it as-is (no need for -f) as you have done. 
- This command supports -prev which calls MRPC 0x3A (previous table)

port-eq-fs-lf
- rename to port-eq-txfslf 
- This command supports -prev which calls MRPC 0x3A (previous fs/lf)

recv-obj
- rename to rcvr-obj (this is our nomenclature)
- Values look good
- This command supports -prev which calls MRPC 0x3A (previous rcvr-obj)

rx-mode
- rename to rcvr-extended (if extended is too long, use rcvr-ext)
- This command dumps rx-mode as well as additional objects so I wanted to ensure it captures that it is extended data
- Can you express MODE as a hex value, this is more useful for our usage
- This command supports -prev which calls MRPC 0x3A (previous rcvr-ext)

Let me look into your issues with the "-prev".  I am pretty sure your version of FW works with this command and the PFX/PSX type should also work.

lsgunth:
Ok, sounds good. I've force pushed with all the renames and the hex value for mode, plus a couple other minor nits I noticed.

Now that you mention it, looking closer it seems the --prev is returning a proper error code indicating there is no previous data. I've added a commit to ensure this error message is printed appropriately.

Also, after rebooting I was able to get the previous data to appear. So it's all working properly now.

cli/diag.c
```diff
@@ -0,0 +1,317 @@
...
+	};
+
+	ret = diag_parse_common_cfg(argc, argv, CMD_DESC_RCVR_EXTENDED,
+				    &cfg, opts);
+	if (ret)
+		return ret;
+
+	printf("Mode and DTCLCK for physical port %d %s\n\n",
+	       cfg.port_id, cfg.prev ? "(Previous Link-Up)" : "");
+	printf("  Lane      MODE   DTCLK_5  DTCLK_8_6  DTCLK_9\n");
```
kelvin-cao, line 283: Any reason for the extra leading white space here? I don't see them in other commands.

kelvin-cao:
Some of the command names are not updated in the commit messages.

cli/diag.c
```diff
@@ -0,0 +1,317 @@
...
+	};
+
+	ret = diag_parse_common_cfg(argc, argv, CMD_DESC_RCVR_OBJ,
+				    &cfg, opts);
+	if (ret)
+		return ret;
+
+	printf("Coefficients for physical port %d %s\n\n", cfg.port_id,
+	       cfg.prev ? "(Previous Link-Up)" : "");
+	printf("Lane  CTLE  Tgt Amp  Spec DFE  DFE0 DFE1 DFE2 DFE3 DFE4 DFE5 DFE6\n");
```
kelvin-cao, line 244: The 'Tgt Amp' and 'Spec DFE' seems like 4 column header in the output. Use a '-' to connect them?

lsgunth:
Force pushed correcting the issues Kelvin noticed.

kelvin-cao (approved):
Thanks.

lsgunth:
Thanks. I have another branch in the works that depends on this PR (it adds to diag.c)... When can I expect this to be merged to devel?

kelvin-cao:
Another PR was merged which introduced a conflict. @lsgunth can you rebase to resolve it? Afterwards you can merge it yourself.

PR #266, opened 2021-05-06
Support dumping logs in TXT format 

xyan264:
**Background** 

In the previous design:
- `log-dump` dumps a binary log file
- `log-parse` parses the binary log file using a `log definition` file, which is included in the release package

This creates a maintenance issue: since different version of FW might generate different log data format, when parsing the log file using `log-parse`, user needs to ensure they are using the right version of log definition file, otherwise the parsed data could contain errors.

**Solution**
The solution is to keep a copy of the log definition files on the device, so we can dump the log definition file from the device, and use when needed. This ensures the versions of the log definition file always match those of the log data file.

In the case when user dumps log files in binary format, we add a header that contains the FW/SDK version info to the binary log file.

**This PR**
In this PR we add the following features:
- add a library function to allow user download log def file from device
- add option `-f BIN/TXT` to allow user specify the output format. In the case of TXT, we download log def file and use it to parse the binary data file
- add a header in log file to hold version information
- add version checking to ensure the log def file has the same versions as the log file


lsgunth (commented):
I very much approve of the direction this is taking. Just saw a couple small nits in the implementation.

cli/main.c
```diff
@@ -936,12 +982,43 @@ static int log_dump(int argc, char **argv)
...
+	if (cfg.format == LOG_FMT_BIN) {
+		log_def_to_use = NULL;
+	} else {
+		log_def_to_use = cfg.log_def_file;
+
+		ret = get_log_def_file(cfg.dev, cfg.type, cfg.format,
+				       &log_def_to_use);
+		if (ret < 0)
+			return ret;
+	}
```
lsgunth, line 1018: I feel like this would be slightly less awkward if written like this:

```c
if (cfg.format == LOG_FMT_BIN) {
    log_def_to_use = NULL; 
} else if (cfg.log_def_file) {
    log_def_to_use = cfg.log_def_file;
} else {
    log_def_to_use = get_log_def_file(...);
    if (!log_def_to_use)
          return -1;
}
```

xyan264: Good idea. 

Also added `INFO` print to notify user of an incorrect usage of `-f BIN` for other log types.

cli/main.c
```diff
@@ -936,12 +973,49 @@ static int log_dump(int argc, char **argv)
...
+
+	if (cfg.format == LOG_FMT_BIN) {
+		log_def_to_use = NULL;
+	} else if (cfg.log_def_file) {
+		log_def_to_use = cfg.log_def_file;
+	} else {
+		log_def_to_use = get_log_def_file(cfg.dev, cfg.type,
+						  cfg.format);
+		if (log_def_to_use == NULL)
```
lsgunth, line 1007: Small nit: the correct style is just `if (!logdef_to_use)`... Same down below near fclose().

lsgunth (approved):
Looks good to me save a small nit. Thanks!

PR #267, opened 2021-05-11
Header and version check in log-parse command

xyan264:
**Background:** 
  The previous #266 adds a header to the binary log file generated from `log-dump` command. The header includes signature and version information.

**This PREP:**
- Updates `switchtec_parse_log` function to handle the header that _might_ exist in the input log file 
- Adds version checking to `log-parse` command, with a print to remind user when the input log file and log def file versions don't match

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -1043,7 +1043,7 @@ static int log_dump(int argc, char **argv)
 static int log_parse(int argc, char **argv)
 {
 	int ret;
-
+	struct switchtec_log_file_ver_info info;
 	const struct argconfig_choice log_types[] = {
 		{"APP", SWITCHTEC_LOG_PARSE_TYPE_APP, "app log"},
 		{"MAILBOX", SWITCHTEC_LOG_PARSE_TYPE_MAILBOX, "mailbox log"},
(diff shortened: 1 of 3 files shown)


lsgunth (approved):
Looks good to me.

cli/main.c
```diff
@@ -1097,6 +1097,10 @@ static int log_parse(int argc, char **argv)
 		fprintf(stderr, "\nParsed log saved to %s.\n",
 			cfg.parsed_log_filename);
 
+	if (ret == ENOEXEC)
+		fprintf(stderr, "\nWARNING: The two input files have different version numbers.\n"
+				"The log file is parsed but the output file might contain errors.\n");
+
```
kelvin-cao, line 1103: It can be helpful to print the versions that mismatch.

xyan264: At this level there is no function to get the version numbers of log file and log-def file. 

Do you want me to add 2 functions like:
```
int switchtec_log_get_header(struct log_header *header)    //returns bin header info
int switchtec_log_def_get_header(struct log_def_header *header) //returns log-def header info
```
so we can print the versions from `header`?

I don't see many use cases for these 2 functions, but it doesn't hurt to add these.

kelvin-cao: How about add an out structure parameter to switchtec_parse_log() to pass out the versions of log and def files?

xyan264: Good idea. Code is updated. Output looks like this:
```./switchtec log-parse switchtec-ram.log oslog-def-445.dat

Parsed log saved to log.txt.

WARNING: The two input files have different version numbers.
                FW Version      SDK Version
Log file:       0x06010d0e      0x00000000
Log def file:   0x06010d0e      0x06010d0e

The log file is parsed but the output file might contain errors.```

kelvin-cao (approved):
Looks good. Thanks.

PR #269, opened 2021-05-14
Permission Table

lsgunth:
This PR implements the permission table MRPC and a list-mrpc's command in the diag option.

It's pretty straightforward, however one odd thing I noticed is it seems the firmware always advertises any MRPCs that it doesn't no about as available. Which is odd. For example my switch reports SECURITY_CONFIG_GET_GEN5 is available when I expect it is not (seeing it's not gen5). But also MRPCs from 0x10e to 0x13f. There's nothing the sotfware can really do about this though.


kelvin-cao:
I guess the permission table doesn't actually reflect the availability of commands, so a permitted MRPC command (ID) may not actually be available in the running firmware. I'll need to confirm this.

kelvin-cao:
I got the answer that firmware will only return true for the permission of a MRPC when there's a valid MRPC handle for it in the firmware. @lsgunth, what did you see for MRPC 0x10e to 0x13f?

yungwesl:
The purpose of the permission table is for customers to establish which commands are available to the MRPC interface. This is specifically there for customers to check that any MRPC that have blocked via the OTP Settings block or via the config file permission tables are implemented correctly. 

Logan, I believe this command can also be run via the I2C interface and may preset a different list of permitted commands since BL1 does not have the same visibility as Main FW.  I assume you do not have I2C access to the TWI recovery interface so Kelvin can you help me verify this portion of the code?

kelvin-cao:
@yungwesl sure, I can do that.

lsgunth:
> Logan, I believe this command can also be run via the I2C interface and may preset a different list of permitted commands since BL1 does not have the same visibility as Main FW. I assume you do not have I2C access to the TWI recovery interface so Kelvin can you help me verify this portion of the code?

That was my understanding as well. The code *should* work with I2C mode and just return different results. But testing that is a good idea.

kelvin-cao:
Did some test over both in-band and TWI interface with log [diag list-mrpc.txt](https://github.com/Microsemi/switchtec-user/files/6511815/diag.list-mrpc.txt).

I noticed that
1. BL1 and BL2 doesn't support this command, which I think is expected.
2. 0x10c and 0x10d is not shown in the TWI run.
3. In-band shows the UNKNOWN lines but TWI doesn't.

yungwesl:
What is the difference between 
kelvin@harpoon:~/git/switchtec-user$ sudo ./switchtec diag list-mrpc 0

and

kelvin@harpoon:~/git/switchtec-user$ sudo ./switchtec diag list-mrpc -a 0

Is the -a intended to list all commands?  How is this different than no argument?  Is -a a master list of all commands you can use it as the master list to compare the non (-a) list.

lsgunth:
> Is the -a intended to list all commands? How is this different than no argument? Is -a a master list of all commands you can use it as the master list to compare the non (-a) list.

-a will list commands that switchtec-user doesn't know about (so it will show a whole mess of "unknown commands").

Without -a it won't print anything for command IDs it has no information on.

yungwesl:
Thanks Logan,

- So the normal run lists all MRPC that switchtec user can use based on the commands that it supports
- -a will list all MRPC commands in the switch permission table

In all cases, @kelvin-cao switchtec-user will not list commands that have been explicitly blocked by either OTP or the MRPC command mask in the config file.

kelvin-cao:
So it seems firmware is returning true for the permission of MRPC that does not actually exist. Will check with firmware. @yungwesl are you satisfied with the command line? If nothing you think need to change, I'll approve it.

lsgunth:
>   So the normal run lists all MRPC that switchtec user can use based on the commands that it supports
    -a will list all MRPC commands in the switch permission table

Not quite.

A normal run will list all commands in the partition table that switchtec-user knows about (not necessarily supports with a command). I've added descriptions of all commands in the documentation.

A -a run will do exactly the same, except for every command in the partition table that switchtec-user does't know about it prints "UNKNOWN".

Here's an excert with -a:

```
  0x02b  PART_INFO                  Get Partition Info
  0x02c  PCIE_GEN_1_2_DUMP          PCIe Gen1/2 Port Tuning Dump
  0x02d  PCIE_GEN_1_2_TUNE          PCIe Gen1/2 Port Tuning
  0x02e  UNKNOWN                    Unknown MRPC Command
  0x02f  EYE_OBSERVE                Eye Observation Monitor
  0x030  RCVR_OBJ_DUMP              Receiver Object Dump
  0x031  UNKNOWN                    Unknown MRPC Command
  0x032  PORT_EQ_STATUS             Port Equalization Status
  0x033  PORT_EQ_CTRL               Port Equalization Control
  0x035  MRPC_ERR_INJ               MRPC Link Error Injection
  0x036  DEV_INFO_GET               Device Info Get
  0x037  MRPC_PERM_TABLE_GET        MRPC Permission Table Get
  0x038  CROSS_HAIR                 Cross Hair
  0x039  RECV_DETECT_STATUS         Receiver Detect Status Get
  0x03a  EXT_RCVR_OBJ_DUMP          Extended Receiver Object Dump
  0x03b  LOG_DEF_GET                Read Application Log
  0x03c  SECURITY_CONFIG_GET_EXT    Secure Configuration Get Extended
  0x041  ECHO                       Echo
  0x100  I2C_TWI_PING               I2C/TWI Ping
  0x101  SECURITY_CONFIG_GET        Secure Configuration Get
  0x102  SECURITY_CONFIG_SET        Secure Configuration Set
  0x103  KMSK_ENTRY_SET             Public Key Entry Hash Key Set
  0x104  SECURE_STATE_SET           Secure State Set
  0x105  ACT_IMG_IDX_GET            Firmware Active Image Index Get
  0x106  ACT_IMG_IDX_SET            Firmware Active Image Index Select
  0x107  FW_TX                      Image Transfer and Execution
  0x108  MAILBOX_GET                Mailbox Log Get
  0x109  SN_VER_GET                 Chip Serial Number and Secure Versions
  0x10a  DBG_UNLOCK                 Resource Unlock
  0x10b  BOOTUP_RESUME              Bootup Resume
  0x10c  SECURITY_CONFIG_GET_GEN5   Secure Configuration Get (Gen5)
  0x10d  SECURITY_CONFIG_SET_GEN5   Secure Configuration Set (Gen5)
  0x10e  UNKNOWN                    Unknown MRPC Command
  0x10f  UNKNOWN                    Unknown MRPC Command
  0x110  UNKNOWN                    Unknown MRPC Command
  0x111  UNKNOWN                    Unknown MRPC Command
```

Without -a, all the UNKNOWNS are hidden.

yungwesl:
Functionally, this code looks good.  The results from -a are a Switch FW problem, not a switchtec_user problem so I am ok with this implementation.  Let me quickly review this output with our lead customer to ensure it meets their requirements.

kelvin-cao:
@lsgunth confirmed with firmware on the following facts.

1. a firmware bug cause the Gen5 MRPC shown on the Gen4 devices.  (for MRPC 0x10e   and above they are also likely caused by the same bug)
2. there are some internal MRPC (not described in dev spec) like `0x024  UNKNOWN                    Unknown MRPC Command`, which we may need to handle differently, like showing as MCHP_RESERVED instead of UNKNOWN to separate them from the MRPCs that are public but unknown to switchtec-user. What do you think? I can provide a list of the internal MRPC when confirmed with firmware.

lsgunth:
Yes, I can add the reserved to the list if you let me know what they are.

kelvin-cao:
> 
> 
> Yes, I can add the reserved to the list if you let me know what they are.

0xe, 0x24, 0x31 are the internal MRPC. Thanks.

lsgunth:
Ok, I've added those values as reserved MRPCs. I also rebased on current devel.

kelvin-cao (approved):
Thanks.

lsgunth:
I've got the next small PR in the queue but it depends on this PR. Can we merge this to devel soon?

kelvin-cao:
Let me check with @yungwesl.

PR #271, opened 2021-05-17
Add SEEPROM info in 'fw-info'

xyan264:
Setting stored in SEEPROM can be considered as a special partition, and the information can be shown in 'fw-info' command. 

This PREP adds SEEPROM info to 'fw-info' output. The output looks like this:
```
Active Partitions: 
  MAP   Version: 6.01 B00C      CRC: 5d821783         (Running) 
  KEY   Version: 6.01 B00C      CRC: b97a337c         (Running) 
  BL2   Version: 6.01 B00C      CRC: cab4e15e   (RO)  (Running) 
  IMG   Version: 6.01 B00C      CRC: 65b9abdf         (Running) 
  CFG   Version: 6.01 B00C      CRC: 9497b951         (Running) 
Inactive Partitions: 
  MAP   Version: 6.01 B00C      CRC: 5d821783 
  BL2   Version: 6.01 B00C      CRC: 96284762 
  IMG   Version: 6.01 B00C      CRC: 9527c150 
  CFG   Version: 6.01 B00E      CRC: 6fb2780b 
Other Images:
  SEEPROM        Version: 6.01 B00C      CRC: 77b5b09d
```


kelvin-cao (approved):
Looks good.

lsgunth (approved):
Looks good to me!

xyan264:
This code has been updated and discussion will continue in PR #272

cli/main.c
```diff
@@ -1458,6 +1458,17 @@ static void print_fw_part_line(const char *tag,
...
 
+static void print_seeprom_part_line(const char *tag,
+				    struct switchtec_fw_image_info *inf)
+{
+	if (!inf)
+		return;
+
+	printf("  %-4s\tVersion: %-8s\tCRC: %08lx\t%s\n",
+	       tag, inf->version, inf->image_crc,
+	       inf->valid ? "" : "  (Invalid)");
```
lsgunth, line 1469: Why not just use print_fw_part_line()? We know read_only, running and redundant should be false don't we?

lib/fw.c
```diff
@@ -1067,6 +1070,25 @@ static int switchtec_fw_part_info_gen4(struct switchtec_dev *dev,
 	case SWITCHTEC_FW_PART_ID_G4_NVLOG:
 		part_info = &all->nvlog;
 		break;
+	case SWITCHTEC_FW_PART_ID_G4_SEEPROM:
+		if (switchtec_gen(dev) <= SWITCHTEC_GEN4)
+			return 0;
```
lsgunth, line 1075: Wait.... this is only applicable to gen5?? It might be better written as < SWITCHTEC_GEN5 then...

And shouldn't we be setting all the inf-> things to invalid values or something?

xyan264: The caller function already initializes all fields to 0 (valid = false, active=false) so it should suffice.

cli/main.c
xyan264: Good idea. Updated.

lib/fw.c
lsgunth: Oh, then maybe we don't need to set the values below to zero.... It's a bit confusing in that way.

lsgunth (approved):
Looks good to me, again

PR #272, opened 2021-05-25
Add SEEPROM info to 'fw-info' (updated)

xyan264:
This is an updated version of #271 , with the following changes:
- A separate API function is provided to get SEEPROM info, this makes error handling much easier
- Add error code and error message for SEEPROM access errors

Below is the same info as #271 , copied here for completeness:
 -- ---
Setting stored in SEEPROM can be considered as a special partition, and the information can be shown in 'fw-info' command. 

This PREP adds SEEPROM info to 'fw-info' output. 

The output looks like this:
```
Active Partitions: 
  MAP   Version: 6.01 B00C      CRC: 5d821783         (Running) 
  KEY   Version: 6.01 B00C      CRC: b97a337c         (Running) 
  BL2   Version: 6.01 B00C      CRC: cab4e15e   (RO)  (Running) 
  IMG   Version: 6.01 B00C      CRC: 65b9abdf         (Running) 
  CFG   Version: 6.01 B00C      CRC: 9497b951         (Running) 
Inactive Partitions: 
  MAP   Version: 6.01 B00C      CRC: 5d821783 
  BL2   Version: 6.01 B00C      CRC: 96284762 
  IMG   Version: 6.01 B00C      CRC: 9527c150 
  CFG   Version: 6.01 B00E      CRC: 6fb2780b 
Other Images:
  SEEPROM        Version: 6.01 B00C      CRC: 77b5b09d
```


lsgunth:
#271 looks a lot cleaner and adds less ugly special case logic to the CLI... I don't see how this is an improvement...

xyan264:
SEEPROM partition is slightly different, compared to other partitions:
1. It's optional: only available on GEN5; not on GEN3/GEN4
2. Hardware may return one of several possible error codes: HW not accessible, or data stored is not valid (possibly more error codes in the foreseeable future)

In the previous PR #271 , the `get_seeprom_info` function is integrated in `switchtec_fw_part_summary`. If it returns error, then `switchtec_fw_part_summary` returns, and none of the other partitions info is shown. This would be quite frustrating to end user, as SEEPROM is only an optional device. 

Using a separate API to get SEEPROM info gives CLI the flexibility to handle the error code properly. CLI can choose to print the error message, as done in this PR, without affecting the display of other partition info. Or CLI can simply ignore the error and skip printing SEEPROM info.

lsgunth:
So fix it so that fw_part_summary returns sensibly with an invalid seeprom section if get_seeprom_info() fails. We really don't need an ugly new API that callers have to deal with. 

I also find it particularly ugly to have generation based ifs in the CLI.

xyan264:
The previous approach #271  is preferred. This PR is closed.

PR #273, opened 2021-06-04
REFCLK Output Control Command

lsgunth:
This is a simple PR to add a command to control the refclk output (SOW 3.1.8).

The command is quite straight forward:

switchtec diag refclk /dev/switchtec0 (--enable | --disable) --stack STACK


yungwesl:
Looks good Logan

I approve.

kelvin-cao:
Will there be a command or option to enable the status checking of refclk output?

lsgunth:
There doesn't appear to be an MRPC command that allows you to query the status. If there was I could add it easily.

kelvin-cao (approved):
Looks good to me. Thanks.

yungwesl:
Oh wow REFCLK_S control doesn't have a readback MRPC subcommand.  That seems like bad practice :)

Let me ask if it can be added as an enhancement for future. For now we can move ahead as-is.  We can only use command success as being the "confirmation"

Wes

henryliuhl:
A new REFCLK_S Status sub-command (ID 0x02) has been added into Gen5 PCIe Switch firmware. It can be used to read the REFCLK_S enabling status of available stacks.

PR #277, opened 2021-07-21
Add 'log overflow

xyan264:
To allow user to get 'log overflow' flag, new MRPC subcommand 0xd/0xe have been added to 'log dump' command.

This PR uses the new subcommand to dump log data and saves the flags in a structure. CLI can check the fields in the structure and print warnings if necessary.

Unfortunately it is hard to further divide the patch, since all the changes are required to compile successfully.

The key changes are in function `log_ram_flash_to_file`.


lib/switchtec.c
```diff
@@ -1333,36 +1333,64 @@ static int log_c_to_file(struct switchtec_dev *dev, int sub_cmd_id, int fd)
...
+{
+	int ret;
+	int subcmd;
+
+	if (switchtec_is_gen5(dev)) {
+		subcmd = subcmd1;
+		return log_a_to_file(dev, subcmd, fd, log_def_file,
+				     info);
+	}
+	else {
```
lsgunth, line 1349: else on same line as close brace

inc/switchtec/switchtec.h
```diff
@@ -391,11 +394,12 @@ void switchtec_perror(const char *str);
 int switchtec_log_to_file(struct switchtec_dev *dev,
 			  enum switchtec_log_type type,
 			  int fd,
-			  FILE *log_def_file);
+			  FILE *log_def_file,
+			  struct switchtec_log_file_info *info);
```
lsgunth, line 398: This seems really odd given that the info struct is just being used to return warning flags and the _version fields aren't even set by this call. Why not just return some positive flags or something on success with warnings?

lib/switchtec.c
```diff
@@ -1333,36 +1333,64 @@ static int log_c_to_file(struct switchtec_dev *dev, int sub_cmd_id, int fd)
 	return 0;
 }
 
+static int log_ram_flash_to_file(struct switchtec_dev *dev,
+				 int subcmd1, int subcmd2, int subcmd3,
```
lsgunth, line 1337: Some better names for subcmd1, subcmd2 and subcmd3 would be appropriate.... subcmd_gen5, subcmd_flags, subcmd_legacy... or some such.

inc/switchtec/log.h
```diff
@@ -46,7 +46,8 @@ struct log_a_retr_result {
 	struct log_a_retr_hdr {
 		uint8_t sub_cmd_id;
 		uint8_t from_end;
-		uint8_t reserved1[2];
+		uint8_t overflow;
+		uint8_t reserved1;
```
lsgunth, line 50: Push this into the previous commit and probably add a comment to say overflow only applies with the _FLAGS subcmd....

Though I question why a new subcmd ID was even necessary in the firmware? Can't it just start setting the reserved field or not? Presumably firmware that doesn't support will just leave the reserved field zeroed.

lib/switchtec.c
```diff
@@ -1333,36 +1333,64 @@ static int log_c_to_file(struct switchtec_dev *dev, int sub_cmd_id, int fd)
...
+				 int fd, FILE *log_def_file,
+				 struct switchtec_log_file_info *info)
+{
+	int ret;
+	int subcmd;
```
lsgunth, line 1342: Probably don't need this extra variable....

lib/switchtec.c
```diff
@@ -1333,36 +1333,64 @@ static int log_c_to_file(struct switchtec_dev *dev, int sub_cmd_id, int fd)
...
+	if (switchtec_is_gen5(dev)) {
+		subcmd = subcmd1;
+		return log_a_to_file(dev, subcmd, fd, log_def_file,
+				     info);
+	}
+	else {
+		subcmd = subcmd2;
+		ret = log_a_to_file(dev, subcmd, fd, log_def_file,
+				    info);
+		if (ret > 0) {
```
lsgunth, line 1353: Shouldn't  this be < 0? Or rather we should be checking for the error specifically saying that the subcmd was not accepeted.

Though, the error handling in log_a_to_file seems to be a mess..... Would be better to follow the convention of returning -1 on error and setting errno.

inc/switchtec/log.h
xyan264: Moved to previous commit along with comments.

Yes it seems overkill to add new subcommands for this purpose, but it's the hardware convention so we will follow the implementation.

inc/switchtec/switchtec.h
xyan264: Yes I forgot to set the version fields in the previous patch. Update now.

lib/switchtec.c
xyan264: Yes, in fact we should check for all errors, so it's updated to `if(ret)` now.

lib/switchtec.c
```diff
@@ -1333,36 +1340,58 @@ static int log_c_to_file(struct switchtec_dev *dev, int sub_cmd_id, int fd)
...
+{
+	int ret;
+
+	if (switchtec_is_gen5(dev)) {
+		return log_a_to_file(dev, gen5_cmd, fd, log_def_file,
+				     info);
+	} else {
+		ret = log_a_to_file(dev, gen4_cmd, fd, log_def_file,
+				    info);
+		if (!ret)
```
lsgunth, line 1356: I still think we should only try the legacy command if this failure indicates a bad MRPC command.

lib/switchtec.c
```diff
@@ -1333,36 +1340,58 @@ static int log_c_to_file(struct switchtec_dev *dev, int sub_cmd_id, int fd)
...
  * @param[in]  log_def_file - Log definition file
+ * @param[out] info         - Log file information
  * @return 0 on success, error code on failure
  */
 int switchtec_log_to_file(struct switchtec_dev *dev,
 			  enum switchtec_log_type type,
 			  int fd,
-			  FILE *log_def_file)
+			  FILE *log_def_file,
+			  struct switchtec_log_file_info *info)
```
lsgunth, line 1377: Certainly this can be wrapped better.

inc/switchtec/switchtec.h
```diff
@@ -210,11 +210,14 @@ enum switchtec_log_parse_type {
...
  */
-struct switchtec_log_file_ver_info {
+struct switchtec_log_file_info {
 	unsigned int log_fw_version;
 	unsigned int log_sdk_version;
 	unsigned int def_fw_version;
 	unsigned int def_sdk_version;
+	bool version_mismatch;
+	bool overflow;
+	bool rsvd[6];
```
lsgunth, line 220: Do we need these reserved bools? Seems we can just expand this structure if we need to...

lib/switchtec.c
xyan264: What do you suggest?

lsgunth: Well log_def_file can easily be on the same line as fd....

But I'll often also use two tab indents instead of aligning to the open parenthesis to avoid too many lines (which is valid Linux style):

```
int switchtec_log_to_file(struct switchtec_dev *dev,
		enum switchtec_log_type type, int fd, FILE *log_def_file,
		struct switchtec_log_file_info *info)
```

lib/switchtec.c
xyan264: Somehow hardware gives random error number if this new subcommand is not supported. So checking for all errors is the safe way. 

I've added comments to reflect this limitation.

lsgunth: What? That seems like hardware is really broken if it returns a random error number! That needs to be fixed ASAP.

But this is not safe. If it fails for some other legitimate reason then it falls back on a command that may or may not exist, then the reason for the failure is masked.... That is not a good idea.

xyan264: Sorry about the confusion, by 'random' I mean the error code returned is not the expected 'Invalid subcommand' error code, not that it returns a different number on each run. 

Yes there could be other issues masked if we simply check for non-zero. 

I have updated the code to check for the exact error code returned, and updated the comments.

```
		/* somehow hardware returns ERR_LOGC_PORT_ARDY_BIND
		 * instead of ERR_SUBCMD_INVALID if this subcommand
		 * is not supported, so we fall back to legacy
		 * subcommand on ERR_LOGC_PORT_ARDY_BIND error */
```

lib/switchtec.c
```diff
@@ -1333,36 +1340,62 @@ static int log_c_to_file(struct switchtec_dev *dev, int sub_cmd_id, int fd)
...
+	} else {
+		ret = log_a_to_file(dev, gen4_cmd, fd, log_def_file,
+				    info);
+
+		/* somehow hardware returns ERR_LOGC_PORT_ARDY_BIND
+		 * instead of ERR_SUBCMD_INVALID if this subcommand
+		 * is not supported, so we fall back to legacy
+		 * subcommand on ERR_LOGC_PORT_ARDY_BIND error */
+		if (ret > 0 &&
+		    ERRNO_MRPC(errno) == ERR_LOGC_PORT_ARDY_BIND)
```
lsgunth, line 1362: Maybe also check for ERR_SUBCMD_INVALID in case firmware fixes this bug?

xyan264: Good idea. Updated.

kelvin-cao (approved):
Looks good to me. Thanks.

lsgunth (approved):
Looks good to me.

PR #278, opened 2021-08-12
Fixed Windows error message with FW toggle command

Chetana2791:
On windows, even when FW toggle succeeds, the error message is still seen as function call is made to platform_strerror().
It results in the following message: **The switchtec utility will show “firmware toggle: A device attached to the system is not functioning”.**

This is a false alarm.  To fix this issue, when the error number is 0, instead of making a function call to platform_strerror(), a 'Success' message is returned.

Code compilation details:
esdapps@esdapps-rome-2:~/error_msg/switchtec-user$ ./configure
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables...
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking for windres... no
checking for initscr in -lncurses... yes
checking how to run the C preprocessor... gcc -E
checking for grep that handles long lines and -e... /bin/grep
checking for egrep... /bin/grep -E
checking for ANSI C header files... yes
checking for sys/types.h... yes
checking for sys/stat.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for memory.h... yes
checking for strings.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for unistd.h... yes
checking curses.h usability... yes
checking curses.h presence... yes
checking for curses.h... yes
checking ncurses/curses.h usability... no
checking ncurses/curses.h presence... no
checking for ncurses/curses.h... no
checking openssl/pem.h usability... yes
checking openssl/pem.h presence... yes
checking for openssl/pem.h... yes
checking for RSA_free in -lcrypto... yes
checking whether RSA_get0_key is declared... yes
configure: creating ./config.status
config.status: creating Makefile
config.status: creating config.h
esdapps@esdapps-rome-2:~/error_msg/switchtec-user$ make
  VER   2.5
  CC    lib/mrpc.c
  CC    lib/fabric.c
  CC    lib/crc.c
  CC    lib/pmon.c
  CC    lib/fw.c
  CC    lib/gas_mrpc.c
  CC    lib/switchtec.c
  CC    lib/events.c
  CC    lib/diag.c
  CC    lib/mfg.c
  CC    lib/platform/platform.c
  CC    lib/platform/windows.c
  CC    lib/platform/linux-i2c.c
  CC    lib/platform/linux-uart.c
  CC    lib/platform/gasops.c
  CC    lib/platform/linux.c
  CC    lib/platform/linux-eth.c
  AR    libswitchtec.a
  LD    libswitchtec.so
  CC    cli/suffix.c
  CC    cli/commands.c
  CC    cli/gui.c
  CC    cli/progress.c
  CC    cli/main.c
  CC    cli/gas.c
  CC    cli/diag.c
  CC    cli/mfg.c
  CC    cli/argconfig.c
  CC    cli/fabric.c
  LD    switchtec
  CC    examples/temp.c
  LD    examples/temp
rm examples/temp.o
esdapps@esdapps-rome-2:~/error_msg/switchtec-user$

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -1807,7 +1807,10 @@ static int fw_toggle(int argc, char **argv)
 	printf("\n");
 
 	errno = err;
-	switchtec_perror("firmware toggle");
+	if (errno)
+		switchtec_perror("firmware toggle");
+	else
+		printf("firmware toggle: Success\n");
 
 	return ret;
 }


lsgunth:
I don't think this is the right solution as it disables all errors that might result from GetLastError().

We probably need to ensure we don't call switchtec_perror() if no error occurred.

Chetana2791:
Hi Logan,
I have made modifications to the fw_toggle() function. So when there is no error, it does not have to make a function call to switchtec_perror(). Instead, it prints a success message to the CLI screen keeping up with backwards compatibility.

lib/switchtec.c
```diff
@@ -615,7 +615,7 @@ const char *switchtec_strerror(void)
 		if (errno)
 			return strerror(errno);
 		else
-			return "Success";
+			return platform_strerror();
```
lsgunth: This part of the patch has to be removed.

lsgunth (commented):
Looks good once we remove the old wrong change.

Also it would be appreciated if a proper commit message could be added. The proper format of a commit message should be:

```
prefix: short one line title

Paragraph describing what the change is and why it was made.
```

Something like this for the commit:

```
cli: fix fw_toggle success message on windows

On windows, even on success, the fw_toggle command prints an error. This is because 
it is invalid to access errno (or call switchtec_perror()) unless a function actually returns
an error.  So print a specific success message if 'err' is not set.
```

Chetana2791:
Thanks Logan. Will make the required changes.

lsgunth (approved):
Looks good to me, thanks!

PR #280, opened 2021-08-16
Crosshair Support

lsgunth:
Building on the eye monitor this PR adds support for obtaining crosshairs.

The crosshair CLI is very similar to the eye CLI, just the command that's executed and the output is different. Both the eye command and crosshair command can optionally plot eachother's output superimposed.

With the curses mode, the crosshair command displays and draws the crosshair as it is measured. For an example, see the following asciinema:

https://asciinema.org/a/fZliEy7wN2xjDB6u4de1NJT5n

It displays the W2H calculation in the status line at the bottom when complete.

Like the eye command it also supports outputting a CSV file or simple text display.

It also supports the --all command which will measure all ports simultaneously and output a CSV file per active port.


yungwesl:
Hey Logan,

Very cool.  I have some questions for clarity:
1. W2H is defined by our spec as a capture of the Width at y=0 and height@x=28 and x=36 with a dwell time of 200ms.  By this measure it is not valid to draw a W2H on a normal eye capture with a dwell of 10ms for example. So if possible can we express the dwell time somewhere to ensure someone does not accidentally plot a W2H on data that is not set to the correct dwell? 
2. In the asciinema I noticed that the x=28 and x=36 vertical openings dont' go to the first non-zero enry
image
3. If you run this command without pre-loading a CSV does it draw the whole eye or only the crosshair (which is faster)?

lsgunth:
1. The W2H is calculated only from the crosshair data not the eye diagram (which is really just cosmetic here). I don't think there's a way to set the dwell time of the crosshair command so I assume it's always 200ms.

2. I suspect this is because of the difference of dwell times between the crosshair and eye diagram. As the eye diagram is an optional cosmetic feature, I wouldn't worry about it.

3. Yes, if you don't preload an eye CSV it only draws the crosshair.

yungwesl:
Hey Logan,

The cosmetics may look nice but they are confusing.  I'd prefer to have crosshair display what it is actually capturing and have a separate option to overlay a crosshair on top of an EOM.  

I am concerned someone will misunderstand the overlay view as the dwell time of the eye and the dwell time of the crosshairs are not the same and their effective bit error rate is different.  

It seems like the code is configured to allow for plotting of the pure crosshair so the cosmetic view has to be done explicitly so i have two options for you:

1. We print the dwell time in both the crosshair and the Eye capture plots so it is always clear 
2. we just dont allow plotting this overlay

Do you have a preference? 

Thanks
Wes

lsgunth:
Well I think the overlay is worth it even if the dwell times don't match.

We can print the dwell time for the eye capture (which I guess we'll need to add to the eye csv).... But how do we know the dwell time of the crosshair (is it fixed?).

yungwesl:
Yes, the crosshair has a fixed dwell value of 200ms we can assume this in your implementation.

lsgunth:
Force pushed with the dwell time in the status bar. If the eye CSV is used, it shows it's dwell time separately.

yungwesl:
Looks good Logan.

I like the real-time plot on the crosshair as well!
I ran some tests in text mode, csv, and ncurses.  No issues seen so far.

PR #281, opened 2021-08-27
Fix segmentation fault in 'status' command

xyan264:
Some port numbers are not available on Switchtec devices. These are identified with '0' in the available port bitmap.
`switchtec_status` only parses information for ports that are available.

### DIFF
--- a/lib/switchtec.c
+++ b/lib/switchtec.c
@@ -492,11 +492,17 @@ int switchtec_status(struct switchtec_dev *dev,
 	int i, p;
 	int nr_ports = 0;
 	struct switchtec_status *s;
+	int max_ports;
 
 	if (!status) {
 		errno = EINVAL;
 		return -errno;
 	}
+
+	if (switchtec_is_gen5(dev))
+		max_ports = 60;
+	else
+		max_ports = 52;
 
 	struct {
 		uint8_t phys_port_id;
@@ -510,7 +516,7 @@ int switchtec_status(struct switchtec_dev *dev,
 		uint16_t LTSSM;
 		uint8_t lane_reversal;
 		uint8_t first_act_lane;
-	} ports[SWITCHTEC_MAX_PORTS];
+	} ports[max_ports];
 
 	ret = switchtec_cmd(dev, MRPC_LNKSTAT, &port_bitmap, sizeof(port_bitmap),
 			    ports, sizeof(ports));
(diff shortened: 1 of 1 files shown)


lib/switchtec.c
```diff
@@ -517,10 +518,16 @@ int switchtec_status(struct switchtec_dev *dev,
...
 
+	if (switchtec_is_gen5(dev))
+		avail_port_bitmap = 0x0F00FFFFFFFFFFFF;
+	else
+		avail_port_bitmap = 0x000FFFFFFFFFFFFF;
 
 	for (i = 0; i < SWITCHTEC_MAX_PORTS; i++) {
-		if ((ports[i].stk_id >> 4) > SWITCHTEC_MAX_STACKS)
+		if ((ports[i].stk_id >> 4) > SWITCHTEC_MAX_STACKS ||
+		    !((avail_port_bitmap >> i) & 1))
```
lsgunth, line 528: Why is this even necessary? Why does the previous check not correctly filter out missing ports? What is the actual problem you are fixing?

On my system 'i' is not the port ID so this doesn't even seem correct.

xyan264:
I'm trying to fix the segmentation fault issue of `status` command, then I realize the above fix does no good either.

The root cause is we ask Firmware for 60 entries of port array data, but Gen4 Firmware can only return 52 valid entries, and the other 8 entries contain random data. 

A full solution requires changes in both firmware and software, which will be implemented when firmware is ready. 

Right now we are preparing a release for Gen4, so I suggest we revert the commit that changes MAX_PORT from 48 to 60 (commit 2bf2c880fb2bf4d6ad9ddec70265fbc32b6ea35a). 

@lsgunth  what do you think?

lsgunth:
I assume the random data is because it is uninitialized going into the MRPC command? So this would probably be an easy solution:



```diff
diff --git a/lib/switchtec.c b/lib/switchtec.c
index 11d3ce2310cb..7debb8834313 100644
--- a/lib/switchtec.c
+++ b/lib/switchtec.c
@@ -512,6 +512,8 @@ int switchtec_status(struct switchtec_dev *dev,
                uint8_t first_act_lane;
        } ports[SWITCHTEC_MAX_PORTS];
 
+       memset(ports, 0xFF, sizeof(ports));
+
        ret = switchtec_cmd(dev, MRPC_LNKSTAT, &port_bitmap, sizeof(port_bitmap),
                            ports, sizeof(ports));
        if (ret)

```

xyan264:
`memset` didn't help either. FW will overwrite the `0xff` with invalid data.

I have added a Gen4/5 check in the code, this fixes the issue. Please take a look.

lsgunth (approved):
Looks good to me, thanks.

kelvin-cao (approved):

PR #283, opened 2021-09-08
Fix timestamp calculation in log parsing

xyan264:
Different generation of  device uses different factor to calculate timestamp value. The previous factor 10 is for Gen3 device, while Gen4/5 device uses factor 8.33.

In the case of offline parsing with `log-parse`, the device generation info may or may not be present in the log file header. We add option `-g` to allow user specify device generation, so that the timestamp can be parsed correctly. 

If the log file header already contains device generation info, then the `-g` option is ignored and a message is printed.

### DIFF
--- a/lib/switchtec.c
+++ b/lib/switchtec.c
@@ -1027,15 +1027,17 @@ err_free_log_defs:
  * @param[in] defs           - log definitions
  * @param[in] log_type       - log type
  * @param[in] log_file	     - log output file
+ * @param[in] ts_factor	     - timestamp conversion factor
  * @return 0 on success, negative value on failure
  */
 static int write_parsed_log(struct log_a_data log_data[],
 			    size_t count, int init_entry_idx,
 			    struct log_defs *defs,
 			    enum switchtec_log_parse_type log_type,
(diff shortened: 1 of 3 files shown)


lib/switchtec.c
```diff
@@ -1553,8 +1577,24 @@ int switchtec_parse_log(FILE *bin_log_file, FILE *log_def_file,
 	/* parse each log entry */
 	while (fread(&log_data, sizeof(struct log_a_data), 1,
 		     bin_log_file) == 1) {
+		fw_version = fw_version_log? fw_version_log:fw_version_def;
```
lsgunth, line 1580: Generally prefer spaces around the ? and : in the ternary operator.

But I also try to avoid the ternary operator except in rare cases.

I might just do:

```
if (fw_version_log)
    gen_file = switchtec_fw_version_to_gen(fw_version_log);
else
    gen_file = switchtec_fw_version_to_gen(fw_version_def);
```

lsgunth (approved):
One nit, otherwise it looks good to me.

kelvin-cao (approved):
Looks good to me. Thanks.

PR #285, opened 2021-09-10
Fix mailbox log def file parsing issue

xyan264:
Gen5 mailbox log def file contains a header which causes issue in previous parsing code.

Add code to skip comments in log def file.


lsgunth (approved):
Looks good to me.

kelvin-cao (approved):
Looks good to me. Thanks.

PR #287, opened 2021-09-13
platform/windows: set errno when MRPC command returns failure

lsgunth:
switchtec_cmd() is expected to set the errno to an MRPC code on
exit. However this was not done on the windows platform.

Thus MRPC errors returned zero instead of the appropriate error message.


kelvin-cao (approved):
Looks good. Thanks.

PR #288, opened 2021-09-13
New Compiler Bugs

lsgunth:
Noticed a couple of bugs that warnings in the newer mingw compilers generated.


kelvin-cao (approved):
Looks good. Thanks!

PR #289, opened 2021-09-14
Fix missing ncurses.h header issue in Windows build

xyan264:
Use the same header include method as in `gui.c`.


lsgunth (approved):
Right, thanks. Looks good to me.

PR #290, opened 2021-10-09
Fix memory overflow of command 'status --help'

kelvin-cao:
The variable output_fmt_choices didn't end up with a NULL element as
boundary so that a memory overflow happened during the argument parsing.

This was only observed on an ARM system.


lsgunth (approved):
Nice catch, thanks.

PR #292, opened 2021-10-29
Fix number of lanes with port-eq-txcoeff

lsgunth:
The firmware for the eq_tx_coeff reports the maximum lane ID, not the
number of lanes and thus was reporting one too few lanes.

To fix this add one to the maximum lane ID to properly return the
number of lanes.

Reported by Rakesh Vitta


kelvin-cao (approved):
Looks good. Thanks.

PR #295, opened 2021-11-23
Fix 'mfg info' command for new Gen5 device

xyan264:
The new Gen5 (Rev-B) device works differently compared to previous Gen4 device:
- Uses new MRPC command to read security settings & security versions
- Adds attestation feature and related configuration bits
- Adds extra OTP status bits
- Supports up to 10 KMSK entries (compared to 4 on Gen4 devices)
- Adds 'debug mode valid' flag


inc/switchtec/mfg.h
```diff
@@ -105,6 +119,9 @@ struct switchtec_security_cfg_state {
 
 	bool otp_valid;
 	struct switchtec_security_cfg_otp_region otp;
+
+	bool attestation_supported;
```
lsgunth, line 163: Why not just have another value in attestation_mode indicating the hardware doesn't support it? Instead of another bool.

lib/mfg.c
```diff
@@ -392,6 +399,26 @@ static int security_config_get_gen5(struct switchtec_dev *dev,
 	memcpy(state->public_key, reply.public_key,
 	       state->public_key_num * SWITCHTEC_KMSK_LEN);
 
+	state->attestation_supported = true;
+	state->attn_state.attestation_mode =
+		(reply.cfg >> SWITCHTEC_ATTEST_BITSHIFT) &
+		SWITCHTEC_ATTEST_BITMASK;
+	state->attn_state.uds_selfgen =
+		(reply.cfg >> SWITCHTEC_UDS_SELFGEN_BITSHIFT) &
+		SWITCHTEC_UDS_SELFGEN_BITMASK;
```
lsgunth, line 467: Should probably be more specific here. Nothing links the values in `enum switchtec_attestation_mode` with the hardware values.  And they probably shouldn't be linked.

lsgunth (approved):
Mostly seems fine except some small issues.

lib/mfg.c
```diff
@@ -149,39 +149,60 @@ struct get_cfgs_reply {
...
+{
+	uint32_t subcmd = 0;
+	int ret;
+
+	ret = switchtec_mfg_cmd(dev,
+				MRPC_SECURITY_CONFIG_GET_GEN5,
+				&subcmd, sizeof(subcmd),
+				cfgs, sizeof(struct get_cfgs_reply_gen5));
+	if (!ret)
+		*otp_valid = true;
```
kelvin-cao, line 208: It seems opt_valid is always true when we are able to get the cfgs, then we probably don't need the opt_valid as an out parameter.

kelvin-cao (approved):
Looks good to me except for a minor concern.

PR #296, opened 2021-12-03
Use new MRPC command on Gen5 device

xyan264:
News MRPC command IDs are allocated to Gen5 devices. Also, for some commands, a _subcommand_ field is added to the MRPC command data.

The general approach is to define `xxx_gen4()` and `xxx_gen5()` functions, and call the right functions according to device generation.


lib/mfg.c
```diff
@@ -1041,20 +1044,26 @@ int switchtec_kmsk_set(struct switchtec_dev *dev,
 		       struct switchtec_kmsk *kmsk)
 {
 	int ret;
+	uint32_t cmd_id;
+
+	if (switchtec_is_gen5(dev))
+		cmd_id = MRPC_KMSK_ENTRY_SET_GEN5;
+	else
+		cmd_id = MRPC_KMSK_ENTRY_SET;
```
lsgunth, line 1139: A bit odd that the only difference between gen5 and gen4 is the command id. Normally you'd change the command ID because the interface changed some how. So I'd rather expect to see different send functions with different payloads.

lsgunth (approved):
Looks good to me.

kelvin-cao (approved):
Looks good to me. Thanks.

PR #297, opened 2021-12-13
Update 'mfg config-set' code for Gen5 RevB device

xyan264:
Gen5 RevB device introduced the following changes to 'mfg config-set' command:
- use new MRPC command (and input data format) to send OTP data
- use new OTP config file format which includes bit-fields for attestation settings
- allow user to specify an optional UDS file in the command line (`-u` option)


lib/mfg.c
```diff
@@ -992,70 +1025,154 @@ int switchtec_read_sec_cfg_file(struct switchtec_dev *dev,
...
+ * @param[out] set		Security settings
+ * @return 0 on success, error code on failure
+ */
+int switchtec_read_sec_cfg_file(struct switchtec_dev *dev,
+				FILE *setting_file,
+				struct switchtec_security_cfg_set *set)
+{
+	int ret;
+
+	ret = read_sec_cfg_header(dev, setting_file);
```
kelvin-cao, line 1166: Does check_sec_cfg_header() make more sense?

lib/mfg.c
```diff
@@ -992,70 +1025,154 @@ int switchtec_read_sec_cfg_file(struct switchtec_dev *dev,
...
+	int attest_mode;
+
+	ret = get_configs_gen5(dev, &reply);
+	if (ret)
+		return ret;
 
 	memset(set, 0, sizeof(struct switchtec_security_cfg_set));
 
-	file_data.data.cfg = le64toh(file_data.data.cfg);
+	ret = fread(&data, 1, sizeof(data), setting_file);
```
kelvin-cao, line 1099: Swap the second and third parameter?

kelvin-cao (approved):
Seems good to me. Thanks.

lsgunth (approved):
Could use some better commit messages, but otherwise it looks good once Kelvin's comments are addressed.

PR #298, opened 2021-12-21
Gen5 Rev-B new commands, part 2

xyan264:
This PR includes the following changes/features:
- New `image-list`, `image-select` implementation for Gen5 device
- New MRPC command ID for `fw-update`, `fw-transfer` command
- Add RIOT image to the list of switchtec firmware images

(RIOT can be understood as a firmware image, just like a boot-loader, that runs before main firmware. It is mainly used in secure boot applications.)


lsgunth (approved):
I don't see any obvious issues.

kelvin-cao (approved):
Seems good to me. Thanks!

PR #300, opened 2021-12-30
Add NULL file check in 'log-parse' command

xyan264:
The input log file being empty is a frequent issue for 'log-parse' command. 

We add this extra checking to ensure it's handled nicely.


kelvin-cao (approved):
Looks good to me. Thanks.

PR #301, opened 2022-01-04
Fix 'mfg info' command in BL2 boot phase

xyan264:
Currently 'mfg info' will return 'Unknown MRPC command xxx' in BL2 boot phase. 

This is due to function `switchtec_security_config_get()` is not available in BL2. But the other functions required for 'mfg info' are still supported in BL2.

We change the call sequence so that boot phase, chip serial and security versions can be printed in BL2.


lsgunth (approved):
Looks fine to me.

kelvin-cao (approved):
Looks good. Thanks.

PR #310, opened 2022-03-24
Misc fixes for 'diag' commands

xyan264:
This PR adds the following fixes:
- add missing '\n' in error messages
- make _stack_ parameter mandatory in `refclk` command 
- add delay to eye diagram start/cancel command: hardware needs some time to complete the start/cancel operation. The ideal solution would be to check hardware state and return after the state has changed to STARTED/STOPPED. Unfortunately, hardware does not provide a 'check state' command, so we can only add a long enough delay before returning.

### DIFF
--- a/cli/diag.c
+++ b/cli/diag.c
@@ -1979,7 +1979,9 @@ static int refclk(int argc, char **argv)
 		int stack_id;
 		int enable;
 		int disable;
-	} cfg = {};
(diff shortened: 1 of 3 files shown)


lsgunth (approved):
Looks good to me. thanks!

kelvin-cao (approved):
Looks good.

PR #315, opened 2022-06-02
'fw-toggle' command without retry

xyan264:
There is a need to make MRPC command _MRPC_FW_TX_ subcommand _MRPC_FW_TX_TOGGLE_ a 'no-retry' version, so that Switchtec-user won't keep sending toggle command while hardware is busy toggling the partitions. 

Previously _gasop_cmd_ only supports 'no retry' on one MRPC command, regardless of the subcommand being issued, this has been updated in this PR to support subcommand-level 'no retry'.

### DIFF
--- a/lib/platform/gasops.c
+++ b/lib/platform/gasops.c
@@ -47,24 +47,44 @@
 #define gas_reg_write64(dev, val, reg) __gas_write64(dev, val, \
 						     &dev->gas_map->reg)
 
-static const char gasop_noretry_cmds[] = {
-	[MRPC_SECURITY_CONFIG_SET] = 1,
-	[MRPC_KMSK_ENTRY_SET] = 1,
-	[MRPC_SECURE_STATE_SET] = 1,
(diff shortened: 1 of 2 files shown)


lsgunth (approved):
Looks ok to me.

lib/platform/gasops.c
```diff
@@ -47,24 +47,44 @@
...
-static const char gasop_noretry_cmds[] = {
-	[MRPC_SECURITY_CONFIG_SET] = 1,
-	[MRPC_KMSK_ENTRY_SET] = 1,
-	[MRPC_SECURE_STATE_SET] = 1,
-	[MRPC_BOOTUP_RESUME] = 1,
-	[MRPC_DBG_UNLOCK] = 1,
-	[MRPC_SECURITY_CONFIG_SET_GEN5] = 1,
+struct no_retry_struct {
+	int no_retry;
+	int subcmd_num;
```
kelvin-cao, line 52: A bit confusing, use num_subcmd?

PR #318, opened 2022-08-10
HPECRAY-6: Wait for EP Tunnel Status to go Enabled

NateThornton:
Signed-off-by: Nate Roiger <nate.roiger@hpe.com>

### DIFF
--- a/cli/fabric.c
+++ b/cli/fabric.c
@@ -1596,6 +1596,16 @@ static int ep_tunnel_cfg(int argc, char **argv)
 	switch (cfg.cmd) {
 	case MRPC_EP_TUNNEL_ENABLE:
 		ret = switchtec_ep_tunnel_enable(cfg.dev, cfg.pdfid);
+		if (ret)
+			break;
+
+		while(1) {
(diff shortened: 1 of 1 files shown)


cli/fabric.c
```diff
@@ -1596,6 +1596,16 @@ static int ep_tunnel_cfg(int argc, char **argv)
 	switch (cfg.cmd) {
 	case MRPC_EP_TUNNEL_ENABLE:
 		ret = switchtec_ep_tunnel_enable(cfg.dev, cfg.pdfid);
+		if (ret)
+			break;
+
+		while(1) {
```
kelvin-cao, line 1602: Given the tunnel enabling operation is not a blocking one from the firmware point of view, I'm thinking should we move the check and wait loop to the script which calls this enabling command? As we probably don't want to wait forever, and if we added a timeout value here we would have to know what would be a proper value for it as it seems it could be EP specific. So keeping it just as is and move the status check to the script seems make more sense to me.

NateThornton: @kelvin-cao  Can you go ahead and fix this? I don't have the expertise to unpack what you said. The ticket, HPECRAY-6, remains open.

PR #319, opened 2022-08-12
build: Don't strip binary by default

wak-google:
Our meta build system doesn't like that the binary is pre-stripped prior to consumption. We should avoid stripping and let the user strip the binary if needed.

If you have any opinions about making this a flag, that is fine with me.


lsgunth (approved):
Seems fine to me

PR #320, opened 2022-08-13
Fixes related to the max number of ports

### LINKED ISSUE #317: seeing crashes with "switchtec status" command
We're seeing crashes with "switchtec status /dev/switchtec0".

I've traced the issue part way to this code:

lib/switchtec.c:
static void generate_lane_str(struct switchtec_status *s)
{
	int i, l;

	for (i = 0; i < s->cfg_lnk_width; i++)
		s->lanes[i] = 'x';

If cfg_lnk_width is > 16, there's a buffer overrun while writing "x".

At the moment, however, I don't know why cfg_lnk_width is garbage.  It's possible that the problem is local to our hardware.  It bothers me that every one doesn't see this.  :-)

anitazha:
We were running into the same memory corruption issues as seen in #317 because we're in the process of upgrading from an older version of `switchtec`. These were not gen4 or gen5 devices and so various places that used `SWITCHTEC_MAX_PORTS` would go out of bounds when accessing arrays.

I attempted to fix this by correcting the logic in a helper function ~and removing some places where there was a risk of going out of bounds. I did not remove all uses of `SWITCHTEC_MAX_PORTS` since most appeared to be safe (i.e. used to initialize arrays but bounds were checked properly). I'm not super knowledgeable about this tool so maybe a maintainer can verify if the other places need to be fixed...~

Fixes #317

### DIFF
--- a/inc/switchtec/switchtec.h
+++ b/inc/switchtec/switchtec.h
@@ -441,6 +441,15 @@ static inline int switchtec_is_gen5(struct switchtec_dev *dev)
 }
 
 /**
+ * @brief Return the max number of ports of a Switchtec device.
+ */
+static inline int switchtec_max_supported_ports(struct switchtec_dev *dev)
+{
+	return switchtec_is_gen5(dev) ? SWITCHTEC_MAX_PORTS :
+	       switchtec_is_gen4(dev) ? 52 : 48;
+}
+
+/**
  * @brief Return whether a Switchtec device is PFX.
  */
 static inline int switchtec_is_pfx(struct switchtec_dev *dev)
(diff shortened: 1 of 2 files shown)


cli/fabric.c
```diff
@@ -484,7 +485,9 @@ static int topo_info(int argc, char **argv)
 
 	argconfig_parse(argc, argv, CMD_DESC_TOPO_INFO, opts, &cfg, sizeof(cfg));
 
-	for(i = 0; i < SWITCHTEC_MAX_PORTS; i++)
+	max_ports = switchtec_max_supported_ports(cfg.dev);
+
+	for(i = 0; i < max_ports; i++)
```
lsgunth, line 490: I don't really see the point of this patch. It's initializing a structure that's larger than the what the hardware might support and that's fine. It doesn't seem like it solves any actual bugs and just makes the code more complicated.

anitazha: For this commit, I was trying to remove the possibility of accessing beyond the max number of ports supported by the hardware. For `switchtec status` calls, reading beyond the max supported number of ports resulted in garbage and seg faults. If that's not an issue in this commit I can drop it.

lsgunth: Yeah, the first two commits look good, I'd drop the last one.

lsgunth (approved):
Looks good to me, thanks!

Issue #323, opened 2022-10-26
info command returning "Bad message"


arshad512:
I have been trying to debug an issue I am facing with "info" command with little luck. Can anyone help me chase down an issue I am facing. The details are given below and any futher info required please let me know.

Thanks

Issue.
I am getting "Bad message" on my device when issuing an "info" command
```
# ./switchtec info /dev/switchtec0
dev info: Bad message
```
Details follows

This is the version I am using.
```
# ./switchtec version
switchtec cli version 4.0
```

Device test output
```
# ./switchtec test /dev/switchtec0
test: success
```

Debugging session analysis :

My understanding is that read_resp() will have first 4 bytes as return value. Followed by variable length data.
I could see that it is returning 0x00000000, however it is still failing with -1. Where is it going wrong. 

```
(gdb) x /12x buf
0x7fffffffde70:	0x00000000	etc...
```

More detailed debugging session(important parts)

```
(gdb) s
read_resp (resp=resp@entry=0x7fffffffdf58, resp_len=resp_len@entry=8, ldev=<optimized out>) at lib/platform/linux.c:331
331	    size_t bufsize = sizeof(uint32_t) + resp_len;
(gdb) n
332		char buf[bufsize];
(gdb) 
334		ret = read(ldev->fd, buf, bufsize);
(gdb) p ldev->fd
value has been optimized out
(gdb) p bufsize
$19 = 12
(gdb) n
336		if (ret < 0)
(gdb) p ret
$20 = -1
(gdb) x /12x buf
0x7fffffffde70:	0x00000000	0x63592f39	0x00179ec8	0x00000000
0x7fffffffde80:	0xffffdf28	0x00007fff	0xf7fe76c0	0x00007fff
0x7fffffffde90:	0xffffdec0	0x00007fff	0x00000008	0x00000000
(gdb) n
353		return ret;
(gdb) n
switchtec_cmd (dev=dev@entry=0x648330, cmd=4294705408, cmd@entry=256, payload=payload@entry=0x7fffffffdf54, payload_len=payload_len@entry=4, 
    resp=resp@entry=0x7fffffffdf58, resp_len=resp_len@entry=8) at lib/platform/platform.c:174
174		if (ret > 0) {
(gdb) p ret
$26 = -1
(gdb) n
179		return ret;
(gdb) n
[...]

PR #324, opened 2022-10-27
lib: Fallback on EBADMSG error code

lsgunth:
Some GEN3 hardware ends up responding to MRPC_I2C_TWI_PING with a bad status code and the kernel translates into an EBADMSG error code. When this happens, the code should fall back into the GEN3 path.

Resolves: #323 ("info command returning 'Bad message'")


kelvin-cao (approved):
Looks good to me. Thanks Logan.

Issue #325, opened 2022-10-28
events command returning "Invalid argument"


arshad512:
Issue:
Running "events" command reports "Invalid argument"

Expected:
Should work successfully and not throw ("Invalid argument")

Please let me know if any further information/debuging info is required. 

Thanks

Output currently seen:
```
# ./switchtec events /dev/switchtec0
event_ctl: Invalid argument
```

**Details follows**

Switchtec version

```
# ./switchtec version
switchtec cli version 4.0
```
Hardware and firmware details

```
# cat /sys/class/switchtec/switchtec0/fw_version 
01090063
# cat /sys/class/switchtec/switchtec0/vendor_id 
MICROSEM
# cat /sys/class/switchtec/switchtec0/component_id
PM8532
```
Debugging session(important parts)
linux_event_ctl()  passes for SWITCHTEC_PART_EVT_PART_RESET and SWITCHTEC_PFF_EVT_TLP_THROTTLING. It fails for SWITCHTEC_PFF_EVT_AER_IN_P2P

```
Breakpoint 1, linux_event_ctl (dev=0x648390, e=SWITCHTEC_PART_EVT_PART_RESET, index=0, flags=0, data=0x0) at lib/platform/linux.c:906
906		if (e >= SWITCHTEC_MAX_EVENTS)


Breakpoint 1, linux_event_ctl (dev=0x648390, e=SWITCHTEC_PFF_EVT_TLP_THROTTLING, index=0, flags=0, data=0x0) at lib/platform/linux.c:906
906		if (e >= SWITCHTEC_MAX_EVENTS)
(gdb) c
Continuing.

Breakpoint 1, linux_event_ctl (dev=0x648390, e=SWITCHTEC_PFF_EVT_AER_IN_P2P, index=197, flags=0, data=0x0) at lib/platform/linux.c:906
906		if (e >= SWITCHTEC_MAX_EVENTS)
(gdb) c
Continuing.
event_ctl: Invalid argument
```

lsgunth:
After staring at the code for a bit, I think the problem is with the legacy event summary ioctl (SWITCHTEC_IOCTL_EVENT_SUMMARY_LEGACY) fallback path.

I'm assuming you are using an older kernel module that is falling back to this. If you are using a kernel older than about v5.2 then you will be exercising the fallback code. Please confirm.

[...]

PR #326, opened 2022-10-31
Ensure entire switchtec_event_summary struct is set

lsgunth:
In normal cases, event_summary_copy() will set the entire switchtec_event_summary struct. However, if the library falls back to the smaller SWITCHTEC_IOCTL_EVENT_SUMMARY_LEGACY, then a size less than SWITCHTEC_MAX_PFF_CSR will be passed to the function and a portion of the function will be left unset.

There are a number of places in the code where this function gets called pointing to an unitialized structure on the stack. So fix this issue by ensuring event_summary_copy() sets the entire structure.

Resolves: #325 ("events command returning 'Invalid argument'")


kelvin-cao (approved):
Looks good to me, thanks!

PR #328, opened 2022-11-14
Info

kelvin-cao:
Generally two fixes in this PR.

1. In BL2 boot phase, use MRPC instead of GAS access to retrieve device id, to make command info be capable to report them correctly.
2. Make command info's output more reasonable in BL1 boot phase by showing the boot phase in all boot phases and 'N/A' for unavailable information in BL1. 

The updated output looks like below.
```
# sudo ./switchtec info /dev/i2c-5:0x58
/dev/i2c-5:0x58 (BL1):
    Generation:  GEN4
    HW Revision: REVB
    Variant:     N/A
    Device ID:   N/A
    FW Version:  N/A

# sudo ./switchtec mfg boot-resume /dev/i2c-5:0x58
WARNING: if your system does not support hotplug,
your device might not be immediately accessible
after a normal boot process. In this case, be sure
to reboot your system after sending this command.

Do you want to continue? [y/N] y

# sudo ./switchtec info /dev/i2c-5:0x58
/dev/i2c-5:0x58 (BL2):
    Generation:  GEN4
    HW Revision: REVB
    Variant:     PSX
    Device ID:   0x4100
    FW Version:  3.90 B059
```

### DIFF
--- a/inc/switchtec/switchtec.h
+++ b/inc/switchtec/switchtec.h
@@ -610,6 +610,24 @@ static inline const char *switchtec_variant_str(struct switchtec_dev *dev)
 	return str;
 }
 
+/**
+ * @brief Return the phase string for a phase id.
+ */
+static inline const char* switchtec_phase_id_str(
+		enum switchtec_boot_phase phase_id)
+{
+	switch(phase_id) {
+	case SWITCHTEC_BOOT_PHASE_BL1:
+		return "BL1";
+	case SWITCHTEC_BOOT_PHASE_BL2:
+		return "BL2";
+	case SWITCHTEC_BOOT_PHASE_FW:
+		return "Main Firmware";
+	default:
+		return "Unknown Phase";
+	}
+}
+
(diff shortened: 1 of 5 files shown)


inc/switchtec/switchtec.h
```diff
@@ -610,6 +610,23 @@ static inline const char *switchtec_variant_str(struct switchtec_dev *dev)
 	return str;
 }
 
+/**
+ * @brief Return the phase string for a phase id.
+ */
+static inline const char* phase_id_to_string(enum switchtec_boot_phase phase_id)
```
lsgunth, line 616: Might be a good idea to rename this with a switchtec_prefix, or some such.

lsgunth (approved):
One minor nit I noticed. Otherwise it looks good to me.

inc/switchtec/switchtec.h
kelvin-cao: Updated, thanks for the review!

PR #329, opened 2023-01-11
cli/progress: Fix int overflow with progress bar

amaier17:
If an image size of 40MB is passed in an integer overflow occurs (assuming a terminal width >100) because the current is multipled by the bar width (40MB * 100 > 32-bit signed int). Casting this to a long long fixes the issue.

### DIFF
--- a/cli/progress.c
+++ b/cli/progress.c
@@ -77,7 +77,7 @@ static void timeval_subtract(struct timeval *result,
 
 static void print_bar(int cur, int total)
 {
-	int bar_width;
+	long long bar_width;
 	int i;
 	float progress = cur * 100.0 / total;
 	int pos;


cli/progress.c
```diff
@@ -83,7 +83,7 @@ static void print_bar(int cur, int total)
 	int pos;
 
 	bar_width = get_columns() - 33;
-	pos = bar_width * cur / total;
+	pos = (long long)bar_width * cur / total;
```
lsgunth, line 86: Why not make bar width a (long long) instead of the ugly cast?

Not sure swithctec images are ever greater than 40MB, but I guess it would be good to fix it here.

lsgunth (approved):
Looks good, thanks!

PR #335, opened 2024-04-18
Crosshair fix

Chetana2791:
Problem: Switchtec user utility for cross hair scan outputs incorrect results for right scan.
Fix: The output messaging for top right and bottom right scan has been fixed.


lsgunth (approved):
Thanks! Looks like a good fix.

Chetana2791:
recreating the PR with the latest updates from devel branch.

PR #341, opened 2025-03-06
Log parse support for FTDC binary logs

BenReed161:
Added new type option to the log parser command to handle FTDC binary logs which do not contain the "magic string" at the beginning of app logs. -t FTDC


lsgunth (approved):
Looks good to me.

kelvin-cao (approved):
The code looks good to me. Thanks.

BenReed161 (commented):
reviewed

PR #342, opened 2025-03-20
Support diag ltssm cmd for gen5

BenReed161:


lib/diag.c
```diff
@@ -907,14 +907,151 @@ int switchtec_diag_refclk_ctl(struct switchtec_dev *dev, int stack_id, bool en)
...
+		uint8_t port;
+		uint8_t freeze;
+		uint8_t unused;
+	} ltssm_freeze;
+
+	struct {
+		uint8_t sub_cmd;
+		uint8_t port;
+	} status;
+	struct {
```
kelvin-cao, line 932: Need a line in between the two structure definitions.

lib/diag.c
```diff
@@ -907,14 +907,151 @@ int switchtec_diag_refclk_ctl(struct switchtec_dev *dev, int stack_id, bool en)
...
+	} status_output;
+
+	struct {
+		uint8_t sub_cmd;
+		uint8_t port;
+		uint16_t log_index;
+		uint16_t no_of_logs;
+	} log_dump;
+
+
```
kelvin-cao: One blank line here.

lib/diag.c
```diff
@@ -907,14 +907,151 @@ int switchtec_diag_refclk_ctl(struct switchtec_dev *dev, int stack_id, bool en)
...
+	struct {
+		uint16_t length;
+		uint16_t dw1_unused;
+		uint16_t dw1;
+		uint32_t ram_timestamp;
+		uint32_t unused;
+	} log_dump_out[512];
```
kelvin-cao, line 951: Is this structure definition aligned with dev spec?

lib/diag.c
```diff
@@ -907,14 +907,151 @@ int switchtec_diag_refclk_ctl(struct switchtec_dev *dev, int stack_id, bool en)
...
+	uint32_t timestamp;
+
+	int major;
+	int minor;
+	int rate;
+	int ret;
+	int i;
+
+	/* freeze logs */
+	ltssm_freeze.sub_cmd = 14;
```
kelvin-cao, line 963: Please define sub command in inc/switchtec/mrpc.h.

lib/diag.c
```diff
@@ -907,14 +907,151 @@ int switchtec_diag_refclk_ctl(struct switchtec_dev *dev, int stack_id, bool en)
...
  * @param[in]	dev    Switchtec device handle
  * @param[in]	port   Switchtec Port
  * @param[inout] log_count number of log entries
  * @param[out] log    A pointer to an array containing the log
  *
  */
-int switchtec_diag_ltssm_log(struct switchtec_dev *dev,
+int switchtec_diag_ltssm_log_gen5(struct switchtec_dev *dev,
+	int port, int *log_count,
+	struct switchtec_diag_ltssm_log *log_data)
```
kelvin-cao, line 919: We probably can keep switchtec_diag_ltssm_log() and differentiate gen4 and gen5 within it and call xx_log_gen4 and xx_log_gen5 accordingly so that we won't break the API for current gen4 users.

lsgunth (approved):
Did a quick review, didn't find any issues. Thanks for many of the cleanups!

cli/diag.c
```diff
@@ -126,16 +126,20 @@ static int ltssm_log(int argc, char **argv) {
...
-	int log_count = 128;
 	int i;
 
 	ret = diag_parse_common_cfg(argc, argv, CMD_DESC_LTSSM_LOG,
 				    &cfg, opts);
 	if (ret)
 		return ret;
+	
+	int log_count = 512;
+	if (!switchtec_is_gen5(cfg.dev)) {
```
kelvin-cao, line 139: Checking for gen4 probably makes more sense here.

inc/switchtec/switchtec.h
```diff
@@ -1239,8 +1239,8 @@ int switchtec_diag_perm_table(struct switchtec_dev *dev,
 			      struct switchtec_mrpc table[MRPC_MAX_ID]);
 int switchtec_diag_refclk_ctl(struct switchtec_dev *dev, int stack_id, bool en);
 int switchtec_diag_ltssm_log(struct switchtec_dev *dev,
-			     int port, int *log_count,
-			     struct switchtec_diag_ltssm_log *log_data);
+				int port, int *log_count,
```
kelvin-cao, line 1242: Seems no alignment issue with original code.

lib/diag.c
```diff
@@ -907,16 +909,180 @@ int switchtec_diag_refclk_ctl(struct switchtec_dev *dev, int stack_id, bool en)
...
+				    log_dump.no_of_logs * sizeof(struct log_dump_out) + 4);
+		if (ret)
+			return ret;
+		for (int i = 0; i < log_dump.no_of_logs; i++) {
+			dw0 = log_dump_out_ptr[i].dw0;
+			timestamp = log_dump_out_ptr[i].ram_timestamp;
+	
+			rate = (dw0 >> 13) & SWITCHTEC_LTSSM_LOG_MASK;
+			major = (dw0 >> 7) & SWITCHTEC_LTSSM_LOG_MASK;
+			minor = (dw0 >> 3) & SWITCHTEC_LTSSM_LOG_MASK;
```
kelvin-cao, line 1006: According to what I read from the spec, the above three fields are of different lengths and therefore should use different masks, why all use 0xF here?
bit [15:13] LTSSM_RAM_LINK_RATE
bit [12:7] LTSSM_RAM_MAIN_STATE
bit [6:3] LTSSM_RAM_SUBSTATE

cli/diag.c
```diff
@@ -126,16 +126,20 @@ static int ltssm_log(int argc, char **argv) {
...
+		log_count = 128;
+	}
```
kelvin-cao, line 141: The surrounding {} should be omitted for this single-statement block inside the if condition, as well as other similar instances.

lib/diag.c
BenReed161: Updated masks to reflect the different lengths of each field.

lib/diag.c
```diff
@@ -907,16 +911,180 @@ int switchtec_diag_refclk_ctl(struct switchtec_dev *dev, int stack_id, bool en)
...
+				timestamp = log_dump_out_ptr[i].ram_timestamp;
+		
+				rate = (dw0 >> 13) & SWITCHTEC_LTSSM_LOG_MASK_RATE;
+				major = (dw0 >> 7) & SWITCHTEC_LTSSM_LOG_MASK_MAJOR;
+				minor = (dw0 >> 3) & SWITCHTEC_LTSSM_LOG_MASK_MINOR;
+		
+				log_data[curr_idx + i].timestamp = timestamp;
+				log_data[curr_idx + i].link_rate = switchtec_gen_transfers[rate+1];
+				log_data[curr_idx + i].link_state = major | (minor << 8);
+			}
```
kelvin-cao, line 1062: The logic of issuing MRPC and assigning retrieved log to log_data repeated several times in the code, possible to extract it to make it a helper function?

lib/diag.c
```diff
@@ -907,16 +911,180 @@ int switchtec_diag_refclk_ctl(struct switchtec_dev *dev, int stack_id, bool en)
...
+			if (ret)
+				return ret;
+			log_dump_out_ptr = (struct log_dump_out *)&(log_buffer[4]);
+			for (int i = 0; i < log_dump.no_of_logs; i++) {
+				dw0 = log_dump_out_ptr[i].dw0;
+				timestamp = log_dump_out_ptr[i].ram_timestamp;
+		
+				rate = (dw0 >> 13) & SWITCHTEC_LTSSM_LOG_MASK_RATE;
+				major = (dw0 >> 7) & SWITCHTEC_LTSSM_LOG_MASK_MAJOR;
+				minor = (dw0 >> 3) & SWITCHTEC_LTSSM_LOG_MASK_MINOR;
```
kelvin-cao, line 1057: If a helper function is possible, we probably can remove the mask macros and just expand the value in the statement, just as we put shift offset values here.

lib/diag.c
```diff
@@ -1027,4 +1195,26 @@ int switchtec_diag_ltssm_log(struct switchtec_dev *dev,
...
+ *
+ */
+int switchtec_diag_ltssm_log(struct switchtec_dev *dev,
+			    int port, int *log_count,
+			    struct switchtec_diag_ltssm_log *log_data)
+{
+	int ret;
+	if (switchtec_is_gen5(dev)) {
+		ret = switchtec_diag_ltssm_log_gen5(dev, port, log_count, log_data);
+	}
```
kelvin-cao, line 1213: Surrounding {} not necessary.

PR #343, opened 2025-04-02
Added minor fixes to memory management

BenReed161:
- First change in lib/fw.c relates to the summary pointer not being freed after returning from the default case in the switch statement Added a call to free the summary pointer before returning.

- Second change in cli/main.c changed the 64 byte buffer sent to switchtec_get_fw_version function to 32. Since we are expecting the version information to be at most 32 bytes there is no need to have the destination buffer set to 64 when strncpy is called in the switchtec_get_fw_version function within lib/platform/platform.c.

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -188,7 +188,7 @@ static int print_dev_info(struct switchtec_dev *dev)
 {
 	int ret;
 	int device_id;
-	char version[64];
+	char version[32];
 	enum switchtec_boot_phase phase;
 	enum switchtec_rev hw_rev;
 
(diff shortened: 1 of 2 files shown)


lsgunth (approved):
Good catch on the missing free(). That's definitely worth while.

I'm not sure the point of reducing the version size at all. Saving a tiny amount of stack memory at the cost of a potential future overflow doesn't really seem worthwhile.  At the very least these should have been in two separate patches.

kelvin-cao (approved):

PR #345, opened 2025-04-15
Add tlp inject command

BenReed161:


cli/diag.c
```diff
@@ -2104,6 +2104,7 @@ static int tlp_inject (int argc, char **argv)
 	}
 	if (num_dwords > 132) {
 		fprintf(stderr, "TLP data cannot exceed 132 dwords \n");
+		free(raw_tlp_dwords);
```
lsgunth, line 2109: This change is not mentioned in the commit message. The free() call looks a bit suspicious but its removal should probably have some comments with it.

lib/diag.c
```diff
@@ -1214,4 +1214,26 @@ int switchtec_diag_ltssm_log(struct switchtec_dev *dev,
...
+int switchtec_tlp_inject(struct switchtec_dev * dev, int port_id, int tlp_type, 
+			 int tlp_length, int ecrc, uint32_t * raw_tlp_data)
+{
+	uint32_t tlp_out;
+	int ret = 1;
+	struct switchtec_tlp_inject_in tlp_in = {
+		.dest_port = port_id,
+		.tlp_type = tlp_type,
+		.tlp_length = tlp_length,
+		.ecrc = ecrc
```
kelvin-cao, line 1226: For multi-byte field, need to handle endianness properly on input/output data.

cli/diag.c
```diff
@@ -2026,6 +2026,91 @@ static int refclk(int argc, char **argv)
...
+	const struct argconfig_options opts[] = {
+		DEVICE_OPTION,
+		{"dest_port", 'p', "PORT_ID", CFG_NONNEGATIVE, &cfg.port_id, 
+			required_argument, "destination port ID"},
+		{"tlp_type", 't', "TYPE", CFG_NONNEGATIVE, &cfg.tlp_type, 
+			required_argument, "tlp type 0: P, 1: NP, 2: CP (default 0)"},
+		{"enable_ecrc", 'e', "", CFG_NONE, &cfg.ecrc, no_argument, 
+			"Enable the ecrc to be included at the end of the input data (Default: disabled)"},
+		{"tlp_data", 'd', "DW0 DW1 ... DW131", CFG_STRING, &cfg.raw_tlp_data, required_argument, 
+			"DW to be sent as part of the raw TLP (Maximum 132 DWs)"},
```
kelvin-cao: If we expect every DW must start with '0x', probably need to let user know.

cli/diag.c
```diff
@@ -2026,6 +2026,100 @@ static int refclk(int argc, char **argv)
...
+		fprintf(stderr, "Must set tlp data --tlp_data -d \n");
+		return -1;
+	}
+	ret = convert_str_to_dwords(cfg.raw_tlp_data, &raw_tlp_dwords, 
+				    &num_dwords);
+	if (ret) {
+		fprintf(stderr, "Error with tlp data provided \n");
+		return -1;
+	}
+	if (num_dwords > 132) {
```
kelvin-cao, line 2107: 132 is used multiple times, please use a macro instead.

kelvin-cao:
As a rule of thumb, if you've made corrections to a commit, consider squashing it and performing a hard push. It's better to avoid introducing additional commits to rectify an earlier one within the same pull request.

lib/diag.c
```diff
@@ -1214,4 +1214,26 @@ int switchtec_diag_ltssm_log(struct switchtec_dev *dev,
...
+	};
+	for (int i = 0; i < tlp_in.tlp_length; i++)
+	{
```
kelvin-cao, line 1229: Style nit: '{' should be on the same line of 'for'

inc/switchtec/diag.h
```diff
@@ -279,5 +279,13 @@ struct switchtec_diag_ltssm_log_dmp_out {
 	uint32_t arc;
 };
 
+struct switchtec_tlp_inject_in {
+	uint32_t dest_port;
+	uint32_t tlp_type;
+	uint32_t tlp_length;
+	uint32_t ecrc;
+	uint32_t raw_tlp_data[132];
```
kelvin-cao, line 287: Probably need the macro TLP_MAX_DWS  in the header and replace this '132' instance as well.

cli/diag.c
```diff
@@ -2026,6 +2026,91 @@ static int refclk(int argc, char **argv)
...
+	}
+
+	*dwords = (uint32_t *)malloc(*num_dwords * sizeof(uint32_t));
+	if (*dwords == NULL)
+		return 1;
+
+	ptr = str;
+	for (int i = 0; i < *num_dwords; i++) {
+		char *endptr;
+		(*dwords)[i] = (uint32_t)strtoul(ptr, &endptr, 0);
```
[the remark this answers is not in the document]
kelvin-cao: Do we handle this somewhere else?

PR #349, opened 2025-06-11
add minor version to info command

BenReed161:
Added the minor version number to the info command through reading the GAS address for Firmware Version. Add additional version info to the HW Revision section.

### DIFF
--- a/lib/platform/gasops.c
+++ b/lib/platform/gasops.c
@@ -188,6 +188,15 @@ int gasop_get_fw_version(struct switchtec_dev *dev, char *buf,
 	return 0;
 }
 
+int gasop_get_device_version(struct switchtec_dev *dev, int *res)
+{
+	uint32_t dev_ver;
+	dev_ver = gas_reg_read32(dev, sys_info.device_version);
+
+	*res = dev_ver;
+	return 0;
+}
+
 int gasop_pff_to_port(struct switchtec_dev *dev, int pff,
 		      int *partition, int *port)
 {
--- a/lib/platform/linux.c
+++ b/lib/platform/linux.c
@@ -300,6 +300,26 @@ static int linux_get_fw_version(struct switchtec_dev *dev, char *buf,
 	return 0;
 }
 
+static int linux_get_device_version(struct switchtec_dev *dev, int *version_res)
+{
+	int ret;
+	int version;
+	char syspath[PATH_MAX];
+	struct switchtec_linux *ldev = to_switchtec_linux(dev);
+
+	ret = dev_to_sysfs_path(ldev, "device_version", syspath, sizeof(syspath));
+	if (ret)
+		return ret;
+
+	version = sysfs_read_int(syspath, 16);
+	if (version < 0)
+		return version;
+	
+	memcpy(version_res, &version, sizeof(int));
+
+	return 0;
+}
+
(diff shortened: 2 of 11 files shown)


lib/platform/linux.c
```diff
@@ -971,6 +971,7 @@ static const struct switchtec_ops linux_ops = {
 	.close = linux_close,
 	.get_device_id = linux_get_device_id,
 	.get_fw_version = linux_get_fw_version,
+	.get_device_minor_ver = gasop_get_device_minor_ver,
```
lsgunth, line 974: Seems a bit suspicious to me that the linux version is using the gasop here... Why is there even an op when all implementations use the same function?

lib/platform/gasops.c
```diff
@@ -188,6 +188,32 @@ int gasop_get_fw_version(struct switchtec_dev *dev, char *buf,
...
+{
+	void __gas * addr;
+	int offset = 0x2004;
+	gasptr_t map;
+	size_t map_size;
+	uint32_t gas_result;
+	char result[5];
+
+	uint8_t minor;
+	map = switchtec_gas_map(dev, 0, &map_size);
```
lsgunth, line 201: Also highly questionable that this function is calling switchtec_gas_map()

lib/platform/gasops.c
```diff
@@ -188,6 +188,32 @@ int gasop_get_fw_version(struct switchtec_dev *dev, char *buf,
 	return 0;
 }
 
+int gasop_get_device_minor_ver(struct switchtec_dev *dev, char ** res)
```
lsgunth, line 191: Small nit: should be 'char **res'.

BenReed161: fixed

lib/platform/gasops.c
BenReed161: Removed gas map functions from the get device minor function. Moved function to platform file.

lib/platform/linux.c
BenReed161: Removed the gasop function that was added in the linux platform files. Instead the functionality for getting the minor version is shifted to the platform file.

lib/platform/platform.c
```diff
@@ -150,6 +150,20 @@ int switchtec_get_fw_version(struct switchtec_dev *dev, char *buf,
...
 }
 
+int switchtec_get_device_minor_ver(struct switchtec_dev *dev, char **res)
+{
+	uint32_t addr = 0x2004;
+	uint32_t gas_result;
+	char result[5];
+	uint8_t minor;
+	
+	gas_result = __gas_read32(dev, &addr);
```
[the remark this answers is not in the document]
BenReed161: Added dev->ops->gasmap/unmap to surround gas_read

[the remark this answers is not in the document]
BenReed161: Would this addition make these changes legal for the gas?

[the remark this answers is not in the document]
BenReed161: Instead of calling gas_read directly mapping/unmapping and accessing the gas this way, could using the MRPC interface and calling the MPRC_GAS_READ command be better in this situation? Or would this be just as expensive as the previous?

lsgunth (commented):
This looks good to me, thanks.

The one thing I'm unclear on is whether the op implemented is only for Linux? Am I missing the gasop implementation?

kelvin-cao (approved):

PR #352, opened 2025-07-02
Add ordered set analyzer commands

BenReed161:
Add ordered set commands
Add requested OSA ordered set analyzer coammnds to the switchtec user
commands to handle OSA split into 7 commands which reflect each MRPC
sub command as part of the OSA group.

osa - command which triggers specific osa operations on a given stack

osa config type - configure the osa type 4 different masks. Lane mask
to trigger on specific lanes, a directional mask, a link rate mask, and
ordered set type mask, each on a given stack

osa config pattern - configure osa pattern with similar masks as
previous minus a directional mask. Value and mask dwords are added to
configure the value of the pattern and the mask for the pattern.
The corresponding bit in the pattern value is used for matching.

osa config misc - This is for triggering the configuration on trigger
types not covered by the previous two commands. There a trigger enable
mask for 3 corresponding types: LTMON, other HW blocks, general
purpose.

osa capture control - This command allows the user to control how the
OSA is captured along lane and direction using masks as well how many
entries to capture, what types, and specific modes to capture on.

osa dump config - This command will dump the config on the specified
stack.

osa dump data - This command will dump the data for the osa given a
lane mask and directional mask, for a specific stack.


lsgunth (approved):
I only did a very cursory look. Lots of code here. Seems reasonable from a high level.

One issue is the patches appear to be out of order seeing the first patch uses  convert_hex_str() which is introduced by the second patch.

There used to be continuous integration on this project to prevent things like this. Seems to be broken now.

lib/diag.c
```diff
@@ -1933,4 +1933,434 @@ int switchtec_inject_err_cto(struct switchtec_dev *dev, int phys_port_id)
...
+	int curr_idx = 0;
+	uint32_t timestamp_upper = 0;
+	uint32_t timestamp_lower = 0;
+	uint64_t timestamp = 0;
+	char data_string[45];
+	uint32_t osa_dword_data[4];
+
+	printf("IDX\tTIMESTAMP\tCNT\tRATE\tDRP\tTRIG\tDATA\n");
+	for (int i = 0; i < entries_read; i++)
+	{
```
kelvin-cao, line 1958: style nit: `{` should be on the same line of `for`, there's multiple instances of such issue for `for` blocks and for `if else` blocks

BenReed161: fixed

cli/diag.c
```diff
@@ -2472,6 +2472,553 @@ static int linkerr_inject(int argc, char ** argv)
...
+	uint32_t * mask_dwords_arr = NULL;
+	uint32_t * lane_mask = NULL;
+	uint32_t * direction_mask = NULL;
+	uint32_t * link_rate_mask = NULL;
+	int num_dwords = 0;
+	int total_dwords = 0;
+	static struct {
+		struct switchtec_dev *dev;
+		int stack_id;
+		char * direction;
```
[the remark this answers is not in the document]
BenReed161: fixed

PR #355, opened 2025-08-12
Switchtec fixes

BenReed161:
Minor fixes and small feature requests.
- Added fix for observed seg fault in code.
- Gen5 case for the txtable skips columns FOM, Post-Up Pre-Up since they are Gen4 columns
- Add clear argument for the ltssm log command
- Add a status print for all stacks for the refclk command


lib/platform/linux.c
```diff
@@ -362,8 +362,11 @@ static int read_resp(struct switchtec_linux *ldev, void *resp,
 	}
 
 	memcpy(&ret, buf, sizeof(ret));
-	if (ret)
-		errno = ret;
+
+	if (ret) {
+		errno = ENODATA;
+		return -errno;
```
lsgunth, line 368: I thought it was a little odd to be returning -errno here. I see it's already in this code above as well. Probably not that big a deal.

lsgunth (approved):
I saw one very minor point that could maybe be cleaned up later. Otherwise it looks good.

inc/switchtec/switchtec.h
```diff
@@ -1545,6 +1545,7 @@ int switchtec_diag_port_eq_tx_fslf(struct switchtec_dev *dev, int port_id,
 int switchtec_diag_perm_table(struct switchtec_dev *dev,
 			      struct switchtec_mrpc table[MRPC_MAX_ID]);
 int switchtec_diag_refclk_ctl(struct switchtec_dev *dev, int stack_id, bool en);
+int switchtec_diag_refclk_status(struct switchtec_dev *dev, uint8_t * stack_info);
```
kelvin-cao, line 1548: -> `*stack_info`

PR #357, opened 2025-08-26
Add redundant image partition code

BenReed161:
Add the image partition redundancy MRPC option as part of new command fw_redundant. Option -R which sets the given partition to be set as redundant. 0/1


kelvin-cao:
It looks weird to me to add a redundancy option to 'toggle' which supposes do one thing only. I'd suggest add a new command entry for this.

kelvin-cao:
@Chetana2791 do we see requirement of this feature for the utility?

Chetana2791:
> @Chetana2791 do we see requirement of this feature for the utility?

Yes, this is an urgent requirement from a customer

PR #368, opened 2026-03-09
support twi command

BenReed161:
Support the twi mrpc reset, write, and read sub commands as part of a new CLI command. Support 7/10 bit slave address.


lsgunth (approved):
Looks good to me

PR #369, opened 2026-03-19
cleanup OSA commands

BenReed161:
Original OSA code printf in the lib rather than in the cli code. Changed so OSA printing is handled in the CLI portion, cut down on the function size.


lsgunth (approved):
Makes sense to me

PR #372, opened 2026-04-24
Gen6 updates

BenReed161:


lib/platform/linux.c
```diff
@@ -364,8 +364,8 @@ static int read_resp(struct switchtec_linux *ldev, void *resp,
 	memcpy(&ret, buf, sizeof(ret));
 
 	if (ret) {
-		errno = ENODATA;
-		return -errno;
+		errno = ret;
+		return ret;
```
lsgunth: I'm not quite following this. Can we add more information to the commit message?

Is ret likely to be an errno value? If it isn't, we really shouldn't set errno to it. Returning the value may make sense.

lsgunth (approved):
Largely looks good. One small question about a bit I didn't fully follow.

lib/platform/linux.c
BenReed161: Instead of reporting from the firmware MRPC error it would report the "no data available" message when the port given was invalid. I think since the MRPC error it is giving isn't helpful to the user either, probably this commit can be removed in that case.

lsgunth: Ok, I dug a bit deeper into the code. Part of this is I don't remember how it all worked.

But if the cmd returns a positive number it is translated into an MRPC error in switchtec_cmd(). So I'm thinking the patch is correct but it might be good to just get some of this explanation in the commit message so it's a bit more understandable.

BenReed161: Updated the patch so the commit message is clearer

lsgunth (approved):
LGTM, thanks!

PR #373, opened 2026-04-27
Add Multicast configuration extension

BenReed161:
Add a new extension to switchtec user, called 'cap' which can be used to configure the PCIe capabilities available on the switch for each USP and DSP.
Added a new command set in the extension to configure the Multicast capability. Multicast-show and Multicast-set, provide full configuration of the PCIe space through GAS read and write.


lsgunth (approved):
LGTM

PR #375, opened 2026-04-30
Gen6 secure changes

BenReed161:
Changes copied over from internal branch, these also need to be included in switchtec-user main repo.


lsgunth (approved):
I skimmed through this. Nothing jumps out as an issue.

PR #377, opened 2026-05-14
Secure changes 2

BenReed161:
Second pass of security changes taken from internal branch, commits where adapted using Claude due to the size and the differences in internal branch versus main branch git


lsgunth (approved):
A couple of those patches are a lot bigger than I'm able to understand in my had. They seem fine on the surface but I can't comment on deeper issues.

PR #378, opened 2026-05-18
Osa enhancements

BenReed161:


cli/diag.c
```diff
@@ -3007,7 +3007,7 @@ static int osa_config_type(int argc, char **argv)
 	}
 	if (cfg.os_types) {
 		ret = convert_hex_str(cfg.os_types, &os_type_mask,
-				      &num_dwords, 1);
+				      &num_dwords, 2);
```
lsgunth: I'm finding this code extremely difficult to understand.  convert_hex_str() is weird. The documentation for os_types seems to indicate it just inputs a hex string like "0xA" with only 4 valid bits which should be really simple and shouldn't need a bunch of memory allocations and arrays of integers.

CFG_LONG already supports strings that have a 0x prefix so they can be specified in hexadecimal without all this extra work.

The documentation above for the command line parameter says there are only 4 bits in os_types. But this seems to be increasing it to take two hex characters (I assume). Then num_dwords is confusing because a d-word usually refers to something that is 4-bytes long or 8 hex characters.

lsgunth (approved):
That makes more sense, thanks!

cli/diag.c
BenReed161: Hi Logan, originally I used the convert_hex_str on a string of multiple dword hexadecimal values hence the allocating of arrays and such. But later down the line I also applied it to be used with regular hex values that weren't even dwords, without much thought as to easier solutions (as you mentioned). So I refactor the OSA functions which use convert_hex_str instead to use the default CFG_ args that have built-in 0x prefix handling. Some functions still use convert_hex_str but those are the original functions which needed it for multiple dwords in a string.
Hope that clears it up a little.
Thanks

PR #379, opened 2026-05-19
Info compat fix

BenReed161:
The B012 firmware (0a050012) does not support subcmd=1 for MRPC 0x124 returns error 0x7d020. The newer B013 firmware (0a060013) does support it. Since both subcmds return the same reply struct, this is a firmware feature-level incompatibility. Add fallback if first subcmd=1 case fails


lsgunth (approved):
The change looks good.

Weird to have a version number change in what appears to be a bug fix PR.

BenReed161:
Removed version commit in this PR, will add as a separate commit (not part of this PR).

PR #380, opened 2026-05-20
RTC counter was truncating 64 bit vals

BenReed161:
Add inttype to correct for platform type to 64 bits for the RTC counter.


lsgunth (approved):
LGTM

PR #382, opened 2026-05-21
Fix toggle in bl2 stage

BenReed161:
re-order the toggle, configuration and firmware first since we are
toggling whilst running in BL2. Add to the BL2 toggle and main-fw toggle
a no-poll option to sending the command (we don't want switchtec to poll in bl2 since
it could render i2c bus unresponsive in a bl2 toggle). Firmware takes longer
to toggle due to image CRC verfication, so we send without polling and give some
time for it to finish CRC checks. (usleep)

Add switchtec_cmd_nopoll as a alternative to polling ver

Add a no-retry function for Gen6 to match previous

### DIFF
--- a/lib/fw.c
+++ b/lib/fw.c
@@ -454,6 +454,16 @@ int switchtec_fw_img_get(struct switchtec_dev *dev, int fd,
 	return 0;
 }
 
+static int switchtec_cmd_nopoll(struct switchtec_dev *dev, uint32_t cmd_id,
+				const void *payload, size_t payload_len)
+{
+	struct mrpc_regs __gas *mrpc = &dev->gas_map->mrpc;
+
+	__memcpy_to_gas(dev, &mrpc->input_data, payload, payload_len);
+	__gas_write32_no_retry(dev, cmd_id, &mrpc->cmd);
+	return 0;
+}
+
 /**
  * @brief Toggle the active firmware partition for the main or configuration
  *	images.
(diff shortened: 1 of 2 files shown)


lib/fw.c
```diff
@@ -485,21 +495,32 @@ int switchtec_fw_toggle_active_partition(struct switchtec_dev *dev,
...
+		}
+
+		if (toggle_fw) {
+			cmd.toggle_fw = 1;
+			ret = switchtec_cmd_nopoll(dev, cmd_id, &cmd,
+						   cmd_size);
+			if (ret)
+				return ret;
+			cmd.toggle_fw = 0;
+			usleep(5000000);
```
lsgunth, line 522: This sleep is a bit suspect. Should it be there? If so, I think it needs a big comment above it.

BenReed161: Added comment explaining the 5 second sleep timer. Mainly this is to prevent hanging whilst the firmware validates the CRC of the main firmware partition.

lsgunth: Ok, that's not spectacular. It would be good if this could be addressed in future work. Getting some kind of signal from hardware for when it is done would be a lot better than a blind long wait. But if we are stuck with what we have then I guess we are stuck.

lsgunth (approved):

lib/fw.c
BenReed161: Yeah, not sure on a signal from hardware definitely something that needs to addressed in a future release. I agree with your sentiment, it's not ideal. For now getting the toggle working is a high priority for Gen6. 5 seconds was shortest I could get it without seeing the hanging issue.

PR #383, opened 2026-05-22
remove -m support for gen6 execute

BenReed161:
Remove the support for Gen6 -m option since it is no longer supported. Align the return codes of fw_execute function to be standard -1 return rather than -2

### DIFF
--- a/cli/mfg.c
+++ b/cli/mfg.c
@@ -1059,7 +1059,7 @@ static int fw_execute(int argc, char **argv)
 		int assume_yes;
 		enum switchtec_bl2_recovery_mode bl2_rec_mode;
 	} cfg = {
-		.bl2_rec_mode = SWITCHTEC_BL2_RECOVERY_I2C_AND_XMODEM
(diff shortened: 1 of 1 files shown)


cli/mfg.c
```diff
@@ -1059,7 +1059,7 @@ static int fw_execute(int argc, char **argv)
 		int assume_yes;
 		enum switchtec_bl2_recovery_mode bl2_rec_mode;
 	} cfg = {
-		.bl2_rec_mode = SWITCHTEC_BL2_RECOVERY_I2C_AND_XMODEM
+		.bl2_rec_mode = 0
```
lsgunth: This is a very very minor nit, but it's common practice to add a trailing comma on lines like this. Then new line additions don't require changing other lines and  makes for cleaner patchs.

lsgunth (approved):
Besides the very nit picky comment this looks good.

cli/mfg.c
BenReed161: added comma

PR #384, opened 2026-05-27
Update BL1 key hash feature

BenReed161:
Added all missing OTP DWORD defines: indices 11-13, 21-23, 25, and complete mask/LSB defines for I3C fields, algorithm disables, boot/failover disables, token disables, hash table disables, key hash statuses (BIAK0S-BIAK11S), and PUFAC.

Extended security_settings_get_gen6() to parse all these fields from the raw OTP during BL1 phase, matching the internal repo's behavior.


lsgunth (approved):
LGTM

PR #387, opened 2026-06-02
Include dok_key_revoke in libcrypto check

BenReed161:
Users who choose not to install option dependency libssl are seeing issues compiling, dok_key_revoke in the command struct array should be in lib crytpo check too. Along with associated helper functions.


lsgunth (approved):
Excellent, this is timely. I just hit the same bug while trying to build the project on a new VM. Thanks!

PR #388, opened 2026-06-05
Fix requests

BenReed161:


cli/main.c
```diff
@@ -2174,6 +2191,27 @@ static int fw_update(int argc, char **argv)
 		}
 	}
 
+	{
```
lsgunth: Maybe this unnamed block would be better off as a static helper function with a name?

lsgunth (approved):
One nit, otherwise looks good.

cli/main.c
BenReed161: Added static helper function. Thanks

BenReed161: Actually I decided to remove that commit. There was discussion from the Firmware side which there will be a clearer error passed down to the switchtec-user rather than switchtec-user handle and process it.

PR #390, opened 2026-06-09
Fix I/O issue with eye cap inband

BenReed161:
Linux kernel switchtec driver has roughly 500ms set as a timeout for the MRPC command, if firmware does not respond in that time we see kernel time out and the MRPC state comes back as IO_ERROR and subsequent read() calls return -EIO when attempting another MRPC command.

The fix in this commit is when read_resp() returns EIO we will reopen the device using the file desc. This will reset kernel MRPC state allowing next command to succeed. If switchtec_cmd returns EIO in the eye status command, we should wait 6 seconds (firmware to catch up & finish) and then retry, up to 5 mins timeout. Increase poll interval from 200 to 2000 to reduce contention.

### DIFF
--- a/lib/diag.c
+++ b/lib/diag.c
@@ -30,8 +30,9 @@
 #define SWITCHTEC_LIB_CORE
 #define SWITCHTEC_LTSSM_MAX_LOGS 61
 #define EYE_CAP_STATUS_TIMEOUT_MS (5 * 60 * 1000)
-#define EYE_CAP_STATUS_POLL_MS 200
-#define EYE_CAP_PROGRESS_INTERVAL 25
+#define EYE_CAP_STATUS_POLL_MS 2000
+#define EYE_CAP_STATUS_RETRY_MS 6000
+#define EYE_CAP_PROGRESS_INTERVAL 5
 
 #include "switchtec_priv.h"
 #include "switchtec/diag.h"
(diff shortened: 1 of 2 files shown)


lib/diag.c
```diff
@@ -144,9 +146,15 @@ static int switchtec_diag_eye_status_gen5(struct switchtec_dev *dev)
 	struct switchtec_gen5_diag_eye_status_out out;
 
 	do {
-		ret = switchtec_cmd(dev, MRPC_GEN5_EYE_CAPTURE, &in, sizeof(in),
-				    &out, sizeof(out));
+		ret = switchtec_cmd(dev, MRPC_GEN5_EYE_CAPTURE, &in,
+				    sizeof(in), &out, sizeof(out));
 		if (ret) {
+			if (errno == EIO) {
```
lsgunth, line 152: Small nit: the diag changes really should have been a separate patch.

lib/platform/linux.c
```diff
@@ -394,7 +410,12 @@ static int linux_cmd(struct switchtec_dev *dev,  uint32_t cmd,
 	if (ret < 0)
 		return ret;
 
-	return read_resp(ldev, resp, resp_len);
+	ret = read_resp(ldev, resp, resp_len);
```
lsgunth, line 413: This is a good solution that makes sense for what we have now, but if there are more long running mrpc commands maybe we need to update the driver to allow for a longer timeout when it is needed.

lsgunth (approved):
A couple comments, but generally this  looks good.

lib/diag.c
BenReed161: Separated changes.

PR #391, opened 2026-06-09
Add missing cfg arguements for dok add/revoke

BenReed161:


cli/mfg.c
```diff
@@ -3072,7 +3072,9 @@ static int dok_key_add(int argc, char **argv)
 		char *uid_hex;
 		char *psid_hex;
 		FILE *key_fimg;
+		const char *key_filename;
```
lsgunth, line 3075: I'm not following this patch. I don't see where these new elements in the cfg structure are used.

BenReed161: Every CFG_FILE_R or CFG_FD_WR field in any argconfig cfg struct is followed by a const char* field to absorb the filename pointer that argconfig writes at value_addr + sizeof(FILE*). When the key argument gets parsed it opens a file to key_fming pointer and then writes the filename to the next configuration argument which was sig_fimg file pointer, this ends up as a garbage pointer and we do a check later for if we are using check-type arg and sig file which isn't allowed and causes the command to error (even if the user never sets the sig file argument). Even if key_filename is never used it is needed for how the config is parsed. 
Originally when I was given this diff I didn't realize this was the case, the changes related to mfg and security have been passed to me and I didn't review it enough to see this.

BenReed161:
Updated commit message explaining the change.

cli/mfg.c
lsgunth: Right, thanks for reminding me. That's a pretty ugly design detail of the argparse code that I wrote 🙁. I was young and didn't know any better.

lsgunth (approved):

PR #392, opened 2026-06-15
Add Gen6 y-range check, increase timeout

BenReed161:
Add the Gen6 y-range check, Gen5 and 6 share the same y_range settings. Increase the eye timeout MS to 15 mins, instead of 5 mins we need at least 10 minutes to collect a full eye capture across the lane.


lsgunth (approved):
LGTM

PR #394, opened 2026-06-16
Add customer fields and ECC fields for dev cfg set

BenReed161:
Add the customer specific fields and ECC fields to the device config set customer command.


lsgunth (approved):
Seems fine

PR #395, opened 2026-06-23
Add endptr that catches 08/09 invalid octal

BenReed161:
Add *endptr != '\0' which catches "08"/"09" (invalid octal stops early) and any other trailing garbage.


lsgunth (approved):
LGTM

PR #396, opened 2026-06-23
Add check if in MRPC error range

BenReed161:
When a Switchtec device is in INITIALIZED_SECURED state (locked), firmware denies certain MRPC commands with error code 0x64010 (ERR_MRPC_DENIED). Some commands correctly report "MRPC request denied" (fw-info), while others report "Unknown Error (0x64010)" (fw-update, mfg fw-transfer, mfg fw-execute).

Modify switchtec_fw_perror() in lib/fw.c to check whether the return value falls in the 0x64xxx range or matches known MRPC error codes and, if so, fall through to switchtec_perror() which already handles all MRPC errors correctly.


lsgunth (approved):
LGTM

PR #397, opened 2026-06-25
Add firmware and config toggle fix

BenReed161:
If firmware and config are toggled at the same time, we need to toggle as part of the same command. Individual toggle for either will cause the version mismatch error.


lsgunth (approved):
That makes sense to me and it's a nice cleanup.

PR #399, opened 2026-07-13
Add HLC detect function for stack_bif

BenReed161:
Dynamic stack bifurcation is not supported on High lane Gen6 varient. Check device id to detemine which type of device it is. 0x44 = 144 lane, 0x60 - 160 lane. All others are low lane varient.

Add stack bifurcation unsupported error.


lsgunth (approved):
LGTM

PR #401, opened 2026-07-21
Stack and gen6 speed update

BenReed161:


inc/switchtec/switchtec.h
```diff
@@ -546,6 +545,29 @@ static inline int switchtec_max_supported_ports(struct switchtec_dev *dev)
...
+	int dev_type;
+
+	if (switchtec_gen(dev) != SWITCHTEC_GEN6)
+		return 8;
+
+	dev_type = switchtec_device_id(dev) & 0xFF;
+
+	switch (dev_type) {
+	case 0x60:
+	case 0x44:
```
lsgunth: I feel like these bare numbers should be constants/defines somewhere with appropriate names, if they aren't already.

BenReed161: switch case updated with defines that have names with HLC/LLC

lsgunth (approved):
LGTM, thanks for adjusting that!

PR #402, opened 2026-07-23
Add GAS-based firmware download for cascaded devices

BenReed161:
Firmware download to cascaded Switchtec devices fails because the kernel's MRPC path (both DMA and interrupt-based) is unreliable through cascade fabric. Additionally, the kernel ISR races with userspace when both access the MRPC mailbox registers simultaneously.

Add a direct GAS register polling path that bypasses the kernel MRPC entirely. When GAS BAR mapping is available, fw_gasop_cmd() writes commands directly to MRPC registers, uses sfence to flush write-combining buffers, and polls status with a 120s timeout treating 0xffffffff reads as "device busy".

Before starting the GAS-based download, mask EN_IRQ (bit 3) on the mrpc_comp_hdr and mrpc_comp_async_hdr event registers. This prevents the firmware from generating MSI-X completions that would trigger the kernel's switchtec_event_isr, which otherwise reads/writes the same MRPC registers and can submit queued kernel commands that corrupt the download state. Original IRQ settings are restored on exit.

Falls back transparently to the kernel MRPC path if GAS mapping fails.

### DIFF
--- a/lib/fw.c
+++ b/lib/fw.c
@@ -37,10 +37,15 @@
 #include "switchtec/mfg.h"
 
 #include <unistd.h>
+#include <sys/time.h>
 
 #include <errno.h>
 #include <stdio.h>
 #include <string.h>
+
+#define FW_DL_MAX_RETRIES    5
+#define FW_DL_RETRY_DELAY_US 100000
+#define FW_DL_MRPC_TIMEOUT_MS 120000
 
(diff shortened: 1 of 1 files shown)


lsgunth (approved):
This is a lot to review in one patch, but a quick scan shows nothing objectionable.

PR #404, opened 2026-08-05
Fix bug after reset fw-read & slot active/inactive bug

BenReed161:


lib/fw.c
```diff
@@ -407,6 +407,7 @@ int switchtec_fw_img_get(struct switchtec_dev *dev, int fd,
 		return 1;
 	}
 
+
```
lsgunth: Extra blank line?

lsgunth (approved):
One very minor cosmetic thing, other than that it looks good to me. Thanks!

lib/fw.c
BenReed161: Thanks for the catch, reverted change

PR #405, opened 2026-08-10
Add device cfg info

BenReed161:
Add device config info offline tool to print info on the Chiplink otp configuration binary files


lsgunth (approved):
LGTM

PR #406, opened 2026-08-13
Add user routing rules command

BenReed161:
Add a new command 0x149 Custom user routing rules command. This command will be used to configure user defined entries in AR, ID. This is for Gen6 Switchtec.
Added three new routing rule commands. Add, delete, and show AR and ID routing rules, either on a stack or specified port. Added new error codes associated with the new routing rules commands.

[routing_rules_test_res.log](https://github.com/user-attachments/files/31048528/routing_rules_test_res.log)
Attached is the test log from the routing rule commands.


lsgunth (approved):
Would have been easier to review if this were split into a few more patches instead of one giant one. But on balance this looks good.

PR #407, opened 2026-08-14
Initialize firmware image info

learnrahulrai-ui:
Zero the optional switchtec_fw_image_info output before parsing so callers never observe uninitialized fields.

Tested with ./configure and make.

### DIFF
--- a/lib/fw.c
+++ b/lib/fw.c
@@ -1384,6 +1384,9 @@ int switchtec_fw_file_info(int fd, struct switchtec_fw_image_info *info)
 	char magic[4];
 	int ret;
 
+	if (info)
+		memset(info, 0, sizeof(*info));
+
 	ret = read(fd, &magic, sizeof(magic));
 	lseek(fd, 0, SEEK_SET);
 


lib/fw.c
```diff
@@ -1384,6 +1384,16 @@ int switchtec_fw_file_info(int fd, struct switchtec_fw_image_info *info)
...
 
+	/* Define every byte of the caller's structure before any return path
+	 * can be taken. Without this, the DCBI branch below returns 1 - a
+	 * non-error - having written nothing, and the magic/short-read error
+	 * paths return without writing either. Callers that do not
+	 * zero-initialise are then reading whatever was already on the stack.
+	 * info is optional: callers may pass NULL to use this as a validity
+	 * probe, which the sub-parsers honour (:1342, :1204). */
+	if (info)
+		memset(info, 0, sizeof(*info));
```
lsgunth, line 1388: Seems like a huge amount of slop filled commit message and comments for a two line, pretty obvious, change. I did not bother to read most of it.

Any chance we can clean up all the noise?

learnrahulrai-ui: Cleaned up the commit message, PR description, and inline comment. Thanks.

lsgunth (approved):
Might have been better with a touch more (if there was a bug or not) but I asked for brief and that's what I got.

BenReed161 (approved):
lgtm

PR #408, opened 2026-08-19
Fix Windows compile errors

BenReed161:
Errors seen when compiling on Windows,

Move the GEN6 SN_VER_GET subcommands defined macros in mfg.c outside the ifdef linux which were being excluded from the windows build.

Add include for switchtec/endian.h in cli/main.c and lib/cap.c which has ifdefs for both operating systems and includes the winsock implementations for byte endianness convertion.


lsgunth (approved):
Looks good to me.

PR #412, opened 2026-08-31
Add Gen6 support for device id get in BL2

BenReed161:
Device id get default fall back was the Gen5 version of the Part info command. Added an additional fallback if that one fails to use the Gen6 version of the command.


lsgunth (approved):
LGTM

